PR #54, opened 2014-12-18
Make EdDSA work as documented

jedisct1:
Hi,

I'm a bit confused by the `EdDSA` class. It doesn't seem to do what the headers suggest.

For example, `EdDSA::sign()` computes a signature but doesn't store it anywhere.
Here are a few changes to simplify the code and make it work as advertised.


sergiohs84:
This PR is totally outdated. The corresponding basecode is currently in `develop`

PR #72, opened 2015-03-09
Contact revamp (am and jp)

javiserrano:
Contact management using the new API


javiserrano:
Another pull request already added these commits

PR #83, opened 2015-05-07
Java (desktop) example for Windows

sergiohs84:
A new Visual Studio solution allows to create a native library (mega.dll) which enbables Java apps to use the SDK in Windows.
Instructions for Java example have been updated too.


pohutukawa:
It's usually not a good idea to have derived artifacts checked into a source code repository. That can be a nightmare to keep consistent.

Example:
* Alice makes a change to a source file, but does not re-generate the dependent files (maybe for different bindings) --> boom, things break
* Bob is using a different version of the code generator (SWIG), which might have fixed bugs or newer features. The generated files in the repo have been generated with a different version, and not things are incompatible (I've seen version 2.0.10 and 2.0 11 already in the header comments of the generated files) --> boom, things break

We should rather document the exact dependencies and steps to build, than to include derived artifacts, at least in the version control system. However, we may provide ZIP files for convenience that do include already a generated set of derived artifacts, but these are *not* used for development, but only for "end use".

sergiohs84:
@pohutukawa, I see the problem regarding derived artifacts and I agree with you on not to add them to the repository. Instead, these dependencies would be available to download for Windows users (autotools, SWIG, etc. can be hard to configure in MS platform). For Linux users, they can simply "make" the SDK in order to create the Java classes and compile the .so library.

Please, have a look at the two updated README.md files (bindings/java/vs2010 and examples/java) and let me know if you see some improvement, amendment...

bindings/java/vs2010/README.md
```diff
@@ -0,0 +1,33 @@
...
+5. Open the solution `MEGAdll.sln` and hit "Build".
+
+6. If success, you should find a `mega.dll` library in `sdk/bindings/java/vs2010/Debug/`.
+
+
+## Notes
+
+The provided VS solution uses the files `megaapi_wrap.h` and `megaapi_wrap.cpp` located in the folder `sdk/bindings/java`. The generated DLL must be provided with the Java clases in `skd/bindings/java/nz/mega/sdk/`.
+
+Should you need an updated version of any of those files (`megaapi_wrap.cpp`, `megaapi_wrap.h` or any of the Java classes), then you need to build the SDK under Linux by following the instructions [here](https://github.com/meganz/sdk#building), adding the option `--enable-java` to your configure command.
```
pohutukawa: I'd suggest wording it like this:

```Should you need an updated version of any of those files (`megaapi_wrap.cpp`, `megaapi_wrap.h` or any of the Java classes), then you need to generate some build-related dependencies by following the instructions [here](https://github.com/meganz/sdk#building), adding the option `--enable-java` to your configure command. Due to build tool requirements it might be easiest to do this on a different platform (e.g. a Linux system).```

File: examples/java/README.md [hunk 2]
```diff
@@ -20,16 +17,43 @@ To use a prebuilt library (the easy way), follow these steps:
...
     ```
     
-6. Build and run the project in Eclipse
+3. Copy the Java classes into the `src` folder in your project.
+
+    ```
+    cp bindings/java/nz/mega/sdk/*.java examples/java/JavaBindingSample/src/nz/mega/sdk
+    ```
+    
+4. In Eclipse, click "Import" -> "Existing Projects into Workspace" and select the root directory: `sdk/examples/java`
```
pohutukawa: If the Eclipse project files are relocated elsewhere (out of the way of the root, e. g into `contrib/Eclipse`), then there needs to be a word about this. In this case, it might make sense to create a section for Eclipse detailing the steps:
```How to import the project into Eclipse```

File: examples/java/README.md [hunk 3]
```diff
@@ -1,13 +1,10 @@
...
-Eclipse project for developing a MEGA app for Java.
+Eclipse project for developing a MEGA app in Java.
 
 This example shows a simple GUI (Swing) to allow the user log into MEGA. If login is successfull, it retrieves the MEGA filesystem and lists files/folders in the root folder.
 
-## How to run the project:
-
-You have two options, using a prebuilt native library or building it by yourself.
-To use a prebuilt library (the easy way), follow these steps:
+## How to run the project in LINUX:
```
pohutukawa: Let's not have Linux in all caps. "Linux" is just fine ... :)

examples/java/README.md
```diff
@@ -20,16 +17,43 @@ To use a prebuilt library (the easy way), follow these steps:
...
+3. Copy the Java classes into the `src` folder in your project.
+
+    ```
+    cp bindings/java/nz/mega/sdk/*.java examples/java/JavaBindingSample/src/nz/mega/sdk
+    ```
+    
+4. In Eclipse, click "Import" -> "Existing Projects into Workspace" and select the root directory: `sdk/examples/java`
+5. Build and run the project in Eclipse
+
+## How to run the project in WINDOWS
```
pohutukawa: Same with "Windows" (instead of all caps)

examples/java/README.md
```diff
@@ -1,13 +1,10 @@
 # Java Example App
 
-Eclipse project for developing a MEGA app for Java.
+Eclipse project for developing a MEGA app in Java.
 
 This example shows a simple GUI (Swing) to allow the user log into MEGA. If login is successfull, it retrieves the MEGA filesystem and lists files/folders in the root folder.
```
pohutukawa: It might be good to list required build dependencies here. Off the top of my head I can think of the following (good, it's a short list):
* compiler, build tool chain (on Linux probably mostly in `build-essential`)
* SWIG (what version)
* J2SDK (what version? I'd guess >= Java 7)
* anything else?

pohutukawa:
Looks good to me now. Merging ...

PR #84, opened 2015-05-07
Crypto stuff related to payments

javiserrano:
Contains the new required functions and some tests for all of them.


include/mega/utils.h
```diff
@@ -95,6 +95,99 @@ class MEGA_API HashSignature
...
+     */
+    PayCrypter();
+
+    /**
+     * @brief Updates the crypto keys (mainly for testing)
+     * @param newEncKey New AES key (must contain ENC_KEY_BYTES bytes)
+     * @param newHmacKey New HMAC key (must contain MAC_KEY_BYTES bytes)
+     * @param newIv New IV for AES-CBC (must contain IV_BYTES bytes)
+     */
+    void setKeys(byte *newEncKey, byte *newHmacKey, byte *newIv);
```
pohutukawa: shouldn't we mark these parameters as `const`?

include/mega/utils.h
```diff
@@ -95,6 +95,99 @@ class MEGA_API HashSignature
...
+     * @brief Encrypts the cleartext and returns the payload string.
+     *
+     * The clear text is encrypted with AES-CBC, then a HMAC-SHA256 is generated for (IV + ciphertext)
+     * and finally returns (HMAC + IV + ciphertext)
+     *
+     * @param cleartext Clear text to generate the payload
+     * @param result The function will fill this string with the generated payload
+     * @return True if the funcion succeeds, otherwise false
+     */
+    bool encryptPayload(string *cleartext, string *result);
```
pohutukawa: mark `cleartext` as `const`?

include/mega/utils.h
```diff
@@ -95,6 +95,99 @@ class MEGA_API HashSignature
...
+     * The result is padded with random bytes. Then RSA is applied and the result is returned
+     * in the third parameter, with a 2-byte header that contains the size of the result of RSA.
+     *
+     * @param cleartext Clear text to encrypt with RSA
+     * @param base64pubk Public key in Base64 format (result of AsymmCipher::serializekey and Base64::btoa)
+     * @param randompadding Enables padding with random bytes. Otherwise, the cleartext is 0-padded
+     * @param result RSA encrypted text, with a 2-byte header with the size of the RSA buffer in bytes
+     * @return True if the funcion succeeds, otherwise false
+     */
+    bool rsaEncryptKeys(string *cleartext, char *base64pubk, string *result, bool randompadding = true);
```
pohutukawa: mark `cleartext` and `base64pubk` as `const`?
I'm personally not a fan of having different formats for parameters (`base64pubk`), as that leads to massive confusion. In my code in the SDK as well as the webclient I have usually stuck to using binary strings (8-bit `byte`/`unsigned char` arrays) as a canonical default format. This makes using it all *much* easier!

src/crypto/cryptopp.cpp
```diff
@@ -117,6 +117,42 @@ void SymmCipher::cbc_decrypt(byte* data, unsigned len, const byte* iv)
...
+ * The size of the IV is one block in AES-128 (16 bytes).
+ *
+ * @param data Data to be encrypted
+ * @param iv Initialisation vector.
+ * @param result Encrypted message
+ * @return Void.
+ */
+void SymmCipher::cbc_encrypt_pkcs_padding(string *data, const byte *iv, string *result)
+{
+    aescbc_e.Resynchronize(iv ? iv : zeroiv);
```
pohutukawa: I'm not *that* familiar with C++, but to me it seems that `Resynchronize` is a method/function. It would be good to start them with a lower case letter for consistency like the rest (we're not dealing with MS Windows naming conventions here).

src/crypto/cryptopp.cpp
```diff
@@ -117,6 +117,42 @@ void SymmCipher::cbc_decrypt(byte* data, unsigned len, const byte* iv)
...
+ * @brief Encrypt symmetrically using AES in CBC mode and pkcs padding
+ *
+ * The size of the IV is one block in AES-128 (16 bytes).
+ *
+ * @param data Data to be encrypted
+ * @param iv Initialisation vector.
+ * @param result Encrypted message
+ * @return Void.
+ */
+void SymmCipher::cbc_encrypt_pkcs_padding(string *data, const byte *iv, string *result)
```
pohutukawa: mark `data` as `const`?

src/crypto/cryptopp.cpp
```diff
@@ -117,6 +117,42 @@ void SymmCipher::cbc_decrypt(byte* data, unsigned len, const byte* iv)
...
+ * The size of the IV is one block in AES-128 (16 bytes).
+ *
+ * @param data Data to be decrypted
+ * @param iv Initialisation vector.
+ * @param result Decrypted message
+ * @return Void.
+ */
+void SymmCipher::cbc_decrypt_pkcs_padding(string *data, const byte *iv, string *result)
+{
+    aescbc_d.Resynchronize(iv ? iv : zeroiv);
```
pohutukawa: same comments as for `cbc_encrypt_pkcs_padding`

src/utils.cpp
```diff
@@ -20,6 +20,7 @@
  */
 
 #include "mega/utils.h"
+#include "mega/base64.h"
```
pohutukawa: We don't need bases64 here if the `pubk` parameter is in a binary format already.

tests/tests.cpp
```diff
@@ -49,6 +49,160 @@ TEST(Serialize64, serialize) {
     ASSERT_EQ(in, out);
 }
 
+TEST(PayCrypterTest, allFeatures)
```
pohutukawa: Would be good if this could be moved to its own test file, not in the generic `test.cpp` (maybe `paycrypt_test.cpp` or so)

PR #86, opened 2015-05-14
Restore missing attributes on synced folders

javiserrano:


src/megaapi_impl.cpp
```diff
@@ -4511,9 +4511,19 @@ void MegaApiImpl::unlink_result(handle h, error e)
 void MegaApiImpl::fetchnodes_result(error e)
 {
 	MegaError megaError(e);
```
pohutukawa: Indentation doesn't seem to be correct here. The line `MegaError megaError(e);` (as well as some others I'd suspect) is indented with a TAB.

pohutukawa:
Looks pretty good from here. Would be good to get a second opinion, though. I'm not the sharpest tool in the drawer with C++. :-)

PR #88, opened 2015-05-20
Fix the retry of getfa and provide the error code

javiserrano:


src/commands.cpp
```diff
@@ -113,7 +113,16 @@ void CommandGetFA::procresult()
     if (client->json.isnumeric())
     {
         if (it != client->fafcs.end())
-        {
+        {            
+            faf_map::iterator fafsit;
+            for (fafsit = it->second->fafs[0].begin(); fafsit != it->second->fafs[0].end(); )
+            {
+                // move from fresh to pending
+                it->second->fafs[1][fafsit->first] = fafsit->second;
```
javiserrano: Without this code, the error callback isn't sent to the app because `FileAttributeFetchChannel::failed()` uses `fafs[1]` only.

src/fileattributefetch.cpp
```diff
@@ -31,6 +31,7 @@ FileAttributeFetchChannel::FileAttributeFetchChannel()
     urltime = 0;
     fahref = UNDEF;
     inbytes = 0;
+    e = API_EINTERNAL;
```
javiserrano: I set it to `API_EINTERNAL` in the constructor because it should be set again before being used in a callback.

src/fileattributefetch.cpp
```diff
@@ -70,6 +71,7 @@ void FileAttributeFetchChannel::dispatch(MegaClient* client)
 
     if (req.outbuf.size())
     {
+        e = API_EFAILED;
```
javiserrano: After this, only the transfer can fail without any API error code. I set the error to `API_EFAILED` for consistency with errors generated in regular transfers (they use `API_EFAILED` too)

src/fileattributefetch.cpp
```diff
@@ -168,7 +170,11 @@ void FileAttributeFetchChannel::failed(MegaClient* client)
         {
             // retry
             it->second->retries++;
-            it++;
+
+            // move from pending to fresh
```
javiserrano: Without this code, the request isn't sent again. If the state `REQ_FAILED` persists, an error is detected in the next event loop, without retrying.
Also, the `FileAttributeFetch` object is put again in the map in which it starts.

PR #91, opened 2015-05-28
Regular Expressions with PCRE

sergiohs84:
The API and SDK have been modified to provide regular expressions matching. It will be useful to exclude certain files, subfolders or certain patterns (ie. *.bak) when syncing a folder.

The current implementation uses the [PCRE library](http://www.pcre.org/), but additional implementations can be easily added by using a different macro than `USE_PCRE`.

In case you want to compile the SDK without regular expressions support, simply skip the definition `ENABLE_REGEXP` in the file _sdk.pri_.


javiserrano:
Another branch adding this functionality has been already merged into `develop`

PR #116, opened 2015-07-30
Manage the new encryption key associated with the session ID

javiserrano:


src/megaapi_impl.cpp
```diff
@@ -2411,14 +2413,12 @@ char *MegaApiImpl::dumpSession()
...
     char* buf = NULL;
-    int size;
-    size = client->dumpsession(session, sizeof session);
-    if (size > sizeof(client->key.key))
+
+    if (client->loggedin())
     {
-        buf = new char[sizeof(session)*4/3+4];
-        Base64::btoa(session + sizeof(client->key.key), size - sizeof(client->key.key), buf);
+        buf = new char[64 * 4 / 3 + 4];
```
sergiohs84: Here, I guess the value '64' represents the maximum session length. If so, better to use the new macro MAX_SESSION_LENGTH.

javiserrano: Fixed

PR #127, opened 2015-08-13
Missing target email for incoming contact requests

sergiohs84:
Not really necessary, since the target email is not used (by now) during incoming contact requests management, but I think it's better to set all the attributes of the requests.


sergiohs84:
Users are not created yet during the fetchnodes stage where Incoming Contact Request are built, so the own User object is not available. Better to keep the target email empty for those cases.

PR #146, opened 2015-09-07
Changes to Javadoc inline comments and the userguide

PatrickJBaird:
Created new repo for the pull request as couldn't correct my mistake on the old repo. We do have a backup copy of the other repo incase the previous commits are required.


bindings/java/nz/mega/sdk/DelegateMegaGlobalListener.java
```diff
@@ -2,6 +2,25 @@
 
 import java.util.ArrayList;
 
+/**
```
pohutukawa: This comment block goes to the *head* of the Java file, and it's usually also not in JavaDoc (it's informational in the code only, not to be extracted by JavaDoc).

The way it's done now is, that it's the documentation of the `DelegateMegaGlobalListener` class, which it's not, but rather some meta-data for the file itself.

bindings/java/nz/mega/sdk/DelegateMegaGlobalListener.java
```diff
@@ -2,6 +2,25 @@
...
+ * Listener to receive and send global events to the app.
+ * <p>
+ * (c) 2013-2014 by Mega Limited, Auckland, New Zealand
```
pohutukawa: 2013-2015

bindings/java/nz/mega/sdk/DelegateMegaListener.java
```diff
@@ -2,6 +2,25 @@
 
 import java.util.ArrayList;
 
+/**
```
pohutukawa: Same comment as for `DelegateMegaGlobalListener` class, and all following (which will probably have the same issue).

bindings/java/nz/mega/sdk/DelegateMegaTreeProcessor.java
```diff
@@ -1,5 +1,10 @@
 package nz.mega.sdk;
 
+/**
+ * Interface to process node trees.
+ * <p>
```
pohutukawa: Are these paragraph indicators (`<p>`) really required for rendering the JavaDoc output correctly? I think a blank line suffices (but I'm not sure). I think HTML markup is only needed for more detailed layout stuff (such as itemised or numbered lists).

bindings/java/nz/mega/sdk/MegaApiSwing.java
```diff
@@ -2,8 +2,24 @@
...
 public class MegaApiSwing extends MegaApiJava {
 
+    /**
+     * Instantiates a new MEGA API using Swing.
+     *
+     * @param appKey
+     *              AppKey of your application. You can generate your AppKey for free here: <br>
+     *              https://mega.co.nz/#sdk
+     * @param userAgent
+     *              User agent to use in network requests. If you pass null to this parameter,
```
pohutukawa: It's a good idea to *always* use full sentences, that is: end them with a full stop. It reads awkwardly to have two sentences, and the second one isn't terminated. Not a big issue, but it's good practice to terminate *all* sentences (even single ones) in the doc strings.

bindings/java/nz/mega/sdk/MegaGlobalListenerInterface.java
```diff
@@ -3,9 +3,44 @@
 import java.util.ArrayList;
 
 public interface MegaGlobalListenerInterface {
```
pohutukawa: No documentation on the interface?

It's common to document the interfaces, and then use on the implementation (e. g. the one that `implements` or `extends`) a `@see` documentation:
http://www.tutorialspoint.com/java/java_documentation.htm

bindings/java/nz/mega/sdk/MegaLoggerInterface.java
```diff
@@ -1,5 +1,36 @@
...
+     * <p>
+     * By default log level is MegaApi.LOG_LEVEL_INFO.
+     *  
+     * @param time
+     *            Readable string representing the current time.
+     *            The SDK retains the ownership of this string, it will not be valid after this function returns
+     * @param loglevel
+     *            Log level of this message.
+     *            Valid values are: <br>
+     *          • LOG_LEVEL_FATAL = 0 <br>
```
pohutukawa: Try to avoid unicode characters (the bullet). Use an "unordered list" (from HTML) instead, or just an asterisk (`*`) or dash (`-`) to keep it simple.

PR #149, opened 2015-09-11
Megaapi updates

sergiohs84:
- Removal of deprecated function `MegaApi::addContact()`
- Non-blocking methods for sharing information: `MegaNode::isShared()`, `isOutShare()` and `isInShare()`.
- Reduction of memory requirements for `MegaNodePrivate` by using single bits for boolean attributes.


javiserrano:
`MegaApi::isShared` is used at least in Java bindings and WP bindings, `MegaApi::isInShare` and `MegaApi::isOutShare` is used at least in WP bindings, so this pull request at this time would break at least the example app for Java, Android and Windows Phone.

sergiohs84:
By now, we could modify `MegaApi` methods to be a wrapping of the new methods in `MegaNode`. It would be harmless, it doesn't break anything and the benefit is that `MegaApi` doesn't need to block the SDK mutex.

javiserrano:
I think that it's a great idea

PR #157, opened 2015-10-01
Link status

sergiohs84:
- New `PublicLink` structure to manage the information of links
 - New `MegaApi` methods to check the state of links
 - Management of expiration of links
 - Management of takendown links
 - `megacli` adaptation to show state of file links
 - Backwards compatibility for previous databases

### DIFF
--- a/include/megaapi.h
+++ b/include/megaapi.h
@@ -533,13 +533,22 @@ class MegaNode
         virtual int getTag();
 
         /**
-         * @brief Returns the structure representing a PublicLink
+         * @brief Returns the expiration time of a public link, if any
          *
-         * You take the ownership of the returned value.
-         *
-         * @return Structure with the public link information
+         * @return The expiration time as an Epoch timestamp. Returns 0 for non-expire
+         * links, and -1 if the MegaNode is not exported.
          */
-        virtual PublicLink * getPublicLink();
+        virtual int64_t getExpirationTime();
+
+        /**
+         * @brief Returns the public handle of a node
+         *
+         * Only exported nodes have a public handle.
+         *
+         * @return The public handle of an exported node. If the MegaNode
+         * has not been exported, it returns UNDEF.
+         */
+        virtual handle getPublicHandle();
 
         /**
          * @brief Returns true if this node represents a file (type == TYPE_FILE)
@@ -661,13 +670,23 @@ class MegaNode
         virtual bool isPublic();
 
         /**
-         * @brief Returns true if this the node has been exported (has a public link)
+         * @brief Returns true if the node has been exported (has a public link)
          *
          * Public links are created by calling MegaApi::exportNode.
          *
          * @return true if this is an exported node
          */
         virtual bool isExported();
+
+        /**
+         * @brief Returns true if the node has been exported (has a temporal public link)
+         * and the related public link has expired.
+         *
+         * Public links are created by calling MegaApi::exportNode.
+         *
+         * @return true if the public link has expired.
+         */
+        virtual bool isExpired();
 
         /**
          * @brief Returns true if this the node has been exported
--- a/include/megaapi.h
+++ b/include/megaapi.h
@@ -533,7 +533,7 @@ class MegaNode
         virtual int getTag();
 
         /**
-         * @brief Returns a structure with the public link information (handle and expiration time)
+         * @brief Returns the structure representing a PublicLink
          * @return Structure with the public link information
          */
         virtual PublicLink * getPublicLink() const;
@@ -665,6 +665,16 @@ class MegaNode
          * @return true if this is an exported node
          */
         virtual bool isExported();
+
+        /**
+         * @brief Returns true if this the node has been exported
+         * and the related public link has been taken down.
+         *
+         * Public links are created by calling MegaApi::exportNode.
+         *
+         * @return true if the public link has been taken down.
+         */
+        virtual bool isTakenDown();
 
         /**
          * @brief Returns a string that contains the decryption key of the file (in binary format)
--- a/include/megaapi.h
+++ b/include/megaapi.h
@@ -4277,7 +4277,7 @@ class MegaApi
          *
          * @note A Unix timestamp represents the number of seconds since 00:00 hours, Jan 1, 1970 UTC
          */
-        void exportNode(MegaNode *node, int expireTime, MegaRequestListener *listener = NULL);
+        void exportNode(MegaNode *node, int64_t expireTime, MegaRequestListener *listener = NULL);
 
         /**
          * @brief Stop sharing a file/folder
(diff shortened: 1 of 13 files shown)


include/megaapi.h
```diff
@@ -533,6 +533,43 @@ class MegaNode
...
+
+        /**
+         * @brief Returns the public handle of a node
+         *
+         * Only exported nodes have a public handle.
+         *
+         * @return The public handle of an exported node. If the MegaNode
+         * has not been exported, it returns UNDEF.
+         */
+        virtual handle getPublicHandle();
```
javiserrano: It should return a MegaHandle

include/megaapi.h
```diff
@@ -533,6 +533,43 @@ class MegaNode
...
+         /**
+         * @brief Returns a public node corresponding to the exported MegaNode
+         *
+         * @return Public node for the exported node. If the MegaNode has not been
+         * exported, it returns NULL.         *
+         */
+        virtual MegaNode* getPublicNode();
+
+        /**
+         * @brief Returns the URL for the public link of the exported node.
```
javiserrano: Does it return the link even if it is expired?

src/megaapi_impl.cpp
```diff
@@ -291,6 +300,56 @@ int MegaNodePrivate::getTag()
...
+}
+
+MegaNode* MegaNodePrivate::getPublicNode()
+{
+    if (!plink)
+    {
+        return NULL;
+    }
+
+    string key = string(getBase64Key());
```
javiserrano: Memory leak

src/megaapi_impl.cpp
```diff
@@ -291,6 +300,56 @@ int MegaNodePrivate::getTag()
...
+
+    char *base64ph = new char[12];
+    Base64::btoa((byte*)&(plink->ph), MegaClient::NODEHANDLE, base64ph);
+
+    string strlink = "https://mega.nz/#";
+    strlink += (type ? "F" : "");
+    strlink += "!";
+    strlink += base64ph;
+    strlink += "!";
+    strlink += getBase64Key();
```
javiserrano: Memory leak

src/node.cpp
```diff
@@ -280,13 +284,16 @@ Node* Node::unserialize(MegaClient* client, string* d, node_vector* dp)
         fa = NULL;
     }
 
-    for (i = 8; i--;)
+//    for (i = 8; i--;)
```
javiserrano: Remove comment

src/node.cpp
```diff
@@ -280,13 +284,16 @@ Node* Node::unserialize(MegaClient* client, string* d, node_vector* dp)
...
+    for (i = 7; i--;)
     {
         if (ptr + (unsigned char)*ptr < end)
         {
             ptr += (unsigned char)*ptr + 1;
         }
     }
+    version_flag = MemAccess::get<char>(ptr);
```
javiserrano: Maybe it would be better to put this on the first free byte, instead of in the last one.
That way, the other bytes could be used to store optional data depending on the version

src/node.cpp
```diff
@@ -325,6 +332,26 @@ Node* Node::unserialize(MegaClient* client, string* d, node_vector* dp)
 
     ptr = n->attrs.unserialize(ptr);
 
+    PublicLink *plink = NULL;
+    if (version_flag >= 1)
+    {
```
javiserrano: We could save memory by directly having the isExported flag in the reserved bytes above, instead of the version_flag.

src/node.cpp
```diff
@@ -417,7 +444,9 @@ bool Node::serialize(string* d)
         d->append(fileattrstring.c_str(), ll);
     }
 
-    d->append("\0\0\0\0\0\0\0", 8);
+//    d->append("\0\0\0\0\0\0\0", 8);
```
javiserrano: Remove comment

src/node.cpp
```diff
@@ -835,6 +877,25 @@ NodeCore::~NodeCore()
...
+    this->takendown = plink->takendown;
+}
+
+bool PublicLink::isExpired()
+{
+    if (!ets)       // permanent link: ets=0
+        return false;
+
+    time_t t = time(NULL);
+//    struct tm *ptm = gmtime(&t);
```
javiserrano: Remove comments

include/megaapi.h
```diff
@@ -4180,6 +4246,24 @@ class MegaApi
...
+         * Valid data in the MegaRequest object received on callbacks:
+         * - MegaRequest::getNodeHandle - Returns the handle of the node
+         * - MegaRequest::getAccess - Returns true
+         *
+         * Valid data in the MegaRequest object received in onRequestFinish when the error code
+         * is MegaError::API_OK:
+         * - MegaRequest::getLink - Public link
+         *
+         * @param node MegaNode to get the public link
+         * @param expireTime Number of days during the public link will be valid
```
javiserrano: It should be a UNIX timestamp (see the api channel) so the type should be changed (also in MegaApiImpl and MegaClient)

include/megaapi.h
sergiohs84: Done

include/megaapi.h
sergiohs84: Now it doesn't, it returns NULL if the link is expired.

include/megaapi.h
sergiohs84: Actually, it was a UNIX timestamp. The documentation on the MegaApi method was wrong

src/megaapi_impl.cpp
sergiohs84: Right! Didn't notice getBase64Key() was allocating dynamic memory :(

src/megaapi_impl.cpp
sergiohs84: Same here, fixed.

src/node.cpp
sergiohs84: Done

src/node.cpp
sergiohs84: Good idea. Changing that!

src/node.cpp
sergiohs84: But in the future, if the version value changes, let's say `version_flag=2` for an additional data, it will mean that the node is exported, and it may not.
I think is better to have one specific byte for that, and to use the reserved byte exclusively for the version number.

src/node.cpp
sergiohs84: Done

src/node.cpp
sergiohs84: I'd keep it there until the expiration feature is live and usable by client apps, so we can test if they are really not mandatory.

The documentation for the `time()` function says:
"The value returned generally represents the number of seconds since 00:00 hours, Jan 1, 1970 UTC (i.e., the current unix timestamp). Although libraries may use a different representation of time: Portable programs should not use the value returned by this function directly, but always rely on calls to other elements of the standard library to translate them to portable types (such as localtime, gmtime or difftime)."

PR #160, opened 2015-10-09
Folder uploads

javiserrano:


src/megaapi_impl.cpp
```diff
@@ -10145,3 +10235,247 @@ FileInputStream::~FileInputStream()
...
+{
+    recursive++;
+    string localPath = pendingFolders.front();
+    pendingFolders.pop_front();
+
+    MegaNode *parent = megaApi->getNodeByHandle(handle);
+
+    string localname;
+    DirAccess* da;
+    da = client->fsaccess->newdiraccess();
```
sergiohs84: `da` is allocated but never released

PR #174, opened 2015-10-30
User attributes

sergiohs84:
- API sends now action packets to the own user and to own contacts for changes in `firstname`, `lastname` --> the SDK notifies the MegaApp via `onUsersUpdate()`
 - Modification of `CommandGetUA`: the parameter `priv` has been removed, since there is not a real scope for attributes visibility, but a full-specified name. Better to handle the attribute name as a whole instead of based on a prefix (which doesn't really indicate the scope of the attribute)
 - Fix for non-existing avatar: the value is not Base64 encoded and needs special treatment.
 - Adaptation of `megacli`
 - New SDK integration test for settting/getting user attributes


sergiohs84:
API required changes are already live on production servers. We can merge this branch and start using the new features-improvements on mobile apps.

PR #179, opened 2015-11-13
Creation of chat keys

sergiohs84:
- After login, private keys are requested (keyring attribute). If not available, they are created and saved into MEGA.
- At reception of private keys, public keys are derived and also requested (puEd255 and puCu255). 
- At reception of public keys, they are checked against the derived keys. If they don't match, the derived version is saved into MEGA.
- A new class `ECDH` wraps the functionality of chat keys, as it does the `EdDSA` for signing keys.
- The methods to `cipher()` and `ECDH::uncipher()` are a first approach, not tested yet.
- Modification of `CryptoPP::gcm_encrypt/decrypt()` to use the authentication tag.


include/mega/crypto/sodium.h
```diff
@@ -95,6 +95,85 @@ class MEGA_API EdDSA
...
+     * @param m Message to be encrypted. Caller must ensure that the first
+     * crypto_box_ZEROBYTES bytes of the message m are all 0.
+     * @param n Number used once. The same nonce must never be used to encrypt another
+     * packet from the sender's private key to this receiver's public key or viceversa.
+     * @param pubKey Public key of the receiver.
+     * @param privKey Private key of the sender.
+     *
+     * @return 1 on success, 0 on failure.
+     */
+    int cipher(unsigned char* c, const unsigned char* m,
```
pohutukawa: Can we name this function `encrypt`, "cipher" can be ambiguous and is less of a common term used for this.

Let's (please) give variables more meaningful names than just `c`, `m`, etc. Even if the library in use does so ... It's way too difficult to understand code otherwise

include/mega/crypto/sodium.h
```diff
@@ -95,6 +95,85 @@ class MEGA_API EdDSA
...
+     * @param c Ciphered text to be decrypted. Caller must ensure that the first
+     * crypto_box_ZEROBYTES bytes of the chipered text c are all 0.
+     * @param n Number used once. The same nonce must never be used to encrypt another
+     * packet from the sender's private key to this receiver's public key or viceversa.
+     * @param pubKey Public key of the sender.
+     * @param privKey Private key of the receiver.
+     *
+     * @return 1 on success, 0 on failure.
+     */
+    int uncipher(unsigned char* m, const unsigned char* c,
```
pohutukawa: Similarly here, name it `decrypt` and change the parameter names.

include/mega/crypto/sodium.h
```diff
@@ -59,18 +57,20 @@ class MEGA_API EdDSA
      * @param msg The message to sign.
      * @param msglen Length of the message.
      * @param sig Buffer to take the signature.
+     * @param siglen Size of the buffer to take the signature.
      * @return Number of bytes for signed message (msg length + signature),
      *     0 on failure.
      */
-    int sign(unsigned char* msg, unsigned long long msglen, char* sig);
+    int sign(const unsigned char* m, const unsigned long long mlen,
```
pohutukawa: Can we please use more meaningful parameter/variable names? Also, the parameters don't match those in the Doxygen string above.

include/mega/megaclient.h
```diff
@@ -77,15 +77,12 @@ class MEGA_API MegaClient
...
     void setkeypair();
 
-    /**
-     * @brief Initialises the Ed25519 EdDSA key user properties.
-     *
-     * A key pair will be added, if not present, yet.
-     *
-     * @return Error code (default: 1 on success).
-     */
+    // initialises the Ed25519 EdDSA key user properties
```
pohutukawa: This is the function signature, why has the Doxygen string been removed here?

It (A) should have a docstring, and (B) parameters/return values should be documented.

include/mega/megaclient.h
```diff
@@ -77,15 +77,12 @@ class MEGA_API MegaClient
...
     int inited25519();
 
+    // initialises the x25519 key user properties
+    int initx25519();
```
pohutukawa: Same as above.

include/mega/megaclient.h
```diff
@@ -752,8 +749,12 @@ class MEGA_API MegaClient
     AsymmCipher asymkey;
 
 #ifdef USE_SODIUM
-    /// EdDSA signing key (Ed25519 privte key seed).
+
+    /// EdDSA signing key (Ed25519 private key seed).
     EdDSA signkey;
+
+    /// ECDH key (x25519 private key seed).
```
pohutukawa: For x25519 (Curve25519) this *is* the key, not a seed (only for Ed25519 it's a seed)

src/commands.cpp
```diff
@@ -1286,6 +1286,10 @@ void CommandLogin::procresult()
                     client->sessionkey.assign((const char *)sek, sizeof(sek));
                 }
 
+                // retrieve signing and chat keys to force their creation if doesn't exist
+                client->getua(client->finduser(me,1), "+puEd255");
```
pohutukawa: Would be good to abstract this to a generic `getUserAttribute` function that handles en/decryption transparently for private attributes as well as public attributes

src/commands.cpp
```diff
@@ -1934,38 +1938,64 @@ void CommandGetUA::procresult()
...
         {
-            // We apparently don't have Ed25519 keys, yet. Let's make 'em.
-            if(!client->inited25519())
-            {
-                return(client->app->getua_result(API_EINTERNAL));
-            }
 
-            // Return the required key data.
-            if (strncmp(attributename.c_str(), "prEd255", 7))
+            if (attributename == "*prEd255" || attributename == "+puEd255")
```
pohutukawa: Now I know where the naming of `prEd255` comes from. That was way early design, before we decided together with Mathias to store private attributes in containers/collections, as the crypto overhead is too big for many little private attributes. Sorry for that, the whole libsodium stuff is pretty outdated.

Michael Holmwood was working on that, but unfortunately it never got fully finished/merged :-(

src/crypto/sodium.cpp
```diff
@@ -107,52 +113,141 @@ int EdDSA::verify(const unsigned char* msg, unsigned long long msglen,
...
     }
-    return(1);
+
+    return 1;
 }
 
 
 // Derives the Ed25519 public key from the stored private key seed.
-int EdDSA::publicKey(unsigned char* pubKey) {
+int EdDSA::publicKey(unsigned char* pubKey)
```
pohutukawa: It might be a good idea to keep the public key in memory (private attribute of the class), and only derive it from the seed if not present, yet. This is a somewhat expensive operation.

src/megaapi_impl.cpp
```diff
@@ -2609,11 +2609,6 @@ void MegaApiImpl::addEntropy(char *data, unsigned int size)
     if(PrnGen::rng.CanIncorporateEntropy())
```
pohutukawa: It's a good idea to *always* put curly brackets around all conditional blocks, even if they're just a single line.

src/megaapi_impl.cpp
```diff
@@ -8726,20 +8721,7 @@ void MegaApiImpl::sendPendingRequests()
...
-                    break;
-                }
-
-                case MegaApi::USER_ATTR_LAST_INTERACTION:
-                {
-                    attrname = "*!lstint";
-                    attrvalue = value;
-                    break;
-                }
+                // authentication ring and last interaction are ro attributes
```
pohutukawa: Why is that? They can be (and are) modified frequently. At least on the webclient.

src/megaclient.cpp
```diff
@@ -6762,14 +6762,49 @@ int MegaClient::inited25519()
...
+        return 0;
     }
 
-    // Store the key pair to user attributes.
-    putua("prEd255", (const byte*)signkey.keySeed, crypto_sign_SEEDBYTES, 1);
-    putua("puEd255", (const byte*)pubKey, crypto_sign_PUBLICKEYBYTES, 0);
+    // Store the key pair to user attributes (skipping the procresult())
+    int creqtag = reqtag;
+    reqtag = 0;
+    putua("*prEd255", (const byte*)signkey.keySeed, crypto_sign_SEEDBYTES);
```
pohutukawa: This is the one that will go terribly wrong. If it's a private attribute (and prefixed with `*`), it *MUST* be TLV-encoded before storage, as it can contain a (hash) map of multiple elements.

src/megaclient.cpp
```diff
@@ -6762,14 +6762,49 @@ int MegaClient::inited25519()
...
+    if(!chatkey.genKeys())
+    {
+        LOG_err << "Error creating the x25519 key pair";
+        return 0;
+    }
+
+    // Store the key pair to user attributes (skipping the procresult())
+    int creqtag = reqtag;
+    reqtag = 0;
+    putua("*prCu255", (const byte*)chatkey.privateKey(), crypto_box_SECRETKEYBYTES);
```
pohutukawa: Same as above, goes into `*keyring`.

pohutukawa:
For further context, look at the comments I have provided in the #crypto channel on Slack.

include/mega/megaclient.h
sergiohs84: Doxygen string was duplicated, both in header and source file. For readability, we keep a different style of comments in header files, so I kept your doxygen string in source file.
Anyway, this function does not accept any parameter :)

src/crypto/sodium.cpp
sergiohs84: Good idea! Then, in this function, we can check if the private key already exists and, instead of deriving the public key (and private key) from the seed, do it directly from the private key: `crypto_sign_ed25519_sk_to_pk()`

include/mega/crypto/sodium.h
```diff
@@ -78,22 +73,90 @@ class MEGA_API EdDSA
...
+     * @param pubKey Public key of the receiver.
+     * @param privKey Private key of the sender.
      *
-     * @param pubKey Public key.
      * @return 1 on success, 0 on failure.
      */
-    int publicKey(unsigned char* pubKey);
-};
+    int encrypt(unsigned char* encmsg, const unsigned char* msg,
+               const unsigned long long msglen, const unsigned char* nounce,
```
pohutukawa: It's a `nonce`, not a `nounce`:
https://en.wikipedia.org/wiki/Cryptographic_nonce

include/mega/crypto/sodium.h
```diff
@@ -78,22 +73,90 @@ class MEGA_API EdDSA
...
+     * @param encmsglen Length of the encrypted text.
+     * @param nounce Number used once. The same nonce must never be used to encrypt another
+     * packet from the sender's private key to this receiver's public key or viceversa.
+     * @param pubKey Public key of the sender.
+     * @param privKey Private key of the receiver.
+     *
+     * @return 1 on success, 0 on failure.
+     */
+    int decrypt(unsigned char* msg, const unsigned char* encmsg,
+                 const unsigned long long encmsglen, const unsigned char* nounce,
```
pohutukawa: `nonce`

include/mega/utils.h
```diff
@@ -222,6 +222,63 @@ struct MEGA_API MemAccess
...
+    /**
+     * @brief containerToTLVrecords Builds a TLV object with records from a container
+     * @param data Binary byte array representing the TLV records
+     * @param datalen Length of the byte array.
+     * @return A new TLVstore object. You take the ownership of the object.
+     */
+    static TLVstore * containerToTLVrecords(const string data);
+
+
+    string TLVrecordsToContainer(SymmCipher *key, encryptionmode_t mode = AES_CCM_12_16);
```
pohutukawa: Let's make `GCM_12_16` the default

include/mega/utils.h
```diff
@@ -222,6 +222,63 @@ struct MEGA_API MemAccess
...
+
+    string TLVrecordsToContainer();
+
+    /**
+     * @brief get Get the value for a given key
+     * @param type Type of the value (without scope nor non-historic modifiers).
+     * @return String containing the array with the value, or NULL if error.
+     */
+    string get(string type)   { return tlv.at(type); }
```
pohutukawa: doesn't our coding style forbid inlining blocks?

src/commands.cpp
```diff
@@ -1974,60 +1988,178 @@ void CommandGetUA::procresult()
...
+            }
 
-            // Is the data a multiple of the cipher blocksize, then we're using
-            // a zero IV.
-            if (l % client->key.BLOCKSIZE == 0)
+            if (attributename == "*keyring")
             {
-                if (!PaddedCBC::decrypt(&d, &client->key))
+                // Ed25519 not initialized yet
+                if (!client->signkey)
```
pohutukawa: This looks like it's only executed if the local value `signkey` is still missing. Shouldn't it be executed if the keyring contains a `signkey`? Similar with the `chatkey` below ...

src/commands.cpp
```diff
@@ -1974,60 +1988,178 @@ void CommandGetUA::procresult()
...
             }
+
+            client->app->getua_result(tlvRecords);
+            delete tlvRecords;
+
+            break;
+
+        case '+':   // public
+
+            // legacy public attributes: '+' prefix but not TLV
```
pohutukawa: public attributes are **never** by default TLV encoded. So none of these is legacy. So, even `+a` should not be legacy, as it's just containing binary avatar content. The other ones are **definitely* not legacy and fully well behaved.

It's **only** private attributes that are TLV encoded, and are in need of automatic TLV unwrapping.

src/commands.cpp
```diff
@@ -1974,60 +1988,178 @@ void CommandGetUA::procresult()
...
-        {
-            client->app->getua_result(data, l);
+            break;
+
+        case '#':   // protected
+            client->app->getua_result(data, datalen);
+            break;
+
+        default:    // legacy private attributes or unknown attribute
+            if (attributename == "firstname"    ||
```
pohutukawa: according to your notes on the wiki, `firstname` and `lastname` are protected:
https://wiki.developers.mega.co.nz/MegaHttpApi/UserAttributes

src/crypto/sodium.cpp
```diff
@@ -107,52 +121,145 @@ int EdDSA::verify(const unsigned char* msg, unsigned long long msglen,
...
-// Generates a new Ed25519 private key seed. The key seed is stored in the object.
-int EdDSA::genKeySeed(unsigned char* privKey) {
-    // Make space for a new key seed (if not present).
-    if (!this->keySeed) {
-        this->keySeed = (unsigned char*)malloc(crypto_sign_SEEDBYTES);
-        if (this->keySeed == NULL) {
+// Derives the Ed25519 private and public keys from the stored private key seed.
+int EdDSA::genKeys()
+{
+    int check = 1;  // error code != 0
```
pohutukawa: What's wrong with using the C++ data type `bool` for this? It always seems safer to me than to resort to a signed integer.

src/megaclient.cpp
```diff
@@ -6008,8 +6008,9 @@ error MegaClient::invite(const char* email, visibility_t show)
  * Attributes are stored as base64-encoded binary blobs. They use internal
  * attribute name prefixes:
  *
- * "*" - Private and CBC-encrypted.
- * "+" - Public and plain text.
+ * "*" - Private and CBC-encrypted. Use a TLV container (key-value)
```
pohutukawa: AES-CCM or -GCM encrypted, not CBC (cipher block chaining is used for other simple keys only, but user attributes use CCM or GCM)

src/utils.cpp
```diff
@@ -360,4 +360,257 @@ int mega_snprintf(char *s, size_t n, const char *format, ...)
...
+    case AES_GCM_12_16_BROKEN:
+        ivlen = 12;
+        taglen = 16;
+        useCCM = true;
+        break;
+
+    case AES_CCM_10_16:
+        ivlen = 10;
+        taglen = 16;
+        useCCM = true;
```
pohutukawa: this seems quite "binary". What if we were to use not AES-CCM and AES-GCM at some point, but also offer Salsa20 with Poly1305? But then I guess we can refactor ... :-)

pohutukawa:
Looks mostly pretty clean (I love clean code :-) ).

But there are a few things that should need addressing. (See notes added)

include/mega/utils.h
sergiohs84: I don't see any mention on that regard in the wiki (https://wiki.developers.mega.co.nz/SoftwareEngineering/CppCodingStyle) nor in the config file for N'SIQ CppStyle (https://github.com/meganz/sdk/blob/master/contrib/nsiq_filefilter.txt), but I'll move them to the source file :)

src/commands.cpp
sergiohs84: OK, I've touched that code to check received private keys in the keyring, even if the SDK has its own ones already set. If any private key doesn't match, it's set to the received value and it updates the corresponding public key.

src/commands.cpp
sergiohs84: Oh, I don't know where I got that idea from :( Fixing

src/commands.cpp
sergiohs84: True, amending the comments

src/crypto/sodium.cpp
sergiohs84: Documentation of libsodium: 
```int	crypto_sign_seed_keypair(unsigned char *pk, unsigned	 char *sk, const unsigned char	 *seed);``` They could add different values we may need to filter.

Comment (sergiohs84) [inline src/crypto/sodium.cpp]: Anyway, this is legacy code from Michael Holmwood branch.

File: src/megaclient.cpp (hunk 22 above)
Comment (sergiohs84) [inline src/megaclient.cpp]: Yep, legacy code again

File: src/utils.cpp (hunk 23 above)
Comment (sergiohs84) [inline src/utils.cpp]: Yes, if we start using a new mode, then we also need to add the new definition for such mode, and then I'll prepare a more generic solution.

File: src/commands.cpp [hunk 24]
```diff
@@ -2003,60 +2017,211 @@ void CommandGetUA::procresult()
...
+                    client->signkey = new EdDSA();
+                    signkeyUpdated = true;
+                }
+
+                // x25519 sanity checkup
+                if (tlvRecords->find(ECDH::TLV_KEY))
+                {
+                    tmpstr = tlvRecords->get(ECDH::TLV_KEY);
+
+                    // check if received private key seed is different from existing one
```
pohutukawa: for x25519 it's not a "key seed" but really actually a "private key". But that's a technicality :-)

src/commands.cpp
```diff
@@ -2003,60 +2017,211 @@ void CommandGetUA::procresult()
...
+            if (!tlvRecords)
             {
-                if (!PaddedCBC::decrypt(&d, &client->key))
+                LOG_err << "Cannot extract TLV records for private attribute " << attributename;
+                client->app->getua_result(API_EINTERNAL);
+                delete [] data;
+                return;
+            }
+
+            if (attributename == "*keyring")
```
pohutukawa: I think it might be a good idea to extract processing for certain private and public attributes that require special treatment (keyring, pub keys, signatures) for sanity checks/"healing" to an external function to not inflate and overcomplicate this function.
But that's more an issue for further refactoring, the procedure here seems good in general.

src/crypto/sodium.cpp
pohutukawa: All good.

src/commands.cpp
```diff
@@ -1958,43 +1958,57 @@ CommandGetUA::CommandGetUA(MegaClient* client, const char* uid, const char* an)
...
 
 #ifdef USE_SODIUM
-        if ((e == API_ENOENT) && (user->userhandle == client->me)
-                && ((priv && strncmp(attributename.c_str(), "prEd255", 7))
-                        || (!priv && strncmp(attributename.c_str(), "puEd255", 7))))
+        if ((e == API_ENOENT) && (user->userhandle == client->me))
         {
-            // We apparently don't have Ed25519 keys, yet. Let's make 'em.
-            if(!client->inited25519())
+            // If keyring is not found, create keypairs and send them to MEGA
```
pohutukawa: I think it might be a good idea to extract processing for certain private and public attributes that require special treatment (keyring, pub keys, signatures) for sanity checks/"healing" to an external function to not inflate and overcomplicate this function.
But that's more an issue for further refactoring, the procedure here seems good in general.

src/megaapi_impl.cpp
sergiohs84: Currently, no SDK client is using the last interaction value or the authentication ring (yet). I suggest to keep those changes for a future branch and to release this code to generate the required keys for chat without lstint and authring.
In any case, the missing part in the SDK is the ability to set those attributes, but not to read them. Every private attribute is TLV-decoded after decryption (`CommandGetUA::procresult()`)

include/mega/utils.h
```diff
@@ -222,6 +222,90 @@ struct MEGA_API MemAccess
...
+     */
+    static TLVstore * containerToTLVrecords(const string *data);
+
+    /**
+     * @brief Converts the TLV records into an encrypted byte array
+     * @param key Master key to decrypt the container
+     * @param mode Block encryption mode to be used by AES
+     * @return A new string holding the encrypted byte array. You take the ownership of the string.
+     */
+    string *TLVrecordsToContainer(SymmCipher *key, encryptionmode_t mode = AES_GCM_12_16);
```
pohutukawa: I don't like a function starting with an upper case letter. I just thought this was the constructor of a class ...

Methods and functions should *always* start with a lower case letter!

src/commands.cpp
```diff
@@ -2003,77 +2018,219 @@ void CommandGetUA::procresult()
...
+        unsigned int datalen;
         byte* data;
-        l = (end - ptr) / 4 * 3 + 3;
-        data = new byte[l];
-        l = Base64::atob(ptr, data, l);
+        datalen = (end - ptr) / 4 * 3 + 3;
+        data = new byte[datalen];
+        datalen = Base64::atob(ptr, data, datalen);
+
+//        bool nonHistoric = (attributename.at(1) == '!');
```
pohutukawa: Do we need this line? Should be possible to remove it, right?

src/utils.cpp
```diff
@@ -360,4 +361,328 @@ int mega_snprintf(char *s, size_t n, const char *format, ...)
...
+    string *container = TLVrecordsToContainer();
+
+    // generate IV array
+    byte *iv = new byte[ivlen];
+    PrnGen::genblock(iv, ivlen);
+
+    string cipherText;
+
+    // encrypt the bytes using the specified mode
+    if (encMode == 1)   // CCM or GCM_BROKEN (same than CCM)
```
pohutukawa: I'd prefer to use a named constant here, not a literal constant (e. g. use `CCM` or whichever it's in this code, and not `1`)

src/utils.cpp
```diff
@@ -360,4 +361,328 @@ int mega_snprintf(char *s, size_t n, const char *format, ...)
...
+    {
+        key->ccm_encrypt(container, iv, ivlen, taglen, &cipherText);
+    }
+    else if (encMode == 2)   // then use GCM
```
pohutukawa: Same here, preferably a constant `GCM` (or as appropriate).

src/utils.cpp
```diff
@@ -360,4 +361,328 @@ int mega_snprintf(char *s, size_t n, const char *format, ...)
...
+    memcpy(iv, &(data->data()[offset]), ivlen);
+    offset += ivlen;
+
+    unsigned cipherTextLen = data->length() - offset;
+    string cipherText = data->substr(offset, cipherTextLen);
+
+    unsigned clearTextLen = cipherTextLen - taglen;
+    string clearText;
+
+    if (encMode == 1)   // CCM or GCM_BROKEN (same than CCM)
```
pohutukawa: Same with mode constants above: e. g. `CCM_MODE`

pohutukawa:
Looks in general pretty good. I have not tested the logic directly, but had a look at the general code. (I was hoping that there were unit tests to test the accuracy of the logic. :-( ).

Besides the comments, it looks like this branch needs merging with `master` first to resolve some merge conflicts.

src/commands.cpp
sergiohs84: Yes, right now it's not going to be used. But I'd like to keep it there for reference and for a near future where SDK might need to adapt the behaviour based on the historical property of attributes.

tests/tests.cpp
```diff
@@ -49,6 +51,102 @@ TEST(Serialize64, serialize) {
     ASSERT_EQ(in, out);
 }
 
+// Test encryption/decryption using AES in mode GCM
+// (test vectors from 'tlvstore_test.js', in Webclient)
+TEST(CryptoPP, AES_GCM)
+{
+//    string keyStr = "dGQhii+B7+eLLHRiOA690w==";   // Base64
```
pohutukawa: Please, kick this commented line out.

sergiohs84: It's a reminder to remark why the key is not the same than in `tlvstore_test.js`, where Base64URL encoding is not used, as it happens in the SDK. I'd prefer to keep it if you don't mind.

sergiohs84:
This branch has become obsolete. All its functionality is included in the `chat-attributes` branch. You can safely checkout that branch instead.

PR #188, opened 2015-12-01
Caching of names + Fix serialization of Localnodes

sergiohs84:


src/megaclient.cpp
```diff
@@ -6078,6 +6082,18 @@ void MegaClient::getua(User* u, const char* an)
 {
     if (an)
     {
+        // if we can solve those requests locally (runtime cached values)...
+        if (!strcmp(an, "firstname") && u->firstname)
+        {
+            app->getua_result((byte*) u->firstname->data(), u->firstname->size());
```
javiserrano: client->restag should be set before calling callbacks

PR #192, opened 2015-12-11
Chat commands

alxvasilev:


src/megaapi_impl.cpp
```diff
@@ -1452,12 +1452,9 @@ MegaRequestPrivate::MegaRequestPrivate(MegaRequestPrivate *request)
         *(this->chatPeerList) = *(request->getMegaTextChatPeerList());
     }
 
-    this->chatList = NULL;
-    if(request->getMegaTextChatPeerList())
-    {
-        this->chatList = new MegaTextChatListPrivate;
-        *(this->chatList) = *(request->getMegaTextChatList());
```
alxvasilev: Seems this does not do copying, I get empty chat list

PR #193, opened 2015-12-15
Chat commands

sergiohs84:
- Addition of new commands: `mcc`, `mcf`, `mci`, `mcr` and `mcurl`
 - Addition of public interfaces in the intermediate layer to: create new chat rooms, fetch chats, invite a new peer, remove an existing peer and retrieve the user-specific url for the chat.
 - New request types for the added methods: `TYPE_CHAT_CREATE`, `TYPE_CHAT_FETCH`...
 - New classes: `MegaTextChat`, `MegaTextChatList` and `MegaTextChatPeerList`
 - New callbacks: `MegaApp::chatfecth_result()`, `chatcreate_result()` and so on
 - New notification: `MegaApp::chats_updated()`


examples/megacli.cpp
```diff
@@ -2795,6 +2958,173 @@ static void process_line(char* l)
...
+                                    cout << "Unknown privilege for " << email << endl;
+                                    delete userpriv;
+                                    return;
+                                }
+
+                                userpriv->push_back(userpriv_pair(u->userhandle, priv));
+                                numUsers++;
+                            }
+
+                            client->createChat(group, userpriv);
```
javiserrano: Where is `userpriv` deleted?

src/megaclient.cpp
```diff
@@ -8731,5 +8859,89 @@ void MegaClient::cleanrubbishbin()
     reqs.add(new CommandCleanRubbishBin(this));
 }
 
+#ifdef ENABLE_CHAT
+void MegaClient::createChat(bool group, userpriv_vector *userpriv)
```
javiserrano: Related to the previous comment. I can't see who is the owner of `userpriv`. On most functions, the SDK doesn't take the ownership of input parameters. However, this vector seems to be stored in `CommandChatCreate` and, depending on the result, it's propagated to `chatcreate_result` inside at `TextChat` object. I can't see a deletion on this vector nor in `megacli` nor in the intermediate layer.

src/commands.cpp
```diff
@@ -3464,5 +3464,351 @@ void CommandCleanRubbishBin::procresult()
...
+        {
+            client->app->chatfetch_result(NULL, e);
+        }
+
+        // clean allocated memory
+        TextChat *chat;
+        for (unsigned i = 0; i < chatlist->size(); i++)
+        {
+            chat = chatlist->at(i);
+            delete chat->userpriv;
```
javiserrano: It seems that there is a missing `delete chat;` here.

PR #205, opened 2016-01-08
HTTP proxy server and many enhancements for streaming downloads

javiserrano:


include/megaapi.h
```diff
@@ -6729,6 +6732,203 @@ class MegaApi
...
+         * @param port Port in which the server must accept connections
+         * @return True is the server is ready, false if the initialization failed
+         */
+        bool httpServerStart(bool localOnly = true, int port = 4443);
+
+        /**
+         * @brief Stop the HTTP proxy server
+         *
+         * When this function returns, the server is already shutdown.
+         * If the HTTP proxy server isn't running, this functions doesn't do anything.
```
sergiohs84: I'd rather say: `...this function does nothing.

include/megaapi.h
```diff
@@ -6729,6 +6732,203 @@ class MegaApi
...
+         */
+        void httpServerStop();
+
+        /**
+         * @brief Check if the HTTP proxy server is running
+         * @return 0 if the server is not running, otherwise the port in which it's listening
```
sergiohs84: @return 0 if the server is not running. Otherwise, the port in which it's listening to

include/megaapi.h
```diff
@@ -6729,6 +6732,203 @@ class MegaApi
...
+         * if this buffer is small the streaming can have problems due to the overhead caused by
+         * the excessive number of POST requests.
+         *
+         * It's recommended to set this buffer at least to 1MB
+         *
+         * For connections that request less data than the buffer size, the HTTP proxy server
+         * will only allocate the required memory to complete the request to minimize the
+         * memory usage.
+         *
+         * The new value will be takein into account since the next request received by
```
sergiohs84: `...will be takein into account...` --> `...will be taken into account...`

src/base64.cpp
```diff
@@ -297,4 +297,82 @@ int Base32::atob(const char *a, byte *b, int blen)
...
+        char c = plain->at(i);
+        if (issafe(c))
+        {
+            escaped->push_back(c);
+            escapedIndex++;
+        }
+        else
+        {
+            escaped->resize(escapedIndex + 3);
+            sprintf((char *)escaped->c_str() + escapedIndex, "%%%02x", c);
```
sergiohs84: Not sure if it's your intention, but `escaped->c_str()` will add one extra char (`'\0'`) at the end of the string.

cocia:
This is great feature. I tried to implement similar feature with current android mega sdk. 

How can I use this feature in android/java mega sdk?
I checked this httpserver code and compiled for android. I could see the httpserver api methods were in c++ version sdk but not in android/java version sdk.

javiserrano:
We haven't added those functions yet to the Android/Java bindings.

The only changes that should be required to use them are:
- Build and link `libuv` in the with the native library (like cURL, OpenSSL, etc.)
For that, it will be required to modify the build scripts accordingly.
- Define `HAVE_LIBUV` during the construction of the native library, and also in `bindings/megaapi.i`
- Add the new functions in `MegaApiJava`. It's just needed to forward them to `MegaApi`

We plan to do this soon to use this feature in our official Android app.

cocia:
@javiserrano 
I already can compile mega sdk with libuv. If I'm a swig or api expert I will define all of these for android by myself. But I'm not. 
I have seen this http server streaming api is ready for c++ and iOS version. I'm eagerly to use this streaming feature in Android version. 
Do you have plan to add this feature to android binding? or When?

PR #223, opened 2016-02-09
Fixed a possible deadlock in WP bindings

javiserrano:


javiergm1983:
Reviewed and tested. All OK.

PR #226, opened 2016-02-15
Stop the sync when nodes can't be attached to the account

javiserrano:


sergiohs84:
All good, ready to merge.

PR #234, opened 2016-03-01
Chat attributes and keys

sergiohs84:
- Private user attributes with the new TLV format.
- Versioning of user attributes (`v` parameter and the new `upv` command to put attributes).
- Caching of user attributes with versioning capability.
- Creation of keypairs for signing and chat.
- Sanity checkups + healing procedure for keypairs.
- New dependency: libsodium.
- New encryption modes for AES: CCM and GCM


bindings/qt/sdk.pri
```diff
@@ -216,12 +218,18 @@ unix:!macx {
    LIBS += -lsqlite3 -lrt
 
    exists($$MEGASDK_BASE_PATH/bindings/qt/3rdparty/libs/libcurl.a) {
-    INCLUDEPATH += $$MEGASDK_BASE_PATH/bindings/qt/3rdparty/include
```
sergiohs84: It's included for every platform, so for linux the include paht is redundant.

src/commands.cpp
```diff
@@ -1959,50 +1959,87 @@ void CommandUserRequest::procresult()
...
 
-    // if removing avatar, do not Base64 encode the attribute value
-    if (!strcmp(an, "+a") && !strcmp((const char *)av, "none"))
-    {
-        arg(an,(const char *)av, avl);
-    }
-    else
+    cmd("upv");
+
+    beginarray(an);
```
sergiohs84: The new syntax allows to send several attribute values at once, but SDK doesn't implement such a capability.
Should we add it? It would be useful for very couple attributes, like setting `keyring` + `puCu255` + `puEd255` all at once. It will make the request fail if any of the values cannot be set (ie. outdated version of any of them).

src/commands.cpp
```diff
@@ -1959,50 +1959,87 @@ void CommandUserRequest::procresult()
...
+        string an = string(ptr, (end-ptr));
+
+        if (!(ptr = client->json.getvalue()) || !(end = strchr(ptr, '"')))
+        {
+            client->app->putua_result(API_EINTERNAL);
+            return;
+        }
+        string v = string(ptr, (end-ptr));
+
+        if (an.empty() || v.empty() || (an != an))
```
sergiohs84: In the same way than previous comment, the response of `upv` can return more than one pair attribute_name-attribute_version. Since the SDK doesn't send multiple attributes at once, this `procresult()` only process one attribute.

src/megaapi.cpp
```diff
@@ -1360,6 +1365,11 @@ void MegaApi::setUserAttribute(int type, const char *value, MegaRequestListener
     pImpl->setUserAttribute(type, value, listener);
 }
 
+void MegaApi::setUserAttribute(int type, const MegaStringMap *value, MegaRequestListener *listener)
```
sergiohs84: For private attributes, the required value is a `MegaStringMap`

src/megaapi_impl.cpp
```diff
@@ -2745,18 +2871,84 @@ void MegaApiImpl::retryPendingConnections(bool disconnect, bool includexfers, Me
 void MegaApiImpl::addEntropy(char *data, unsigned int size)
 {
     if(PrnGen::rng.CanIncorporateEntropy())
+    {
         PrnGen::rng.IncorporateEntropy((const byte*)data, size);
-
-#ifdef USE_SODIUM
-    if(EdDSA::rng.CanIncorporateEntropy())
-        EdDSA::rng.IncorporateEntropy((const byte*)data, size);
-#endif
```
sergiohs84: Do not a new random number generator, but reuse the existing one.

src/megaapi_impl.cpp
```diff
@@ -2745,18 +2871,84 @@ void MegaApiImpl::retryPendingConnections(bool disconnect, bool includexfers, Me
...
+    {
+        case MegaApi::USER_ATTR_AVATAR:
+        case MegaApi::USER_ATTR_ED25519_PUBLIC_KEY:
+        case MegaApi::USER_ATTR_CU25519_PUBLIC_KEY:
+            scope = '+';
+            break;
+
+        case MegaApi::USER_ATTR_FIRSTNAME:
+        case MegaApi::USER_ATTR_LASTNAME:
+            scope = '0';
```
sergiohs84: Firstname / lastname are indeed legacy protected attributes and do not have scope prefix, but need to make difference between them.

src/megaclient.cpp
```diff
@@ -6238,50 +6271,41 @@ error MegaClient::invite(const char* email, visibility_t show)
...
  * @param avl Attribute value length.
+ * @param ctag Tag to identify the request at intermediate layer
  * @return Void.
  */
-void MegaClient::putua(const char* an, const byte* av, unsigned avl)
+void MegaClient::putua(const char* an, const byte* av, unsigned avl, int ctag)
 {
     string data;
 
-    if (!strcmp(an, "*!lstint") || !strcmp(an, "*!authring"))
```
sergiohs84: This block was never used and `lstint` / `*!authring` are not handled that way.

PR #236, opened 2016-03-07
Copy-delete instead of move when it's not possible

javiserrano:
This happens when a node is moved to an incoming share


sergiohs84:
All good after fixing conditions. Ready to merge :)

PR #238, opened 2016-03-08
Import public links in outshares

javiserrano:


sergiohs84:
All good! Waiting for deployment into production servers

PR #249, opened 2016-03-29
Changes related to the authentication with API servers

javiserrano:
They are mainly intended to take advantage of the PRO status when nodes are downloaded from the webclient, public folders or public links.

### DIFF
--- a/src/megaapi_impl.cpp
+++ b/src/megaapi_impl.cpp
@@ -70,7 +70,7 @@
 
 using namespace mega;
 
-MegaNodePrivate::MegaNodePrivate(const char *name, int type, int64_t size, int64_t ctime, int64_t mtime, uint64_t nodehandle, string *nodekey, string *attrstring, const char *fingerprint, MegaHandle parentHandle, const char*auth)
+MegaNodePrivate::MegaNodePrivate(const char *name, int type, int64_t size, int64_t ctime, int64_t mtime, uint64_t nodehandle, string *nodekey, string *attrstring, const char *fingerprint, MegaHandle parentHandle, const char *privateauth, const char *publicauth)
 : MegaNode()
 {
     this->name = MegaApi::strdup(name);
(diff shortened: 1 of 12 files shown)


src/megaapi_impl.cpp
```diff
@@ -431,7 +439,8 @@ MegaNode* MegaNodePrivate::getPublicNode()
 
     MegaNode *node = new MegaNodePrivate(
                 name, type, size, ctime, mtime,
-                plink->ph, &key, &attrstring, fingerprint);
+                plink->ph, &key, &attrstring, fingerprint,
+                INVALID_HANDLE);
```
sergiohs84: `INVALID_HANDLE` is the default value of this constructor. Specifying it here is redundant.

src/megaapi_impl.cpp
```diff
@@ -2863,6 +2894,41 @@ char *MegaApiImpl::dumpXMPPSession()
...
+    }
+
+    sdkMutex.unlock();
+    return buf;
+}
+
+void MegaApiImpl::setAccountAuth(const char *auth)
+{
+    sdkMutex.lock();
+    if (!auth)
```
sergiohs84: We could protect this method against empty strings like in `MegaNodePrivate::setPrivateAuth()` by changing the condition by: `if (!auth || !auth[0])`

src/megaapi_impl.cpp
```diff
@@ -7372,7 +7439,8 @@ void MegaApiImpl::openfilelink_result(handle ph, const byte* key, m_off_t size,
 	else
 	{
         request->setPublicNode(new MegaNodePrivate(fileName.c_str(), FILENODE, size, 0, mtime, ph, &keystring, a,
-                                                   fingerprint.size() ? fingerprint.c_str() : NULL));
+                                                   fingerprint.size() ? fingerprint.c_str() : NULL,
+                                                   INVALID_HANDLE));
```
sergiohs84: `INVALID_HANDLE` is the default value of this constructor. Specifying it here is redundant.

PR #261, opened 2016-04-11
Release branch for MEGAsync

javiserrano:


sergiohs84:
Apart from the `isForeign : 1` in the `MegaUser`, all seems correct.

PR #272, opened 2016-04-25
Fix ephemereal user

sergiohs84:
`megacli` creates an ephemereal account as the first step to create a
full account.

Now, the API returns `"m":"none"` for the email of the ephereal user. It
makes the SDK to store the email and identify the account as confirmed,
instead of ephemereal, blocking the signup process from the SDK (in
`megacli`).


sergiohs84:
This bug has been solved by the API. It won't return `none`, but an empty string.

PR #281, opened 2016-05-04
Resumption of uploads and downloads

javiserrano:


src/megaclient.cpp
```diff
@@ -8970,9 +9190,44 @@ bool MegaClient::startxfer(direction_t d, File* f, bool skipdupes)
...
         {
-            t = new Transfer(this, d);
-            *(FileFingerprint*)t = *(FileFingerprint*)f;
-            t->size = f->size;
+            it = cachedtransfers[d].find(f);
+            if (it != cachedtransfers[d].end())
+            {
+                LOG_debug << "Resumable transfer detected";
+                FileAccess* fa = fsaccess->newfileaccess();
+                if (fa->fopen(&it->second->localfilename)
```
sergiohs84: Check whether the fingerprint has not changed upon resumption of upload transfers.

src/megaclient.cpp
```diff
@@ -7216,6 +7278,164 @@ bool MegaClient::fetchsc(DbTable* sctable)
...
+            delete it->first;
+            delete it->second;
+            cachedtransfers[d].erase(it);
+        }
+    }
+}
+
+void MegaClient::enabletransferresumption(const char *loggedoutid)
+{
+    if (!dbaccess)
```
sergiohs84: Check if `tctable` is already opened

src/megaclient.cpp
```diff
@@ -7216,6 +7278,164 @@ bool MegaClient::fetchsc(DbTable* sctable)
     return true;
 }
 
+void MegaClient::closetc()
```
sergiohs84: Add a flag to indicate that the local cache is discarded and the temporal files (being downloaded) should be removed.

src/file.cpp
```diff
@@ -46,6 +46,147 @@ File::~File()
...
+    if (ptr + sizeof(unsigned short) > end)
+    {
+        LOG_err << "File unserialization failed - serialized string too short";
+        return NULL;
+    }
+
+    // read name
+    unsigned short namelen = MemAccess::get<unsigned short>(ptr);
+    ptr += sizeof(namelen);
+    if (ptr + namelen + sizeof(unsigned short) > end)
```
sergiohs84: `ptr` is already moved `sizeof(unsigned short)`. I think the `if` clause should be: 
```if (ptr + namelen > end)```

File: src/file.cpp [hunk 5]
```diff
@@ -46,6 +46,147 @@ File::~File()
...
+        LOG_err << "File unserialization failed - name too long";
+        return NULL;
+    }
+    const char *name = ptr;
+    ptr += namelen;
+
+    // read localname
+    unsigned short localnamelen = MemAccess::get<unsigned short>(ptr);
+    ptr += sizeof(localnamelen);
+    if (ptr + localnamelen + sizeof(unsigned short) > end)
```
sergiohs84: Same than above

src/file.cpp
```diff
@@ -46,6 +46,147 @@ File::~File()
...
+        LOG_err << "File unserialization failed - localname too long";
+        return NULL;
+    }
+    const char *localname = ptr;
+    ptr += localnamelen;
+
+    // read private auth
+    unsigned short privauthlen = MemAccess::get<unsigned short>(ptr);
+    ptr += sizeof(privauthlen);
+    if (ptr + privauthlen + sizeof(unsigned short) > end)
```
sergiohs84: Same than above

src/file.cpp
javiserrano: That `sizeof(unsigned short)` is to ensure that the next `unsigned short` that is going to be read (`localnamelen`, few lines below) is available.

src/file.cpp
javiserrano: Same as above :-)

src/file.cpp
javiserrano: Same as above :-D

src/file.cpp
sergiohs84: Oh, I didn't realize. Then, all fine.

sergiohs84:
All fine with this pull-request!

PR #287, opened 2016-05-06
Accept `ESID` as logout result

sergiohs84:
Also removes 2 warnings: unused variables and comparission between
signed and unsigned variables.

### DIFF
--- a/src/commands.cpp
+++ b/src/commands.cpp
@@ -1163,20 +1163,7 @@ void CommandLogout::procresult()
     MegaApp *app = client->app;
     if(!e)
     {
-        if (client->sctable)
-        {
-            client->sctable->remove();
-        }
-
-#ifdef ENABLE_SYNC
-        for (sync_list::iterator it = client->syncs.begin(); it != client->syncs.end(); it++)
-        {
-            if((*it)->statecachetable)
-            {
-                (*it)->statecachetable->remove();
-            }
-        }
-#endif
+        client->removecaches();
         client->locallogout();
     }
     app->logout_result(e);
--- a/src/commands.cpp
+++ b/src/commands.cpp
@@ -1161,7 +1161,7 @@ void CommandLogout::procresult()
 {
     error e = (error)client->json.getint();
     MegaApp *app = client->app;
-    if(!e || e == API_ESID)
+    if(!e)
     {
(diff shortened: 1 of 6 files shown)


sergiohs84:
Do not merge it yet. The `ESID` result can be received at HTTP request level, instead of at Command level, when there's multiple commands in the HTTP request.

sergiohs84:
We might consider whether MegaApi should return API_OK when the result of a logout is ESID??

src/commands.cpp
```diff
@@ -1161,7 +1161,7 @@ void CommandLogout::procresult()
 {
     error e = (error)client->json.getint();
     MegaApp *app = client->app;
-    if(!e)
+    if(!e || e == API_ESID)
```
javiserrano: I think that API_ESID can't be received as a command-level response, only as a request-level response.

src/megaapi_impl.cpp
```diff
@@ -7295,6 +7295,15 @@ void MegaApiImpl::request_error(error e)
         request->setText(client->sslfakeissuer.c_str());
     }
 
+    if (e == API_ESID)
+    {
+        if (client->sctable)
+        {
+            client->sctable->remove();
+        }
```
javiserrano: I think that the cache of syncs should be also removed here like in `CommandLogout::procresult`

src/megaapi_impl.cpp
```diff
@@ -7336,7 +7345,7 @@ void MegaApiImpl::logout_result(error e)
     MegaRequestPrivate* request = requestMap.at(client->restag);
     if(!request || (request->getType() != MegaRequest::TYPE_LOGOUT)) return;
 
-    if(!e)
+    if(!e || e == API_ESID)
```
javiserrano: I have some doutbs in this case, but I think that it's not possible to receive API_ESID here.

src/commands.cpp
sergiohs84: You're right, only at HTTP level.

src/megaapi_impl.cpp
sergiohs84: Since the command doesn't receive the ESID and the SDK doesn't call `logout_result()` with that error, I agree with you.

PR #290, opened 2016-05-09
Ability to remove users from `megacli`

sergiohs84:
Additionally, this pull-requests:

- Updates the API request to remove/block contacts: `ur` --> `ur2`
- Removes the obsolete method to add contacts without invitation.

### DIFF
--- a/include/megaapi.h
+++ b/include/megaapi.h
@@ -1600,7 +1600,7 @@ class MegaRequest
             TYPE_GET_PUBLIC_NODE, TYPE_GET_ATTR_FILE,
             TYPE_SET_ATTR_FILE, TYPE_GET_ATTR_USER,
             TYPE_SET_ATTR_USER, TYPE_RETRY_PENDING_CONNECTIONS,
-            TYPE_ADD_CONTACT, TYPE_REMOVE_CONTACT, TYPE_CREATE_ACCOUNT,
+            TYPE_REMOVE_CONTACT, TYPE_CREATE_ACCOUNT,
             TYPE_CONFIRM_ACCOUNT,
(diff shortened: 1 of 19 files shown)


include/megaapi.h
```diff
@@ -1600,7 +1600,7 @@ class MegaRequest
             TYPE_GET_PUBLIC_NODE, TYPE_GET_ATTR_FILE,
             TYPE_SET_ATTR_FILE, TYPE_GET_ATTR_USER,
             TYPE_SET_ATTR_USER, TYPE_RETRY_PENDING_CONNECTIONS,
-            TYPE_ADD_CONTACT, TYPE_REMOVE_CONTACT, TYPE_CREATE_ACCOUNT,
```
javiserrano: This change requires to update all bindings. Otherwise client apps would have severe problems.
Maybe it's worth to keep that value for compatibility.

include/megaapi.h
```diff
@@ -5250,14 +5250,7 @@ class MegaApi
...
-         * Valid data in the MegaRequest object received on callbacks:
-         * - MegaRequest::getEmail - Returns the email of the contact
-         *
-         * @param email Email of the new contact
-         * @param listener MegaRequestListener to track this request
-         *
-         * @deprecated: This way to add contacts will be removed in future updates. Please use MegaApi::inviteContact.
+         * @obsolete: This way to add contacts is obsolete. Please use MegaApi::inviteContact.
          */
         void addContact(const char* email, MegaRequestListener* listener = NULL);
```
javiserrano: It seems that this function has been removed in `megaapi.cpp`, so it should be removed here too.

PR #291, opened 2016-05-10
Better management of transfer errors

javiserrano:
to not increase the error count while the app is offline


sergiohs84:
All fine here!

PR #292, opened 2016-05-10
Chat keys & attributes

sergiohs84:
- Private user attributes with the new TLV format.
- Versioning of user attributes (`v` parameter and the new `upv` command to put attributes).
- Caching of user attributes with versioning capability.
- Creation of keypairs for signing and chat.
- Sanity checkups + healing procedure for keypairs.
- New dependency: libsodium.
- New encryption modes for AES: CCM and GCM


sergiohs84:
This pull-request to merge into `develop` replaces the previous one (for `master`).
For reference, the old PR was: https://github.com/meganz/sdk/pull/234

src/crypto/sodium.cpp
```diff
@@ -22,138 +22,139 @@
...
-        if (this->keySeed == NULL) {
-            // Something went wrong allocating the memory.
-            return(0);
-        }
-    }
-    // Now make the new key seed.
-    this->rng.GenerateBlock(this->keySeed, crypto_sign_SEEDBYTES);
-    // Copy it to privKey before returning.
-    if (privKey)
+    return (byte*) binaryhash.substr(0, 20).c_str();
```
javiserrano: It's returning a pointer that belongs to a local variable.
It will be invalid after the call to the function.

src/crypto/sodium.cpp
```diff
@@ -22,138 +22,139 @@
...
+    for (size_t i = 0; i < 20; ++i)
     {
-        memcpy(privKey, this->keySeed, crypto_sign_SEEDBYTES);
+        oss.put(hexchars[(fp[i] >> 4) & 0x0F]);
+        oss.put(hexchars[fp[i] & 0x0F]);
     }
-    return(1);
+    fpHex = oss.str();
+
+    return fpHex.c_str();
```
javiserrano: It's returning a pointer that belongs to a local variable.
It will be invalid after the call to the function.

bindings/qt/sdk.pri
```diff
@@ -227,12 +228,20 @@ unix:!macx {
...
 }
 
 macx {
    INCLUDEPATH += $$MEGASDK_BASE_PATH/include/mega/posix
    SOURCES += $$MEGASDK_BASE_PATH/bindings/qt/3rdparty/sqlite3.c
    INCLUDEPATH += $$MEGASDK_BASE_PATH/bindings/qt/3rdparty/include/curl
+   INCLUDEPATH += $$MEGASDK_BASE_PATH/bindings/qt/3rdparty/include/libsodium
    DEFINES += PCRE_STATIC _DARWIN_FEATURE_64_BIT_INODE
-   LIBS += -L$$MEGASDK_BASE_PATH/bindings/qt/3rdparty/libs/ $$MEGASDK_BASE_PATH/bindings/qt/3rdparty/libs/libcares.a $$MEGASDK_BASE_PATH/bindings/qt/3rdparty/libs/libcurl.a -lz -lssl -lcrypto
+   LIBS += -L$$MEGASDK_BASE_PATH/bindings/qt/3rdparty/libs/ $$MEGASDK_BASE_PATH/bindings/qt/3rdparty/libs/libcares.a $$MEGASDK_BASE_PATH/bindings/qt/3rdparty/libs/libcurl.a -lz -lssl -lcrypto -lsodium
```
javiserrano: On OS X, libsodium will be linked as a static library `.a` like `libcurl.a` or `libcares.a`

examples/megacli.cpp
```diff
@@ -3225,7 +3290,7 @@ static void process_line(char* l)
                     }
                     else if (words[0] == "invite")
                     {
-                        if (client->finduser(client->me)->email.compare(words[1]))
+                        if (client->ownuser()->email.compare(words[1]))
```
javiserrano: This would probably crash if the user is not logged in

include/mega/crypto/sodium.h
```diff
@@ -19,7 +19,6 @@
  * program.
  */
 
-#ifdef USE_SODIUM
```
javiserrano: It might be good to use ENABLE_CHAT here to save space on apps that don't want to link libsodium.

include/megaapi.h
```diff
@@ -1376,7 +1488,7 @@ class MegaStringList
      * @param i Position of the string that we want to get for the list
      * @return string at the position i in the list
      */
-    virtual const char* get(int i);
+    virtual const char* get(int);
```
javiserrano: It's needed to add the name of the parameter again to not break the documentation

src/commands.cpp
```diff
@@ -1998,173 +1998,281 @@ void CommandUserRequest::procresult()
...
-    // if removing avatar, do not Base64 encode the attribute value
-    if (!strcmp(an, "+a") && !strcmp((const char *)av, "none"))
-    {
-        arg(an,(const char *)av, avl);
-    }
-    else
+    cmd("upv");
+
+    beginarray(an);
     {
```
javiserrano: It seems that these brackets are no longer needed

src/commands.cpp
```diff
@@ -1998,173 +1998,281 @@ void CommandUserRequest::procresult()
...
+        }
+        else
+        {
+            element(av, avl);
+        }
+
+        const string *attrv = client->ownuser()->getattrversion(an);
+        if (attrv)
+        {
+            element(attrv->c_str());
```
javiserrano: This could confuse servers because the number of elements in the array could change.

src/commands.cpp
```diff
@@ -1998,173 +1998,281 @@ void CommandUserRequest::procresult()
...
-        if (!(ptr = client->json.getvalue()) || !(end = strchr(ptr, '"')))
+        string v;
+        for (;;)
         {
-            return(client->app->getua_result(API_EINTERNAL));
-        }
+            switch (client->json.getnameid())
+            {
+                case MAKENAMEID2('a','v'):
+                    if (!(ptr = client->json.getvalue()) || !(end = strchr(ptr, '"')))
```
javiserrano: It seems that the value of `av` is not used for anything

src/commands.cpp
```diff
@@ -1998,173 +1998,281 @@ void CommandUserRequest::procresult()
...
-        }
-        else if (attributename == "lastname")
-        {
-            if (!user->lastname)
-            {
-                user->lastname = new string;
-            }
-            user->lastname->assign((char*) data, l);
-        }
+                    // bool nonHistoric = (attributename.at(1) == '!');
```
javiserrano: Forgotten comment?

src/commands.cpp
```diff
@@ -1998,173 +1998,281 @@ void CommandUserRequest::procresult()
...
-                string payload;
-                payload.assign(d, 8, l - 8);
-                d = payload;
-                if (!PaddedCBC::decrypt(&d, &client->key, &iv))
-                {
-                    delete[] data;
-                    client->app->getua_result(API_EINTERNAL);
+                    if (!u) // retrieval of attributes without contact-relationship
+                    {
+                        if (scope == '*' || scope == '#')
```
javiserrano: This could be probably checked before sending the command to the server.

src/megaclient.cpp
```diff
@@ -7601,10 +7627,221 @@ void MegaClient::fetchnodes()
...
+        if ((puEd255.size() != EdDSA::PUBLIC_KEY_LENGTH) || memcmp(puEd255.data(), signkey->pubKey, EdDSA::PUBLIC_KEY_LENGTH))
+        {
+            LOG_warn << "Public key for Ed25519 mismatch.";
+
+            int creqtag = reqtag;
+            reqtag = 0;
+            sendevent(99411, "Ed25519 public key mismatch");
+            reqtag = creqtag;
+
+            fetchingkeys = false;
```
javiserrano: It seems that `fetchingKeys = false` could be put inside `clearKeys()`

src/megaclient.cpp
```diff
@@ -7601,10 +7627,221 @@ void MegaClient::fetchnodes()
...
+    int creqtag = reqtag;
+    reqtag = 0;
+    reqs.add(new CommandPubKeyRequest(this, u));    // public RSA
+    reqtag = creqtag;
+
+    getua(u, "*keyring", 0);   // private Cu25519 & private Ed25519
+    getua(u, "+puEd255", 0);
+    getua(u, "+puCu255", 0);
+    getua(u, "+sigCu255", 0);
+    getua(u, "+sigPubk", 0); // it triggers MegaClient::initializekeys()
```
javiserrano: It seems that there are several code paths in `CommandGetUA::procresult()` that could keep `fetchingkeys` to `true` forever.

src/crypto/cryptopp.cpp
```diff
@@ -548,6 +445,27 @@ int AsymmCipher::setkey(int numints, const byte* data, int len)
...
+    size = key[0].ByteCount() + key[1].ByteCount();
+    d->reserve(d->size() + size);
+
+    for (int j = key[0].ByteCount(); j--;)
+    {
+        c = key[0].GetByte(j);
+        d->append(&c, sizeof c);
+    }
+
+    for (int j = 4; j--;)   // keep 4 bytes for exponent
```
javiserrano: Is this confirmed? (the usage of 4 bytes always for the exponent)

src/crypto/sodium.cpp
```diff
@@ -22,138 +22,190 @@
...
+    {
+        PrnGen::genblock(this->keySeed, EdDSA::SEED_KEY_LENGTH);
+    }
 
-// Sets a private key seed from a buffer.
-void EdDSA::setKeySeed(const char* data) {
-    memcpy(this->keySeed, data, crypto_sign_SEEDBYTES);
+    // derive public and private keys from the seed
+    if (crypto_sign_seed_keypair(this->pubKey, this->privKey, this->keySeed) != 0)
+    {
```
javiserrano: Since we are checking if the initialization is correct, we should avoid posterior usages of this object, maybe setting a flag or something like that.

src/crypto/sodium.cpp
```diff
@@ -22,138 +22,190 @@
...
-    free(signedmsg);
-    if (result) {
-        return(1);
-    } else {
-        // crypto_sign_open() returns -1 on failure.
-        return(0);
-    }
+    int result = crypto_sign_verify_detached(sig, msg, msglen, pubKey);
+
+    return (result == 0) ? 1 : 0;
```
javiserrano: This could be simplified just returning `result`, or `!!result` if it's needed to return `0` or `1`.

src/crypto/sodium.cpp
```diff
@@ -22,138 +22,190 @@
...
+{
+}
+
+int ECDH::encrypt(unsigned char *encmsg, const unsigned char *msg,
+                  const unsigned long long msglen, const unsigned char *nonce,
+                  const unsigned char *pubKey, const unsigned char *privKey)
+{
+    int check = crypto_box(encmsg, msg, msglen, nonce, pubKey, privKey);
+
+    return check ? 0 : 1;
```
javiserrano: This could be simplified returning `!check`, or directly returning `!crypto_box...`

src/crypto/sodium.cpp
```diff
@@ -22,138 +22,190 @@
...
+}
+
+int ECDH::decrypt(unsigned char *msg, const unsigned char *encmsg,
+                  const unsigned long long encmsglen, const unsigned char *nonce,
+                  const unsigned char *pubKey, const unsigned char *privKey)
+{
+    int check = crypto_box_open(msg, encmsg, encmsglen, nonce, pubKey, privKey);
+
+    return check ? 0 : 1;
```
javiserrano: Same as above

src/utils.cpp
```diff
@@ -360,4 +361,353 @@ int mega_snprintf(char *s, size_t n, const char *format, ...)
...
+
+        if (utf8cp1 < 0x80)
+        {
+            res[rescount++] = utf8cp1;
+        }
+        else
+        {
+            if (i < srclen)
+            {
+                utf8cp2 = src[i++];
```
javiserrano: It would be good to check if `utf8cp1` is `0xC2` or `0xC3` and `utf8cp2` is lower than `0xC0` and greater or equal to `0x80` (valid two-bytes UTF-8 character)

src/utils.cpp
```diff
@@ -360,4 +361,353 @@ int mega_snprintf(char *s, size_t n, const char *format, ...)
...
+
+    unsigned offset = 0;
+    encryptionsetting_t encSetting = (encryptionsetting_t) data->at(offset);
+    offset++;
+
+    unsigned ivlen = TLVstore::getIvlen(encSetting);
+    unsigned taglen = TLVstore::getTaglen(encSetting);
+    encryptionmode_t encMode = TLVstore::getMode(encSetting);
+
+    if (!ivlen || !taglen || encMode == AES_MODE_UNKNOWN)
```
javiserrano: It should be checked that data contains enough bytes to continue

src/megaapi_impl.cpp
```diff
@@ -8341,50 +8600,204 @@ void MegaApiImpl::getua_result(byte* data, unsigned len)
...
+
+    if (request->getType() == MegaRequest::TYPE_GET_ATTR_USER)
     {
-        string str((const char*)data,len);
-        request->setText(str.c_str());
+        if (tlv)
+        {
+            // TLV data usually includes byte arrays with zeros in the middle, so values
+            // must be converted into Base64 strings to avoid problems
+            MegaStringMap *stringMap = new MegaStringMapPrivate(tlv->getMap(), true);
```
javiserrano: It seems that the map returned by `tlv->getMap()` is leaked.

src/commands.cpp
sergiohs84: The number of elements can be 1 or 2 indeed. If you don't send a version because it's unknown at the moment of the request, then the API may reject it with `API_EEXPIRED` if the attribute already exists and has a version attached. In that case, you need to provide the latest version of the attribute to be able to update it.
This mechanism aims to prevent modification of user attributes that require to be aware of the latest version of the attribute, such as authentication rings or keys.

src/commands.cpp
sergiohs84: Yes, right now it's not going to be used. But I'd like to keep it there for reference and for a near future where SDK might need to adapt the behaviour based on the historical property of attributes.

src/megaclient.cpp
sergiohs84: As long as there is response to the request of `+sigPubk`, it will call `initializekeys()`. It doesn't really matter the response for the other attributes. Also, it doesn't matter if `+sigPubk` is found or not, it will trigger the initialization.
What's the code path you are concerned? I only see when an `API_EINTERNAL` is detected.

src/crypto/sodium.cpp
sergiohs84: I've implemented the protection for both, EdDSA and ECDH

src/crypto/cryptopp.cpp
sergiohs84: Not yet, launched the question to the Webclient team. I'll keep debugging tomorrow anyway.

src/utils.cpp
sergiohs84: Fixed! Check the line below

src/utils.cpp
sergiohs84: The code where this checkup should be done is in the `containerToTLVrecords()`, which is the one reading from a buffer. That's already done.
The one pointed by this comment is creating the container from the records, so there's no risk at all.

include/mega/crypto/sodium.h
sergiohs84: I added the line after the header file macro `SODIUM_H`. Not sure if it would cause any interference during the building process.

src/user.cpp
```diff
@@ -75,11 +100,13 @@ User* User::unserialize(MegaClient* client, string* d)
...
     unsigned char l;
+    unsigned short ll;
     string m;
     User* u;
     const char* ptr = d->data();
     const char* end = ptr + d->size();
     int i;
+    char attrVersion;
 
     if (ptr + sizeof(handle) + sizeof(time_t) + sizeof(visibility_t) + 2 > end)
```
javiserrano: This check doesn't seem to be enough to ensure that the buffer won't be accessed beyond the end

include/mega/crypto/sodium.h
javiserrano: It shouldn't cause any problem. I think that it's OK

src/megaapi_impl.cpp
```diff
@@ -1895,6 +1926,21 @@ void MegaRequestPrivate::setMegaTextChatList(MegaTextChatList *chatList)
 }
 #endif
 
+MegaStringMap *MegaRequestPrivate::getMegaStringMap() const
+{
+    return stringMap;
+}
+
+void MegaRequestPrivate::setMegaStringMap(const MegaStringMap *stringMap)
```
javiserrano: This function should be prepared to receive a NULL `stringMap`

src/megaclient.cpp
javiserrano: It seems to be fixed now

src/user.cpp
```diff
@@ -99,11 +126,23 @@ User* User::unserialize(MegaClient* client, string* d)
     l = *ptr++;
     if (l)
     {
+        if (ptr + l > end)
+        {
+            return NULL;
```
javiserrano: It would be good to have warning or error logs in these cases to detect problems serializing/unserializing objects

src/utils.cpp
javiserrano: I agree

src/user.cpp
sergiohs84: The code calling `unserialize()` already checks the resulting value and, if `NULL`, it does log an error.

src/megaapi_impl.cpp
sergiohs84: Done!

src/megaapi_impl.cpp
```diff
@@ -3202,6 +3344,27 @@ char *MegaApiImpl::getMyXMPPJid()
...
+    if (ISUNDEF(client->me))
+    {
+        sdkMutex.unlock();
+        return NULL;
+    }
+
+    char *result = NULL;
+    if (client->signkey)
+    {
+        result = MegaApi::strdup(client->signkey->genFingerprintHex());
```
javiserrano: Memory leak. Since the returned value is new, it could be directly set to "result"

src/commands.cpp
```diff
@@ -2173,146 +2325,240 @@ void CommandPutUA::procresult()
...
 
     client->app->putua_result(e);
 }
 
-CommandGetUA::CommandGetUA(MegaClient* client, const char* uid, const char* an)
+CommandGetUA::CommandGetUA(MegaClient* client, const char* uid, attr_t at, int ctag)
 {
-    user = client->finduser((char*)uid);
-    attributename = an;
+    this->u = client->finduser((char*)uid);
```
javiserrano: It's not specific of this pull request and there are other commands doing the same, but it's potentially dangerous to save this pointer. A fetchnodes in the same request before this command would invalidate the pointer.

javiserrano:
Reviewed and tested. All seems to be perfect :+1:

PR #293, opened 2016-05-11
Set a non-repeating timer for cURL on Windows

javiserrano:
The value returned by cURL was being used to set a repeating timer on Windows.
According to the documentation of cURL, that is wrong.
With this fix, the CPU usage will be reduced.


sergiohs84:
All good :+1:

PR #295, opened 2016-05-12
Fixes for the management of backoffs after errors

javiserrano:


src/backofftimer.cpp
```diff
@@ -110,11 +111,16 @@ void BackoffTimer::update(dstime* waituntil)
 {
     if (next)
     {
+        assert(next != 1);
+        if (next == 1)
+        {
+            LOG_warn << "Possible wrong management of timer";
+        }
+
```
javiserrano: Once the timeout has expired, in the next SDK loop the corresponding event should be attended and the backoff timer correctly configured. If `next` is still `1` in the next SDK look, that hasn't been correctly made and will probably cause a high CPU usage.

src/backofftimer.cpp
```diff
@@ -110,11 +111,16 @@ void BackoffTimer::update(dstime* waituntil)
...
         if (next <= Waiter::ds)
         {
             *waituntil = 0;
             next = 1;
-            base = 1;
```
javiserrano: The backoff shouldn't be reinitialized here because exponential backoffs don't work as expected. It's diffucult to reach this code in those circumstances, but it's possible.

src/megaclient.cpp
```diff
@@ -909,10 +909,18 @@ void MegaClient::exec()
 
                         // notify app in case some attributes were not returned, then redispatch
                         fc->failed(this);
+                        fc->req.disconnect();
+                        fc->req.status = REQ_PREPARED;
+                        fc->timeout.reset();
```
javiserrano: Keep the state of the request consistent, reset the timeout and ensure that the request is detached from the network layer. The state of this request is also used to control the state of another request sent using `pendingcs`, so it's important to keep the status updated and the timeout inactive when it isn't needed to prevent a high CPU usage in some circumstances.

src/megaclient.cpp
```diff
@@ -909,10 +909,18 @@ void MegaClient::exec()
...
                         fc->bt.reset();
                         break;
 
                     case REQ_INFLIGHT:
+                        if (!fc->req.httpio)
+                        {
+                            break;
+                        }
+
```
javiserrano: Don't run the code related to this state if the request is not really active. The state `REQ_INFLIGHT` is also set for this request when the download URL is being fetched from API servers (using `pendingcs`) and, during that time, nor the code below nor the timeout should be active.

src/megaclient.cpp
```diff
@@ -926,12 +934,16 @@ void MegaClient::exec()
...
                         // timeout! fall through...
-                        fc->req.disconnect();
                     case REQ_FAILURE:
+                        LOG_warn << "Error getting file attr";
                         fc->failed(this);
+                        fc->timeout.reset();
                         fc->bt.backoff();
                         fc->urltime = 0;
+                        fc->req.disconnect();
+                        fc->req.status = REQ_PREPARED;
```
javiserrano: The same as above. This code ensures that the request is detached from the network layer, that the timeout is deactivated and that the request keep a consistent state when it finishes, mainly to prevent problems with the timeout.

src/megaclient.cpp
```diff
@@ -943,12 +955,15 @@ void MegaClient::exec()
                     if (Waiter::ds - fc->urltime > 600)
                     {
                         // fetches pending for this unconnected channel - dispatch fresh connection
+                        LOG_debug << "Getting fresh download URL";
+                        fc->timeout.reset();
                         reqs.add(new CommandGetFA(this, cit->first, fc->fahref, httpio->chunkedok));
```
javiserrano: Ensure that the timeout is disabled while the download URL is being received from API servers.

src/fileattributefetch.cpp
```diff
@@ -82,6 +84,7 @@ void FileAttributeFetchChannel::dispatch(MegaClient* client)
     else
     {
         timeout.reset();
+        req.status = REQ_PREPARED;
```
javiserrano: Keep the state of the request updated

sergiohs84:
All fine! Lets move these critical fixes to develop for further testing.

PR #300, opened 2016-05-17
Ensure that the upload token is received after the completion of uploads

javiserrano:
If it isn't received, the upload fails and a notification is sent
The changes include compatibility with transfer resumption


sergiohs84:
Reviewed. Seems OK

PR #302, opened 2016-05-18
Authorize nodes to download them with a different MegaApi

javiserrano:


sergiohs84:
If you agree with the latest commit, I'm fine with this useful feature.

PR #304, opened 2016-05-19
Run an extra SDK loop when commands related to file attributes finish

javiserrano:
Fixed also a couple of edge cases managing errors on CommandGetFA


sergiohs84:
Great fix!

PR #306, opened 2016-05-19
Don't upsize previews

javiserrano:
if the image is smaller than the default preview size


sergiohs84:
Much better this way :+1:

PR #308, opened 2016-05-20
Add tests output

sergiohs84:
Tests will generate 2 log files, one for each MegaApi used during tests. The whole set of SDK tests will use the same file, creating sections for each test case.
Those files will be kept along the output of the Jenkins task for further analysis if required.


sergiohs84:
Already merged to `pcr-management`, which is part of `develop` by now.

PR #312, opened 2016-05-24
Remove obsolete interface + Update bindings

sergiohs84:
- Definitely remove the obsolete method to add contacts without confirmation.
- Replace `ur` request by `ur2` in order to remove contacts.
- Add method to remove users from `megacli`.
- Update bindings and prebuild libraries for examples.
- Minor changes to the building steps of some examples.


sergiohs84:
Not to be merged into `master`, but into `develop`

PR #315, opened 2016-05-24
Skip restoration of file attributes

sergiohs84:
When a node was copied/imported from another account, the new owner
cannot upload file attributes (only the former uploader of the file is
allowed to).

To avoid recurrent tries of restoration, the node is tagged with a
custom attribute, 'f', containing the own userhandle. In the future,
when missing attributes are detected, before attempting to restore them,
this attribute is checked.


sergiohs84:
OK!

PR #316, opened 2016-05-25
Speed limits for the cURL network layer

javiserrano:


src/posix/net.cpp
```diff
@@ -936,7 +1097,7 @@ void CurlHttpIO::send_request(CurlHttpContext* httpctx)
             }
         }
 
-        curl_multi_add_handle(httpio->curlm, curl);
+        curl_multi_add_handle((httpctx->data ? len : req->out->size()) ? httpio->curlmupload : httpio->curlmdownload, curl);
```
sergiohs84: As per discussed, review the usage of `httpctx->data` and `req->out` before merge.

PR #320, opened 2016-05-28
Case-sensitive fopen for reading on Windows

javiserrano:
WIth this pull request, fopen refuses to open a file for reading if the name of the local file doesn't exactly match (including case) the provided name.

This prevents a sync bug that detects local renames that really don't exist when there are queued filesystem notifications for a file that has been renamed changing only the case of letters.

Due to this problem, action packets informing about these renames can lead the SDK to send new commands modifying the same node. That can cause a storm of API commands if several SDK instances are syncing the same folder at once.


sergiohs84:
It seems OK

PR #329, opened 2016-06-13
New upload completion

javiserrano:


src/transfer.cpp
```diff
@@ -259,10 +259,10 @@ Transfer *Transfer::unserialize(MegaClient *client, string *d, transfer_map* tra
     ptr += sizeof(m_time_t);
 
 
-    bool hasUltoken = MemAccess::get<char>(ptr);
+    char hasUltoken = MemAccess::get<char>(ptr);
     ptr += sizeof(char);
 
-    ll = hasUltoken ? NewNode::UPLOADTOKENLEN + 1 : 0;
+    ll = hasUltoken ? ((hasUltoken == 1) ? NewNode::OLDUPLOADTOKENLEN + 1 : NewNode::UPLOADTOKENLEN) : 0;
```
sergiohs84: It would be good to check that `hasUltoken` is 0 | 1 | 2. Otherwise, log an error and abort un-serialization.

sergiohs84:
Perfect, much safer :)

PR #332, opened 2016-06-14
Fixed mandatory check for u elementin mcf response

alxvasilev:


sergiohs84:
Thanks for the fix! It seems correct.

PR #333, opened 2016-06-14
posix sempahore no longer uses sem_timedwait

polmr:
using posix mutex and condition variable

### DIFF
--- a/src/thread/posixthread.cpp
+++ b/src/thread/posixthread.cpp
@@ -186,24 +186,26 @@ int PosixSemaphore::timedwait(int milliseconds)
             return -2;
         }
     }
+    count --;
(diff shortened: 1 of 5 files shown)


src/thread/posixthread.cpp
```diff
@@ -134,40 +139,78 @@ void timespec_add_msec(struct timespec *tv, int milliseconds)
...
         {
+            pthread_mutex_unlock(&mtx);
             return 0;
         }
-
-        if (errno == ETIMEDOUT)
+        else if (errno == EAGAIN)
         {
-            return -1;
+            //continue;
```
polmr: do nothing: jump to pthread_cond_timedwait

polmr:
Seems OK

PR #342, opened 2016-06-28
Avatar color based on `userhandle`

sergiohs84:


sergiohs84:
Android is ready to make a release using the new approach.

PR #365, opened 2016-07-19
Chain share + link instead of send the in parallel

javiserrano:
This helps to solve the issue of the server keeping the sharekey of
deleted shares, which makes the initial share request to return the old
sharekey and the folder link creation to fail. This way, the share is
retried first with the old sharekey and later, the link creation works.


sergiohs84:
Ready to merge!

PR #368, opened 2016-07-20
Ability to set binary user attribute in megacli

sergiohs84:
Some new user attributes store binary data, so the `putua <attrname> set <attrvalue>` doesn't work because it expects plain ASCII strings.
This commit adds `putua <attrname> set64 <attrvalue64>`, which converts the value from base64 to binary before putting the new user attr value.


sergiohs84:
A convenience feature for `megacli`

PR #369, opened 2016-07-20
Android cross compile autotools fix

polmr:
This PR is not sufficient: cryptopp v 5.6.2 has a failure that prevents compilation. It is corrected in 5.6.3. Nevertheless, this PR doesn't changes cryptopp to v5.6.3 because:
1) 5.6.3 not fully tested in other compilations/executions.
2) 5.6.3 has a problem that prevents autotools headers cheking when cross-compiling for android (https://github.com/weidai11/cryptopp/issues/222). This has been corrected in cryptopp's github repo (https://github.com/weidai11/cryptopp/commit/1e1762041738e695d52aa89c356853373b804030). It will not be released in a long while, in the meantime it could be tricked with: ``export CXXFLAGS="$CXXFLAGS -DCRYPTOPP_UNCAUGHT_EXCEPTION_AVAILABLE"``

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -226,7 +226,7 @@ bool HashSignature::check(AsymmCipher* pubk, const byte* sig, unsigned len)
     if (size < h.size())
     {
         // left-pad with 0
-        s.insert(0, h.size() - size, 0);
+        s.insert((size_t)0, h.size() - size, 0);
         s.resize(h.size());
     }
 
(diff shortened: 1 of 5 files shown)


src/utils.cpp
```diff
@@ -443,7 +443,7 @@ string * TLVstore::tlvRecordsToContainer()
 
 string TLVstore::get(string type)
 {
-    return tlv.at(type);
+    return tlv[type];
```
sergiohs84: @polmr What's the goal of this change? Doesn't `std::map<key,value>::at()` compile in Android?
The reason of using `at()` instead of the operator `[ ]` is because the second one automatically creates the item in the map for the requested key in case it doesn't exist. And that's not the expected behaviour of a `get()`

polmr:
cross-compiling is unable to compile .at since it is c++11. There are several other calls to `std::map<key,value>::at()`, but all them are in megaapi_impl.cpp which is NOT compiled when invoking build_sdk.sh (see build_android.sh).

polmr:
https://github.com/meganz/sdk/tree/androidc%2B%2B11 makes cross-compiling using autotools with c++11 posible

polmr:
Discarded in favour of https://github.com/meganz/sdk/pull/393

PR #377, opened 2016-07-30
msvc 2015 build patch

alxvasilev:
Fixes to allow proper build with MSVC 2015


javiserrano:
I have merged the latest changes in this branch to be able to merge this pull request and I have also reverted some changes that could cause build problems to other apps.
I have just tested the build of this branch with VS2015 and it works for me. @alxvasilev, if it works for you too I think that we can merge the pull request.

PR #378, opened 2016-08-01
Fix search

sergiohs84:
The `SearchTreeProcessor::processNode()` method includes rootnodes as part of the result if the name of the rootnode matches the searchString. Better to exclude them.


sergiohs84:
Wrong target branch!

PR #380, opened 2016-08-05
Check the reason of being blocked

sergiohs84:
The API checks the session in the POST request and, in case the corresponding user account is blocked, it directly returns `-16` (EBLOCKED, without brackets). It's catched at `request_error()`.

Currently, it logs out immediately. This branch sends a `whyamiblocked` command to API and it logs out after processing the response, so the reason received can be provided to apps prior to disconnect.


sergiohs84:
trigger compilation

PR #384, opened 2016-08-11
Fix the generation of thumbnails for jpg files with QT5

javiserrano:
It seems that the same QImageReader can't read twice the same JPG image so it's needed to create a new one.

The restoration of file attributes has been enabled again.


sergiohs84:
All fine with this PR :+1:

PR #394, opened 2016-08-19
moved glob.h to mega_glob.h

polmr:
deleted glob.c and third_party folder
modified compilation files to point to the new location


sergiohs84:
All fine.

PR #395, opened 2016-08-19
fix memory leak at sync.cpp

polmr:


polmr:
proposed fix is not safe

PR #397, opened 2016-08-22
Feature/down public folders

vtmateos:


polmr:
do not merge, there are some memory issues

polmr:
should be fixed when merging: https://github.com/meganz/sdk/pull/411

src/megaapi_impl.cpp
```diff
@@ -118,6 +119,8 @@ MegaNodePrivate::MegaNodePrivate(MegaNode *node)
     this->name = MegaApi::strdup(node->getName());
     this->fingerprint = MegaApi::strdup(node->getFingerprint());
     this->customAttrs = NULL;
+    this->children = node->getChildren();
+    if (this->children) this->children = this->children->copy()
```
polmr: missin ';'

src/megaapi_impl.cpp
```diff
@@ -10938,13 +11023,13 @@ void MegaApiImpl::sendPendingTransfers()
...
 
-                if (!transfer->isStreamingTransfer() && node && node->type != FILENODE)
+                if (!transfer->isStreamingTransfer() && ((node && node->type != FILENODE) || (publicNode && publicNode->getType() != FILENODE)) )
                 {
                     // Folder download
                     transferMap[nextTag] = transfer;
                     transfer->setTag(nextTag);
                     MegaFolderDownloadController *downloader = new MegaFolderDownloadController(this, transfer);
-                    downloader->start();
+                    downloader->start(publicNode);
```
javiserrano: publicNode is not only set in the case of public nodes. It's set also for the own nodes. I have changed the name of the variable to show it.

PR #401, opened 2016-08-24
Allow to change the number of connections per transfer

javiserrano:


include/megaapi.h
```diff
@@ -6238,6 +6240,25 @@ class MegaApi
...
+         *
+         * The associated request type with this request is MegaRequest::TYPE_SET_MAX_CONNECTIONS
+         * Valid data in the MegaRequest object received on callbacks:
+         * - MegaRequest::getParamType - Returns the first parameter
+         * - MegaRequest::getNumber - Returns the second parameter
+         *
+         * @param direction Direction of transfers
+         * Valid values for this parameter are:
+         * - MegaTransfer::TYPE_DOWNLOAD = 0
+         * - MegaTransfer::TYPE_UPLOAD = 1
```
sergiohs84: The implementation allows to set the number of connections for both directions at once by using `direction = -1`.
We can add a new prototype, like for `pauseTranfers()`, to apply the setting to both download and upload.

src/megaapi_impl.cpp
```diff
@@ -12553,6 +12563,42 @@ void MegaApiImpl::sendPendingRequests()
...
+
+            if (connections <= 0 || (direction != -1
+                    && direction != MegaTransfer::TYPE_DOWNLOAD
+                    && direction != MegaTransfer::TYPE_UPLOAD))
+            {
+                e = API_EARGS;
+                break;
+            }
+
+            if (connections > 6)
```
sergiohs84: I'd make a new definition for `#define MAX_CONNECTIONS 6`

sergiohs84:
Now it's ready to merge :)

sergiohs84:
trigger compilation

PR #409, opened 2016-08-31
Fix ownerkey

sergiohs84:
When a share is removed, in order for the participants to not lose cryptographic access to the not-know-yet nodes, the API retains the share secret. It created problems in the past when creating a public folder link (already deployed a hotfix).
When the share is removed by the owner and there's no more participants, the share secret should be effectively removed. This branch aims to forget the share secret only in those cases.

### DIFF
--- a/include/mega/share.h
+++ b/include/mega/share.h
@@ -56,12 +56,12 @@ struct MEGA_API NewShare
 
     handle pending;
     bool upgrade_pending_to_full;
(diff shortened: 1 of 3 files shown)


sergiohs84:
Good protection against keeping the key when it's deleted in API. Waiting for API to deploy `okd` flag. Merge once the change in API is done.

PR #412, opened 2016-09-02
Megacli api authorize folders

polmr:


polmr:
all fine. however, it might cause conflicts when merging megacliAPI into the sdk (attention to that should be payed)

PR #424, opened 2016-09-21
fixed missing owner access in megacli

polmr:


sergiohs84:
The type `accesslevel_t` has two more access levels defined. Not sure if they should also be included in the array.

sergiohs84:
Much better! Ready to merge. @javiserrano ?

PR #425, opened 2016-09-21
fix nameid2string

polmr:


sergiohs84:
I'm fine with the fix too

PR #426, opened 2016-09-22
Allow to fopen folders, get the fsid and improve the scanning performance

javiserrano:


src/sync.cpp
```diff
@@ -871,12 +891,6 @@ LocalNode* Sync::checkpath(LocalNode* l, string* localpath, string* localname)
     }
     else
     {
-        if (initializing)
-        {
-            delete fa;
-            return NULL;
-        }
-
         LOG_warn << "Error opening file";
```
sergiohs84: With this change, we keep returning NULL, but now we'll do also some actions that we used to not do when we were initializing. Not sure if that's your aim.

javiserrano: If the "initializing" flag is set, the function will already return NULL due to the new "if" above, so this code won't be reached anymore with that flag set.

PR #434, opened 2016-10-04
Disconnect after timeout and server reports no lock

polmr:


polmr:
tested faking a non returned petition -> success

PR #442, opened 2016-10-19
Don't block the SDK during the cancellation of many transfers

javiserrano:


src/megaapi_impl.cpp
```diff
@@ -11329,35 +11329,42 @@ void MegaApiImpl::removeRecursively(const char *path)
...
-		e = API_OK;
+        sdkMutex.lock();
+        if (!request->getTag())
+        {
+            nextTag = client->nextreqtag();
+            request->setTag(nextTag);
+            requestMap[nextTag]=request;
+            fireOnRequestStart(request);
+        }
+        else
```
sergiohs84: Maybe a comment here: `// this case happens when we queue requests already started`

PR #463, opened 2016-11-10
Don't try to load a fetchnodes cache after the reception of API_ETOOMANY

javiserrano:


javiserrano:
It's tested and working. 👍

PR #465, opened 2016-11-14
Return NULL for MegaTextChat::getUrl and getTitle if empty

sergiohs84:


sergiohs84:
Too late, Jenkins!! :smile:

PR #466, opened 2016-11-14
Thread-safety for cURL and OpenSSL

javiserrano:


polmr:
Fails to compile in macos: src/thread/posixthread.cpp:46:12: error: cannot initialize return object of type 'long' with an rvalue of type 'pthread_t' (aka '_opaque_pthread_t *')
    return pthread_self();
           ^~~~~~~~~~~~~~

PR #468, opened 2016-11-16
Improved management of transfers

javiserrano:
- New management of transfer queues (FIFO)
- Concurrent file attribute puts
- Management of transfer priorities
- Allow to pause tranfers
- Correct management of multifile transfers on MegaApi
- New management of transfer callbacks
- Management of transfer states
- Get information about the order of all transfers at once
- Performance improvements for transfer resumption


src/transfer.cpp
```diff
@@ -1199,4 +1252,431 @@ DirectReadSlot::~DirectReadSlot()
...
+    if (dstit == transfers[transfer->type].end()
+            || (dstit + 1) == transfers[transfer->type].end())
+    {
+        LOG_debug << "Moving transfer to the last position";
+        if (transfer->slot)
+        {
+            transfer_list::iterator cit = it + 1;
+            while (cit != transfers[transfer->type].end())
+            {
+                if (!(*cit)->slot)
```
sergiohs84: Check not only that the transfer doesn't have one slot, but also that is electable for dispatching.

src/transfer.cpp
```diff
@@ -1199,4 +1252,431 @@ DirectReadSlot::~DirectReadSlot()
...
+
+        it--;
+        while (it != cit)
+        {
+            if ((*it)->slot && (*it)->state == TRANSFERSTATE_ACTIVE)
+            {
+                (*it)->bt.arm();
+                (*it)->cachedtempurl = (*it)->slot->tempurl;
+                delete (*it)->slot;
+                (*it)->state = TRANSFERSTATE_QUEUED;
```
sergiohs84: Missing `transfercacheadd(*it)`

include/megaapi.h
```diff
@@ -4137,6 +4309,15 @@ class MegaApi
         char* getStringHash(const char* base64pwkey, const char* email);
 
         /**
+         * @brief Get internal timestamp used by the SDK
+         *
+         * This is a time used in certain internal operations.
+         *
+         * @return actual SDK time in deciseconds
+         */
+        int getSDKtime();
```
sergiohs84: The datatype used for transfer's timestamps is `uint64_t`, so better to return `long long` here.

polmr:
trigger compilation

src/transfer.cpp
```diff
@@ -1199,4 +1252,393 @@ DirectReadSlot::~DirectReadSlot()
...
+    {
+        transfer_list::iterator cit = it + 1;
+        while (cit != transfers[transfer->type].end())
+        {
+            if (!(*cit)->slot)
+            {
+                transfer->bt.arm();
+                transfer->cachedtempurl = (*it)->slot->tempurl;
+                delete transfer->slot;
+                transfer->state = TRANSFERSTATE_QUEUED;
```
sergiohs84: Review if the cache should be updated and the app notified.

javiserrano: Yes, it is, I have checked all the usages of the function

include/megaapi.h
javiserrano: Waiter::ds is dstime (uint32_t) but I have changed it anyway because all other functions managing timestamps in MegaApi use "long long".

src/transfer.cpp
javiserrano: Indeed, it's not strictly needed to save that state to the transfer cache in this moment. But it could cause bugs difficult to detect in the future, so I have added it.

src/megaclient.cpp
```diff
@@ -10117,9 +10179,21 @@ bool MegaClient::startxfer(direction_t d, File* f, bool skipdupes)
                 t->size = f->size;
             }
 
+            t->lastaccesstime = time(NULL);
```
javiserrano: Despite we talked in the office, I have checked it and `lastaccesstime` is an absolute time, so that code is correct. That absolute time is used to control the expiration of temporary URLs, so it's needed that it's valid for different executions of the app.

javiserrano:
I think that I have fixed all pending issues

PR #470, opened 2016-11-22
Fix OpenSSL callbacks for 64 bits builds

javiserrano:


polmr:
Seems OK to go to develop. Further usage tests will be carried out soon.

PR #471, opened 2016-11-24
Dissable check of message in gtests

polmr:


javiserrano:
Already fixed on `develop`

PR #479, opened 2016-12-06
Better management of integrity checks

javiserrano:
- CRC for uploads
- Provide the last error in transfers to apps
- Retry the upload of chunks that fails due to the CRC
- Repeat wrong downloads to check if the file is really corrupt
- Move corrupt files to SyncDebris
- Reorganized the management of the variables "progresscompleted", "errorcount" and "failcount"

### DIFF
--- a/src/file.cpp
+++ b/src/file.cpp
@@ -323,13 +323,13 @@ bool File::failed(error e)
 {
     if (e == API_EKEY)
     {
-        if (!transfer->currentmetamacvalid)
+        if (!transfer->hascurrentmetamac)
         {
             // several integrity check errors uploading chunks
             return transfer->failcount < 1;
         }
 
-        if (transfer->prevmetamacvalid && transfer->prevmetamac == transfer->currentmetamac)
+        if (transfer->hasprevmetamac && transfer->prevmetamac == transfer->currentmetamac)
         {
             // integrity check failed after download, two times with the same value
             return false;
@@ -337,8 +337,8 @@ bool File::failed(error e)
 
         // integrity check failed once, try again
         transfer->prevmetamac = transfer->currentmetamac;
-        transfer->prevmetamacvalid = true;
-        return true;
+        transfer->hasprevmetamac = true;
+        return transfer->failcount < 16;
     }
 
     return (e != API_EBLOCKED && e != API_ENOENT && e != API_EINTERNAL && transfer->failcount < 16)
--- a/src/file.cpp
+++ b/src/file.cpp
@@ -321,8 +321,28 @@ void File::terminated()
 // failuresup to 16 times, except I/O errors (6 times)
 bool File::failed(error e)
 {
-    return (e != API_EKEY && e != API_EBLOCKED && e != API_ENOENT && e != API_EINTERNAL && transfer->failcount < 16) &&
-            !((e == API_EREAD || e == API_EWRITE) && transfer->failcount > 6);
+    if (e == API_EKEY)
+    {
+        if (!transfer->currentmetamacvalid)
+        {
+            // several integrity check errors uploading chunks
+            return transfer->failcount < 1;
+        }
+
+        if (transfer->prevmetamacvalid && transfer->prevmetamac == transfer->currentmetamac)
+        {
+            // integrity check failed after download, two times with the same value
+            return false;
+        }
+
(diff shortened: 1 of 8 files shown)


src/file.cpp
```diff
@@ -321,8 +321,28 @@ void File::terminated()
...
-            !((e == API_EREAD || e == API_EWRITE) && transfer->failcount > 6);
+    if (e == API_EKEY)
+    {
+        if (!transfer->currentmetamacvalid)
+        {
+            // several integrity check errors uploading chunks
+            return transfer->failcount < 1;
+        }
+
+        if (transfer->prevmetamacvalid && transfer->prevmetamac == transfer->currentmetamac)
```
sergiohs84: Maybe refactor `prevmetamacvalid` and `currentmetamacvalid` to something that doesn't mean _valid_, like `hasprevmetamac` or similar.

src/file.cpp
```diff
@@ -321,8 +321,28 @@ void File::terminated()
...
+        {
+            // integrity check failed after download, two times with the same value
+            return false;
+        }
+
+        // integrity check failed once, try again
+        transfer->prevmetamac = transfer->currentmetamac;
+        transfer->prevmetamacvalid = true;
+        return true;
```
sergiohs84: As a security meassure to avoid too many retries when every time you get different data from network, better to `return transfer->failcount < 16;`

src/http.cpp
```diff
@@ -660,6 +656,47 @@ bool HttpReqUL::prepare(FileAccess* fa, const char* tempurl, SymmCipher* key,
...
+    const char *data = out->data();
+    byte c[12];
+    memset(c, 0, 12);
+
+#ifndef ALLOW_UNALIGNED_MEMORY_ACCESS
+    for (int l = size; l--;)
+    {
+        c[l % 12] ^= data[l];
+    }
+#else
```
sergiohs84: It could be interesting to perform a benchmarking to check if the code below is actually much faster than the version above.

src/http.cpp
```diff
@@ -660,6 +656,47 @@ bool HttpReqUL::prepare(FileAccess* fa, const char* tempurl, SymmCipher* key,
     // unpad for POSTing
     out->resize(size);
 
+
+    const char *data = out->data();
+    byte c[12];
```
sergiohs84: Better to define the `crcSize` as a #define or constant and change it below when used, instead of 12.

src/transferslot.cpp
```diff
@@ -310,21 +322,31 @@ void TransferSlot::doio(MegaClient* client)
 
                             return transfer->failed(API_EINTERNAL);
                         }
+
+                        errorcount = 0;
+                        transfer->failcount = 0;
```
sergiohs84: Check if `failcount` should actually be reset here, since it's the chunk the one that has been sent/received correctly, but `failcount` could already be greater than zero. Reseting the variable would allow to retry the transfer more than once because after a failure there is a chunk transferred successfully.

src/transferslot.cpp
```diff
@@ -339,16 +361,18 @@ void TransferSlot::doio(MegaClient* client)
...
                                 else
                                 {
-                                    LOG_warn << "MAC verification failed";
+                                    LOG_warn << "MAC verification failed";                                    
                                     transfer->chunkmacs.clear();
                                     return transfer->failed(API_EKEY);
                                 }
                             }
+                            errorcount = 0;
+                            transfer->failcount = 0;
```
sergiohs84: Check if `failcount` should actually be reset here, since it's the chunk the one that has been sent/received correctly, but `failcount` could already be greater than zero. Reseting the variable would allow to retry the transfer more than once because after a failure there is a chunk transferred successfully.

src/transferslot.cpp
javiserrano: I agree, but it's the current behavior in `master` and I think that it's better to not change it during this pull request, since it's a different issue that should be analyzed and tested.

src/transferslot.cpp
javiserrano: Same as above

src/transferslot.cpp
sergiohs84: OK, sounds reasonable

PR #484, opened 2016-12-14
Update utf8proc

javiserrano:


sergiohs84:
It would be great, @polmr , if you can run the tests with random UTF8 chars

javiserrano:
Tests on Windows and Linux are OK after the last commit

PR #489, opened 2016-12-19
Add command to get email from userhandle

sergiohs84:
even when the user doesn't have any relationship with you (like shares or groupchats)

### DIFF
--- a/include/mega/megaclient.h
+++ b/include/mega/megaclient.h
@@ -225,7 +225,7 @@ public:
     void getua(const char* email_handle, const attr_t at = ATTR_UNKNOWN, int ctag = -1);
 
     // retrieve the email address of a user
-    void getUserEmail(const char *uid = NULL);
+    void getUserEmail(const char *uid);
 
 #ifdef DEBUG
     // queue a user attribute removal
--- a/include/mega/megaclient.h
+++ b/include/mega/megaclient.h
@@ -224,6 +224,9 @@ public:
     // queue a user attribute retrieval (for non-contacts)
     void getua(const char* email_handle, const attr_t at = ATTR_UNKNOWN, int ctag = -1);
 
+    // retrieve the email address of a user
+    void getUserEmail(const char *uid = NULL);
+
 #ifdef DEBUG
     // queue a user attribute removal
     void delua(const char* an);
--- a/src/commands.cpp
+++ b/src/commands.cpp
@@ -2688,8 +2688,6 @@ void CommandDelUA::procresult()
 
 #endif
 
-
-
 CommandGetUserEmail::CommandGetUserEmail(MegaClient *client, const char *uid)
 {
     cmd("uge");
@@ -2714,7 +2712,6 @@ void CommandGetUserEmail::procresult()
     {
         return client->app->getuseremail_result(&email, API_OK);
     }
-
 }
 
 // set node keys (e.g. to convert asymmetric keys to symmetric ones)
--- a/src/commands.cpp
+++ b/src/commands.cpp
@@ -4375,7 +4375,7 @@ CommandChatCreate::CommandChatCreate(MegaClient *client, bool group, const userp
 
         handle uh = itupl->first;
         char uid[12];
-        Base64::btoa((byte*)&uh, sizeof uh, uid);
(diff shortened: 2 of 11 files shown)


sergiohs84:
trigger compilation

include/mega/megaclient.h
```diff
@@ -224,6 +224,9 @@ class MEGA_API MegaClient
     // queue a user attribute retrieval (for non-contacts)
     void getua(const char* email_handle, const attr_t at = ATTR_UNKNOWN, int ctag = -1);
 
+    // retrieve the email address of a user
+    void getUserEmail(const char *uid = NULL);
```
javiserrano: I think that the default parameter should be removed. Since it's not checked if the value is NULL, the SDK will probably crash. Also, having it doesn't add any advantage because the function can't succeed without it. It would be different if it could provide the own email when NULL is passed, for example.

src/commands.cpp
```diff
@@ -2688,6 +2688,35 @@ void CommandDelUA::procresult()
 
 #endif
 
+
+
+CommandGetUserEmail::CommandGetUserEmail(MegaClient *client, const char *uid)
+{
+    cmd("uge");
```
javiserrano: `this->client = client;` could be added here, like in other commands.
It's not strictly needed because the SDK already sets it externally before calling `procresult()` but it seems a bit more logical that the command saves the values it needs rather than relying on the external access to its attributes.

src/commands.cpp
```diff
@@ -2688,6 +2688,35 @@ void CommandDelUA::procresult()
...
+    string email;
+    if (!client->json.storeobject(&email))
+    {
+        return client->app->getuseremail_result(NULL, API_EINTERNAL);
+    }
+    else
+    {
+        return client->app->getuseremail_result(&email, API_OK);
+    }
+
```
javiserrano: I would remove this empty line, and a couple of them before the constructor.

src/megaapi_impl.cpp
```diff
@@ -12425,6 +12455,22 @@ void MegaApiImpl::sendPendingRequests()
...
+        {
+            handle uh = request->getNodeHandle();
+            if (uh == INVALID_HANDLE)
+            {
+                e = API_EARGS;
+                break;
+            }
+
+            char uid[12];
+            Base64::btoa((byte*)&uh, sizeof uh, uid);
```
javiserrano: I think that the size should be `MegaClient::USERHANDLE` instead of `sizeof uh` despite in this case both are the same.

src/megaapi_impl.cpp
```diff
@@ -9687,6 +9696,27 @@ void MegaApiImpl::delua_result(error)
...
+    {
+        return;
+    }
+    MegaRequestPrivate* request = requestMap.at(client->restag);
+    if(!request || (request->getType() != MegaRequest::TYPE_GET_USER_EMAIL))
+    {
+        return;
+    }
+
+    if(e == API_OK && email)
```
javiserrano: Add a space after each `if` in this function.

src/megaapi_impl.cpp
```diff
@@ -2650,6 +2650,7 @@ const char *MegaRequestPrivate::getRequestString() const
         case TYPE_CHAT_SET_TITLE: return "CHAT_SET_TITLE";
         case TYPE_SET_MAX_CONNECTIONS: return "SET_MAX_CONNECTIONS";
         case TYPE_CHAT_PRESENCE_URL: return "CHAT_PRESENCE_URL";
+        case TYPE_GET_USER_EMAIL: return "TYPE_GET_USER_EMAIL";
```
javiserrano: Remove `TYPE_`

include/mega/megaclient.h
sergiohs84: I don't think it's the case. The `uid` is mandatory for the command, AFAIK, so better to not default the value.

src/commands.cpp
sergiohs84: Agree! Actually, most of the commands do not set `this->client`, and the right place would be the `Command` ctor. We can plan to amend that for all the commands at once.

PR #494, opened 2016-12-21
Fix memory leaks

javiserrano:


src/megaclient.cpp
```diff
@@ -278,10 +292,14 @@ void MegaClient::mergenewshare(NewShare *s, bool notify)
...
+                                if (n->pendingshares->erase(s->pending))
                                 {
-                                    n->changed.pendingshares = true;
-                                    notifynode(n);
+                                    if (notify)
+                                    {
+                                        n->changed.pendingshares = true;
+                                        notifynode(n);
+                                    }
+                                    delete *sharep;
```
polmr: shouldn't delete *sharep be outside if?

sergiohs84: I agree. Probably, the `if (n->pendingshares->erase(s->pending))` is not required at all, since `*sharep` is assigned to an existing value in the map and checked in the previous `if`, so the `erase()` won't fail ever.

src/megaapi_impl.cpp
```diff
@@ -3613,10 +3613,27 @@ MegaApiImpl::~MegaApiImpl()
...
     delete gfxAccess;
     delete fsAccess;
-//    delete httpio;  do not delete since it could crash
+    delete waiter;
+
+#ifndef DONT_RELEASE_HTTPIO
+    delete httpio;
+#endif
+
+    fireOnRequestFinish(request, MegaError(API_OK));
```
sergiohs84: Nevermind. I missed the `requestMap.erase(request->getTag());`

sergiohs84:
Much readable code for the `mergenewshares()`. From my point of view, it's ready for merging :+1:

PR #502, opened 2016-12-28
Improved the logic to resume transfers

javiserrano:


sergiohs84:
trigger compilation

src/megaclient.cpp
```diff
@@ -10178,39 +10193,62 @@ bool MegaClient::startxfer(direction_t d, File* f, bool skipdupes)
             if (it != cachedtransfers[d].end())
             {
                 LOG_debug << "Resumable transfer detected";
-                Transfer *transfer = it->second;
+                t = it->second;
+                if ((d == GET && !t->pos) || ((time(NULL) - t->lastaccesstime) >= 86400))
```
sergiohs84: Does it discard the URL for download transfers which position is 0? If so, why?

src/megaclient.cpp
```diff
@@ -10178,39 +10193,62 @@ bool MegaClient::startxfer(direction_t d, File* f, bool skipdupes)
...
                 else
                 {
-                    if (d == GET)
+                    if (d == PUT)
                     {
-                        LOG_warn << "Temporary file not found";
+                        if (t->genfingerprint(fa))
+                        {
+                            LOG_warn << "The local file has been modified";
+                            t->cachedtempurl.clear();
```
sergiohs84: Is there any way to notify the storage server about an invalid partial file already uploaded? or we just relay on the purging scripts from server?

src/megaclient.cpp
```diff
@@ -10178,39 +10193,62 @@ bool MegaClient::startxfer(direction_t d, File* f, bool skipdupes)
...
-                    LOG_debug << "Resuming transfer";
-                    t = transfer;
                 }
                 else
                 {
-                    if (d == GET)
+                    if (d == PUT)
                     {
-                        LOG_warn << "Temporary file not found";
+                        if (t->genfingerprint(fa))
```
sergiohs84: We should refactorize this function to `Transfer::checkfingerprint(FileAccess *fa, bool ignoreTime)`, IMHO.

src/megaclient.cpp
```diff
@@ -10178,39 +10193,62 @@ bool MegaClient::startxfer(direction_t d, File* f, bool skipdupes)
...
+                if (!fa->fopen(&t->localfilename))
                 {
-                    if (!transfer->pos)
+                    if (d == PUT)
                     {
-                        // to prevent HTTP 403 errors
-                        transfer->cachedtempurl.clear();
+                        LOG_warn << "Local file not found";
+                        // the transfer will be retried to ensure that the file
+                        // is not just just temporarily blocked
```
sergiohs84: I guess if it's not the case (temporary blocked), the transfer will be removed, right?

src/megaclient.cpp
```diff
@@ -7945,18 +7943,38 @@ bool MegaClient::fetchsc(DbTable* sctable)
...
+
     for (int d = GET; d == GET || d == PUT; d += PUT - GET)
     {
         while (cachedtransfers[d].size())
         {
             transfer_map::iterator it = cachedtransfers[d].begin();
             Transfer *transfer = it->second;
-            m_time_t t = time(NULL) - transfer->lastaccesstime;
-
-            if (remove || (transfer->type == PUT && t >= 86400)
```
javiserrano: The type of transfers can only be `GET` or `PUT`. The value `API` is only internally used by network layers to identify API request. `transfer->type` can't have such value.

src/megaclient.cpp
javiserrano: Yes, the case will be managed exactly like it would be if the local file is deleted during the transfer (at runtime).
It's retried up to 6 times with exponential backoff (see File::failed()).

src/megaclient.cpp
javiserrano: `genfingerprint` checks the fingerprint but also updates the Transfer object with the new one. Since we are going to resume the transfer anyway (to preserve its priority) we need that, after the check, the Transfer object contains the updated fingerprint.

src/megaclient.cpp
javiserrano: AFAIK that feature doesn't exist. The temporary file will be deleted from storage servers in 24h.

src/megaclient.cpp
sergiohs84: Perfect! We're safe then :)

src/megaclient.cpp
sergiohs84: OK

src/megaclient.cpp
sergiohs84: No comments then :+1:

src/megaclient.cpp
sergiohs84: So nothing else we can do from our side.

PR #503, opened 2016-12-29
Improve the network layer

javiserrano:
- Accurate speed control
- Count streaming and file attributes to calculate the speed
- Use the `curl_multi_socket` interface in all platforms
- Removed the support for chunked requests
- Other minor improvements


javiserrano:
trigger compilation

src/http.cpp
```diff
@@ -684,21 +665,20 @@ m_off_t SpeedController::calculateSpeed(long long numBytes)
 
     while (transferBytes.size())
     {
-        dstime deltaTime = currentTime - transferTimes.front();
-        if (deltaTime <= SPEED_MEAN_INTERVAL_DS)
+        map<dstime, m_off_t>::iterator it = transferBytes.begin();
+        dstime deltaTime = currentTime - it->first;
+        if (deltaTime < SPEED_MEAN_INTERVAL_DS)
```
sergiohs84: Did you change the `lower or equal than` by `lower than` in purpose?

src/http.cpp
```diff
@@ -684,21 +665,20 @@ m_off_t SpeedController::calculateSpeed(long long numBytes)
 
     while (transferBytes.size())
```
sergiohs84: I'd replace this loop by a different version, which is probably more efficient since it doesn't create a new iterator for each iteration of the loop:
```
    map<dstime, m_off_t>::iterator it = transferBytes.begin();
    while (it != transferBytes.end())
    {
        dstime dt = it->first;
        if (currentTime - dt < SPEED_MEAN_INTERVAL_DS)
        {
            break;
        }

        partialBytes -= it->second;
        it++;
        transferBytes.erase(dt);
    }
```

javiserrano: I think that it should be more efficient to get an iterator to the beginning of the list on each iteration than calling erase with a key, that requires to find the element in the whole map.

src/http.cpp
javiserrano: Yes, if `SPEED_MEAN_INTERVAL_DS'  is one, for example, only the element corresponding to the current decisecond should be kept in the map. Otherwise two deciseconds would be taken into account.

src/posix/net.cpp
```diff
@@ -362,25 +363,36 @@ void CurlHttpIO::addaresevents(WinWaiter *waiter)
         {
             ((WinWaiter *)waiter)->addhandle(info.handle, Waiter::NEEDEXEC);
         }
-
+#else
```
sergiohs84: Not sure if Windows Phone will work on the new way with Posix. To be tested and confirmed!

src/http.cpp
sergiohs84: You're very right. I didn't pay attention to the cost of the `erase(key)` compared with `erase(pos)`.
```
For the first version (erase(position)), amortized constant.
For the second version (erase(val)), logarithmic in container size.
```

src/posix/net.cpp
```diff
@@ -426,42 +465,143 @@ void CurlHttpIO::closecurlevents(direction_t d)
...
     int dummy = 0;
-    std::map<int, SockInfo> &socketmap = (d == API) ? curlapisockets : ((d == GET) ? curldownloadsockets : curluploadsockets);
+    bool isapipaused = false;
+    std::map<int, SockInfo> *socketmap;
+    m_time_t *timeout;
+    bool *paused;
 
-    bool active = false;
-    for (std::map<int, SockInfo>::iterator it = socketmap.begin(); it != socketmap.end();)
+    if (d == API)
```
sergiohs84: If we had declared these variables like:
```
curlsockets[3];
```
and so on, we could use the `d` as index to access to the corresponding curl socket, curl timeout reset, number of connections...

src/posix/net.cpp
```diff
@@ -426,42 +465,143 @@ void CurlHttpIO::closecurlevents(direction_t d)
...
+    {
+        socketmap = &curlapisockets;
+        timeout = &curlapitimeoutreset;
+        paused = &isapipaused;
```
sergiohs84: Why not to set `paused` directly to `false`? Note `isapipaused` is used nowhere else

src/http.cpp
sergiohs84: We can then call `transferBytes(it++)`, which would be the same result.

src/posix/net.cpp
javiserrano: I agree! I will reorganize the code following that approach.

src/posix/net.cpp
javiserrano: Because it's a pointer :-)

src/posix/net.cpp
```diff
@@ -522,19 +653,18 @@ void CurlHttpIO::disconnect()
     curl_multi_cleanup(curlmdownload);
     curl_multi_cleanup(curlmupload);
 
-#if defined(_WIN32) && !defined(WINDOWS_PHONE)
-    for (unsigned int i = 0; i < aressockets.size(); i++)
+    if (numapiconnections || numdownloadconnections || numuploadconnections)
```
javiserrano: Good idea

src/posix/net.cpp
```diff
@@ -1632,10 +1706,26 @@ void CurlHttpIO::cancel(HttpReq* req)
...
+                numuploadconnections--;
+            }
+            else
+            {
+                multihandle = curlmdownload;
+                numdownloadconnections--;
+            }
+
+            curl_multi_remove_handle(multihandle, httpctx->curl);
+            pauseddownloads.erase(httpctx->curl);
```
javiserrano: It was in purpose, but the reorganization of the code will take care of this

src/posix/net.cpp
```diff
@@ -1680,53 +1770,71 @@ bool CurlHttpIO::doio()
...
     processcurlevents(API);
-#endif
     result |= multidoio(curlmapi);
 
-    if (!maxdownloadspeed || maxdownloadspeed > downloadSpeed)
+    partialdownloaddata = 0;
+    if (aredownloadspaused)
+    {
+        aredownloadspaused = false;
+        while (pauseddownloads.size() && !aredownloadspaused)
```
javiserrano: I think that it should be safe, because insertions are controlled by `aredownloadspaused` and deletions should never happen there. I will check it.

src/posix/net.cpp
```diff
@@ -1680,53 +1770,71 @@ bool CurlHttpIO::doio()
...
-#endif
         result |= multidoio(curlmdownload);
     }
 
-    if (!maxuploadspeed || maxuploadspeed > uploadSpeed)
+    partialuploaddata = 0;
+    if (areuploadspaused)
+    {
+        areuploadspaused = false;
+        while (pauseduploads.size() && !areuploadspaused)
```
javiserrano: It's the same case as above.

src/http.cpp
sergiohs84: OK, let's keep it as it is.

javiergm1983:
Now all work fine for WP and UWP.

PR #506, opened 2016-12-31
New function to retry transfers

javiserrano:


include/megaapi.h
```diff
@@ -6551,6 +6551,22 @@ class MegaApi
         void cancelTransfer(MegaTransfer *transfer, MegaRequestListener *listener = NULL);
 
         /**
+         * @brief Retry a transfer
+         *
+         * This function allows to start a transfer based on a MegaTransfer object. It can be used,
+         * for example, to retry transfers that finished with an error. To do it, you can retain the
+         * MegaTransfer object in onTransferFinish and use it later with this function.
```
sergiohs84: You can do so by calling \c MegaTransfer::copy()

src/megaapi_impl.cpp
```diff
@@ -5688,7 +5688,36 @@ void MegaApiImpl::startStreaming(MegaNode* node, m_off_t startPos, m_off_t size,
 	transfer->setEndPos(startPos + size - 1);
 	transfer->setMaxRetries(maxRetries);
 	transferQueue.push(transfer);
-	waiter->notify();
+    waiter->notify();
```
sergiohs84: Indentation?

include/megaapi.h
sergiohs84: I would add:
`You can do so by calling \c MegaTransfer::copy()`

src/megaapi_impl.cpp
javiserrano: Not in the new code, but I will fix the lines above

include/megaapi.h
javiserrano: I agree

PR #507, opened 2017-01-03
Fetchnodes stats

javiserrano:


include/mega/megaclient.h
```diff
@@ -38,6 +38,136 @@
...
+     */
+    dstime timeToSyncsResumed;
+
+    /**
+     * @brief Time until the filesystem is current
+     *
+     * From DB: time until action packets have been processed
+     * From API: this time is the same as timeToResult
+     */
+    dstime timeToCurrent;
```
sergiohs84: Despite they can be merged into a unique variable, according to our discussion, it's probably worthy to keep two independent variables for an easier understanding of the meaning of each timestamp, without requiring to get into the underneath details of how the fetchnodes works in each situation.

src/megaclient.cpp
```diff
@@ -1159,6 +1159,13 @@ void MegaClient::exec()
                     case REQ_INFLIGHT:
                         if (pendingcs->contentlength > 0)
                         {
+                            if (fetchingnodes && fnstats.timeToFirstByte == NEVER
+                                    && pendingcs->bufpos > 10)
```
sergiohs84: Why `bufpos > 10`??

src/megaclient.cpp
```diff
@@ -1223,9 +1236,31 @@ void MegaClient::exec()
...
+                        {
+                            if (fetchingnodes)
+                            {
+                                fnstats.eagains++;
+                            }
+                        }
 
                     // fall through
                     case REQ_FAILURE:
+                        if (fetchingnodes)
```
sergiohs84: No need for 2 blocks, just `if (fetchingnodes && pendingcs->httpstatus != 200)` :)

src/megaclient.cpp
```diff
@@ -1340,9 +1375,31 @@ void MegaClient::exec()
                                 fetchnodes(true);
                                 reqtag = creqtag;
                             }
+                            else if (e == API_EAGAIN || e == API_ERATELIMIT)
+                            {
+                                if (!statecurrent)
```
sergiohs84: What does this case represent? Cannot we use the same `fetchingnodes` flag?

src/megaclient.cpp
```diff
@@ -1340,9 +1375,31 @@ void MegaClient::exec()
...
+                                {
+                                    fnstats.eagains++;
+                                }
+                            }
                         }
                         // fall through
                     case REQ_FAILURE:
+                        if (pendingsc && !statecurrent)
```
sergiohs84: We can use one single `if` block by adding the condition `&& pendingsc->httpstatus != 200` to this one

src/megaclient.cpp
```diff
@@ -7876,6 +7962,12 @@ bool MegaClient::fetchsc(DbTable* sctable)
 
     while (sctable->next(&id, &data, &key))
     {
+        if (fnstats.timeToFirstByte == NEVER)
```
sergiohs84: Maybe we should move the assignment of `timeToFirstByte` out of the loop, since it will iterate thousands of times and only once the condition will be true.

src/megaclient.cpp
```diff
@@ -8128,6 +8223,22 @@ void MegaClient::disabletransferresumption(const char *loggedoutid)
...
+        return;
+    }
+
+    WAIT_CLASS::bumpds();
+    fnstats.init();
+    if (sid.size() >= SIDLEN)
+    {
+        fnstats.type = FetchNodesStats::TYPE_ACCOUNT;
+    }
+    else if (publichandle != UNDEF)
```
sergiohs84: The way we identify if the session is logged into an account or a folder is duplicated many times along the codebase. We should prepare a couple of checker methods like: `bool MegaClient::isLoggedInFolder()` and `isLoggedInAccount()` or similar.

include/mega/megaclient.h
```diff
@@ -38,6 +38,136 @@
 
 namespace mega {
 
+class MEGA_API FetchNodesStats
+{
+public:
+    enum {
+        MODE_DB = 0,
```
javiserrano: I agree, but I think that it's better to not change it in this case since it doesn't have any impact (there isn't any of those checks) and the specs of the stats have been already agreed with the server side, so I would prefer to not have to change them if there isn't a strong reason.

include/mega/megaclient.h
```diff
@@ -38,6 +38,136 @@
...
+    int actionPackets;
+
+    ////////////
+    // Errors //
+    ////////////
+
+    /**
+     * @brief Number of error -3 or -4 received during the process (including cs and sc requests)
+     */
+    int eagains;
```
javiserrano: I agree, fixing...

include/mega/megaclient.h
javiserrano: Yes, I agree. It's possible to merge them without losing information, but the meaning on the value would have to be inferred from other values (`mode` and `actionPackets`). I think that. to keep the agreed interface and easy the understanding it's better to keep it for now.

src/megaclient.cpp
javiserrano: It's intended to not set the value when error codes are received (usually `-3`).
A limit of 3 should be enough, but I have set it to 10 just in case there are bigger error codes in the future.

src/megaclient.cpp
javiserrano: I agree, Changing it...

src/megaclient.cpp
javiserrano: `pendingsc` is the channel to receive action packets. `fetchingnodes` means that a fetchnodes command is being sent (for that reason it's used in `pendingcs`), but action packets can be required for the initialization of the filesystem even if a fetchnodes command is not sent (when there is a local cache). The intention is to measure the number of action packets processed before the filesystem is current, so the `statecurrent` flag is perfect for that purpose in all cases.

src/megaclient.cpp
javiserrano: I agree. I put it that way because it might be required in the future to process more things while the filesystem is not current, but it's unlikely to have to process more things depending on `httpstatus`. Anyway, it's just an assumption, it's not the case in this moment so I'm going to change it. 👍

src/megaclient.cpp
javiserrano: I agree

src/megaclient.cpp
javiserrano: I think that the best would be that all those checks could rely on MegaClient::loggedin(), creating a new value in the enumeration it returns. That would require a review of all usages of that function and changes not related to this pull request at all, so I think that it's better to change it in another one.

javiserrano:
Please, take a look specially at the changes in `MegaClient::fetchsc` in the last commit.

include/mega/megaclient.h
sergiohs84: OK. There's no other reason, so let's keep it as it is.

src/megaclient.cpp
sergiohs84: OK, good point :)

src/megaclient.cpp
sergiohs84: :pray: 
No comments

include/mega/megaclient.h
sergiohs84: OK. There's no other reason, so let's keep it as it is.

src/megaclient.cpp
sergiohs84: OK, let's add it to the TODO's list :+1:

PR #511, opened 2017-01-04
Adaptative chunk sizes for downloads

javiserrano:
- Chunk sizes depending on the file position, number of connections and remaining data
- Different max chunk sizes for mobile and desktop apps
- Extra parameter for chunkceil to ease its usage
- Compatibility with transfer resumption (backwards compatibility too)
- Verify transfers that were cached with all data already transferred


sergiohs84:
trigger compilation

sergiohs84:
Tests failed due to timeouts

PR #512, opened 2017-01-04
Improve the performance for syncs with many files

javiserrano:


sergiohs84:
Review the creation of a folder to check if the content is notified and processed.

sergiohs84:
Test a locked folder that adds a notification to the `DirNotify::RETRY` queue and check that the file is synced when it is unlocked. Maybe a new _syncup_ is required.

src/megaclient.cpp
```diff
@@ -1825,9 +1787,10 @@ void MegaClient::exec()
                                 if (((*it)->state == SYNC_ACTIVE || (*it)->state == SYNC_INITIALSCAN)
                                  && !(*it)->dirnotify->notifyq[DirNotify::DIREVENTS].size()
                                  && !(*it)->dirnotify->notifyq[DirNotify::RETRY].size()
-                                 && !syncadding)
+                                 && !syncadding && syncuprequired && !syncnagleretry)
```
sergiohs84: `syncuprequired` is per sync, but more than one sync might require a `syncup`. Hence, the flag should be per sync, maybe part of the `Sync` object. The global checkup would be a OR of every single flag.
Note that this issue could be affecting the `syncdownrequired` on the same way.

src/megaclient.cpp
```diff
@@ -1836,6 +1799,7 @@ void MegaClient::exec()
                             {
                                 syncnaglebt.backoff(nds - Waiter::ds);
                                 syncnagleretry = true;
+                                syncuprequired = true;
```
sergiohs84: Having a flag per sync instead of a global flag would make the line above unnecessary.

src/megaclient.cpp
```diff
@@ -2996,11 +2963,9 @@ bool MegaClient::procsc()
                                 // node update
                                 sc_updatenode();
 #ifdef ENABLE_SYNC
-                                if (!fetchingnodes && jsonsc.pos[1] != ']') // there are more packets
-                                {
-                                    applykeys();
-                                    return false;
-                                }
+                                // run syncdown() before continuing
```
sergiohs84: We must control that the AP is not part of a cached fetchnodes. If so, do not return false (do not force any sync scan)

src/megaclient.cpp
```diff
@@ -4718,10 +4682,9 @@ void MegaClient::notifypurge(void)
              && (*it)->localroot.node->changed.removed)
             {
                 delsync(*it);
+                syncactivity = true;
```
sergiohs84: `syncactivity` is set to _true_ in `delsync()`. No need it here

src/megaclient.cpp
javiserrano: This is a very severe bug. Great catch!

src/megaclient.cpp
javiserrano: You are right 👍

javiserrano:
`Review the creation of a folder to check if the content is notified and processed.`
Yes, it works because the filesystem notification is not removed from the queue until all parent nodes are created and the notification is properly processed. I have just tested it and it works.

javiserrano:
```Test a locked folder that adds a notification to the DirNotify::RETRY queue and check that the file is synced when it is unlocked. Maybe a new syncup is required.```
I have added a commit with a review of the management of the retry queue. I have tested it and it works as expected.

File: src/megaclient.cpp (hunk 1 above)
Comment (javiserrano) [inline src/megaclient.cpp]: I have used the same approach as in syncdown(). If any of the calls fails, the flag is set to keep retrying

**Comment by javiserrano** [comment]
Ready to be reviewed again!

File: src/file.cpp [hunk 5]
```diff
@@ -509,7 +509,14 @@ void SyncFileGet::updatelocalname()
 // add corresponding LocalNode (by path), then self-destruct
 void SyncFileGet::completed(Transfer*, LocalNode*)
 {
-    sync->checkpath(NULL, &localname);
+    LocalNode *ll = sync->checkpath(NULL, &localname);
+    if (ll && ll != (LocalNode*)~0)
+    {
+        LOG_debug << "LocalNode created, associating with remote Node";
+        ll->setnode(n);
```
sergiohs84: It would be good to check that the fingerprint, timestamp... of the localnode matches the ones in the node. Otherwise, it would mean that the file has been modified externally in the meantime.

src/file.cpp
```diff
@@ -509,7 +509,14 @@ void SyncFileGet::updatelocalname()
 // add corresponding LocalNode (by path), then self-destruct
 void SyncFileGet::completed(Transfer*, LocalNode*)
 {
-    sync->checkpath(NULL, &localname);
+    LocalNode *ll = sync->checkpath(NULL, &localname);
```
sergiohs84: Review the case when there were an existing localnode before, pointing to the same file. If so, check the former outdated localnode is deleted accordingly.

src/sync.cpp
```diff
@@ -894,8 +898,11 @@ LocalNode* Sync::checkpath(LocalNode* l, string* localpath, string* localname)
         {
             // fopen() signals that the failure is potentially transient - do
             // nothing and request a recheck
+            LOG_warn << "File blocked. Adding notification to the retry queue: " << path;
             dirnotify->notify(DirNotify::RETRY, ll, localpath->data(), localpath->size());
             client->syncfslockretry = true;
+            client->syncfslockretrybt.backoff(SCANNING_DELAY_DS);
+            client->blockedfile = path;
```
sergiohs84: Check the reported `blockedfile` is correct with multiple blocked files.

src/megaclient.cpp
```diff
@@ -1689,42 +1719,34 @@ void MegaClient::exec()
...
-                // set retry interval for locked filesystem items once all
-                // pending items were processed
-                if (syncfslockretry)
-                {
-                    syncfslockretrybt.backoff(2);
-                }
-
                 // notify the app of the length of the pending scan queue
-                if (totalpending < 4)
+                if (!scanningpending)
```
sergiohs84: Consider to keep the threshold  to avoid blinking notifications.

src/file.cpp
javiserrano: I have checked the case in which a file is replaced in the Cloud, and it works correctly

src/file.cpp
javiserrano: I have replied to this in the comment above

PR #514, opened 2017-01-08
Async filesystem layer for the SDK

javiserrano:
- Async filesystem layer and usage of async functions for transfers (from the `jssdk` branch)
- Allow to download data sequentially (from the `jssdk` branch)
- Better management of errors accessing files
- Compatibility with transfer resumption
- Simplified the code to manage downloads


javiserrano:
trigger compilation

src/transferslot.cpp
```diff
@@ -278,6 +321,13 @@ void TransferSlot::doio(MegaClient* client)
                     break;
 
                 case REQ_SUCCESS:
+                    if (client->orderdownloadedchunks && transfer->type == GET && transfer->progresscompleted != ((HttpReqDL *)reqs[i])->dlpos)
+                    {
+                        // postponing unsorted chunk
+                        p += reqs[i]->size;
```
sergiohs84: They do not need processing in that case.

src/transferslot.cpp
```diff
@@ -450,9 +540,140 @@ void TransferSlot::doio(MegaClient* client)
...
+                        LOG_verbose << "Processing finished async fs operation";
+                        if (!asyncIO[i]->failed)
+                        {
+                            if (transfer->type == PUT)
+                            {
+                                LOG_verbose << "Async read succeeded";
+                                m_off_t npos = ChunkedHash::chunkceil(asyncIO[i]->pos, transfer->size);
+
+                                string finaltempurl = tempurl;
+                                if (transfer->type == GET && client->usealtdownport
```
sergiohs84: This block is not necessary, since we're processing uploads, not downloads.

src/transferslot.cpp
```diff
@@ -450,9 +540,140 @@ void TransferSlot::doio(MegaClient* client)
...
+                                         asyncIO[i]->pos, npos);
+
+                                reqs[i]->pos = ChunkedHash::chunkfloor(asyncIO[i]->pos);
+                                reqs[i]->status = REQ_PREPARED;
+                            }
+                            else
+                            {
+                                LOG_verbose << "Async write succeeded";
+                                HttpReqDL *downloadRequest = (HttpReqDL *)reqs[i];
+                                for (chunkmac_map::iterator it = downloadRequest->chunkmacs.begin(); it != downloadRequest->chunkmacs.end(); it++)
```
sergiohs84: Maybe can use `map::insert()` instead of a loop

polmr:
trigger compilation

src/transferslot.cpp
javiserrano: I agree

src/transferslot.cpp
javiserrano: True!

src/transferslot.cpp
javiserrano: Yes, it's only needed to not process them until previous chunks finish

src/transferslot.cpp
```diff
@@ -68,59 +84,76 @@ TransferSlot::TransferSlot(Transfer* ctransfer)
...
+                m_off_t bufsize = downloadRequest->bufpos;
+                if (fa->fwrite(downloadRequest->buf, bufsize, dlpos))
+                {
+                    LOG_verbose << "Sync write succeeded";
+                    for (chunkmac_map::iterator it = downloadRequest->chunkmacs.begin(); it != downloadRequest->chunkmacs.end(); it++)
+                    {
+                        transfer->chunkmacs[it->first] = it->second;
+                    }
+                    downloadRequest->chunkmacs.clear();
+                    transfer->progresscompleted += bufsize;
```
javiserrano: The write above seems to be also wrong, despite it shouldn't have any consequence because the chunkmap is correctly managed in finalize(). I think that the most appropriate is to modify `bufpos` inside finalize(), because it's the function that manages the encryption. That way, this code would be automatically right.

src/transferslot.cpp
```diff
@@ -68,59 +84,76 @@ TransferSlot::TransferSlot(Transfer* ctransfer)
...
+                    }
 
-                    bufstart += chunksize;
-                    buflen -= chunksize;
-                    cachetransfer = true;
+                    if (!asyncIO[i]->failed)
+                    {
+                        LOG_verbose << "Async write succeeded";
+                        HttpReqDL *downloadRequest = (HttpReqDL *)reqs[i];
+                        for (chunkmac_map::iterator it = downloadRequest->chunkmacs.begin(); it != downloadRequest->chunkmacs.end(); it++)
```
javiserrano: I'm not sure if it would be more efficient, but it seems more appropriate.

src/transferslot.cpp
```diff
@@ -68,59 +84,76 @@ TransferSlot::TransferSlot(Transfer* ctransfer)
...
-                    m_off_t chunksize = endpos - startpos;
-                    reqs[i]->finalize(fa, &transfer->key, &transfer->chunkmacs, transfer->ctriv, startpos, endpos);
-                    transfer->progresscompleted += chunksize;
-                    LOG_debug << "Caching finished chunk: " << startpos << " - " << endpos << "   Size: " << chunksize;
-
-                    startpos = endpos;
-                    endpos = ChunkedHash::chunkceil(startpos, transfer->size);
+                    while (!asyncIO[i]->finished)
+                    {
+                        transfer->client->waiter->wait();
```
javiserrano: I will try and will post a comment with my results

src/transferslot.cpp
```diff
@@ -68,59 +84,76 @@ TransferSlot::TransferSlot(Transfer* ctransfer)
 TransferSlot::~TransferSlot()
 {
     if (transfer->type == GET && !transfer->finished
-            && transfer->progresscompleted != transfer->size)
+            && transfer->progresscompleted != transfer->size
+            && !transfer->asyncopencontext)
```
javiserrano: Yes

src/transfer.cpp
```diff
@@ -777,6 +785,12 @@ void Transfer::complete()
         LOG_debug << "Upload complete: " << (files.size() ? files.front()->name : "NO_FILES") << " " << files.size();
 
         // files must not change during a PUT transfer
+        if (slot->fa->asyncavailable())
+        {
+            slot->fa->closef();
+            slot->fa->fopen(&localfilename);
```
javiserrano: Yes, it's needed to open the file to be used synchronously there. Anyway, I have removed the `closef()` because that function should be for internal use only. Instead, I remove the object and create a new one.

src/posix/fs.cpp
```diff
@@ -132,6 +149,188 @@ void PosixFileAccess::sysclose()
...
+    {
+        context->userCallback(context->userData);
+    }
+    PosixFileAccess::asyncmutex.unlock();
+}
+
+void PosixFileAccess::asyncsysopen(AsyncIOContext *context)
+{
+    string path;
+    path.assign((char *)context->buffer, context->len);
```
javiserrano: Yes, it's the path of the file to open. It's not allowed to call it with a NULL path

src/posix/fs.cpp
```diff
@@ -132,6 +149,188 @@ void PosixFileAccess::sysclose()
...
+        return;
+    }
+
+    context->synchronizer = NULL;
+    context->failed = aio_error(aiocbp);
+    if (!context->failed)
+    {
+        if (context->op == AsyncIOContext::READ)
+        {
+            memset((void *)(((char *)(aiocbp->aio_buf)) + aiocbp->aio_nbytes), 0, context->pad);
```
javiserrano: I agree, I will add a check

src/posix/fs.cpp
```diff
@@ -41,8 +41,25 @@ namespace mega {
...
+    synchronizer = NULL;
+}
+
+PosixAsyncIOContext::~PosixAsyncIOContext()
+{
+    LOG_verbose << "Deleting PosixAsyncIOContext";
+    PosixFileAccess::asyncmutex.lock();
+    if (synchronizer)
+    {
+        synchronizer->context = NULL;
```
javiserrano: Exactly

src/megaclient.cpp
```diff
@@ -2463,16 +2471,64 @@ bool MegaClient::dispatch(direction_t d)
...
+            }
+            else
+            {
+                openok = (d == PUT)
+                        ? ts->fa->fopen(&nexttransfer->localfilename)
+                        : ts->fa->fopen(&nexttransfer->localfilename, false, true);
+                openfinished = true;
+            }
 
             // try to open file (PUT transfers: open in nonblocking mode)
```
javiserrano: I have moved it. I will push the change shortly

src/megaclient.cpp
```diff
@@ -2463,16 +2471,64 @@ bool MegaClient::dispatch(direction_t d)
...
+            {
+                if (!nexttransfer->asyncopencontext)
+                {
+                    LOG_debug << "Starting async open";
+                    nexttransfer->asyncopencontext = (d == PUT)
+                        ? ts->fa->asyncfopen(&nexttransfer->localfilename, true, false)
+                        : ts->fa->asyncfopen(&nexttransfer->localfilename, false, true, nexttransfer->size);
+                    asyncfopens++;
+                }
+
```
javiserrano: Yes, really asynchronous file opens are not fully supported yet. I have added a comment below to show what's missing. Anyway, all filesystem layers in the SDK always finishes asyncfopen synchronously.

src/megaclient.cpp
```diff
@@ -2463,16 +2471,64 @@ bool MegaClient::dispatch(direction_t d)
...
+                ts = nexttransfer->slot;
+            }
+
+            if (ts->fa->asyncavailable())
+            {
+                if (!nexttransfer->asyncopencontext)
+                {
+                    LOG_debug << "Starting async open";
+                    nexttransfer->asyncopencontext = (d == PUT)
+                        ? ts->fa->asyncfopen(&nexttransfer->localfilename, true, false)
```
javiserrano: This comment is very right. The synchronous access to the file for uploads wasn't blocking the source file during the whole upload. I have added that feature for the asynchronous access too.

javiserrano:
Finally, it seems that replacing the loop to update the `chunkmac_map` by an `insert` is not a good idea because `insert` don't replace existing elements and that would severely break the logic.

src/filesystem.cpp
```diff
@@ -265,6 +277,191 @@ void FileAccess::closef()
...
+    context->finished = true;
+    context->userCallback(context->userData);
+    return context;
+}
+
+bool FileAccess::asyncopenf()
+{
+    if (!localname.size())
+    {
+        return true;
```
javiserrano: This is the regular code path when the file is open using the fopen specific of each platform. The code below is intended to support file reads during uploads, without keeping the file open during the whole transfer.

src/filesystem.cpp
```diff
@@ -265,6 +277,191 @@ void FileAccess::closef()
...
+    else
+    {
+        LOG_err << "Error opening async file handle";
+    }
+    return result;
+}
+
+void FileAccess::asyncclosef()
+{
+    numops--;
```
javiserrano: It should never happen. `asyncopenf` should be always used at the beginning of an async operation, and `asyncclosef` at the end. If that doesn't happen, we would have a bug.

If we add that check, we could just call `sysclose` more times and we would also make the bug more difficult to detect, because otherwise it should be detected in the destructor of `FileAccess` with the assert:
`assert(!numops);`

src/megaapi_impl.cpp
```diff
@@ -3966,6 +3966,11 @@ char MegaApiImpl::userAttributeToScope(int type)
         case MegaApi::USER_ATTR_KEYRING:
             scope = '*';
             break;
+
+        default:
+            LOG_err << "Getting invalid scope";
+            scope = 0;
```
javiserrano: 👍

PR #517, opened 2017-01-09
Register tokens for push notifications

sergiohs84:


src/commands.cpp
```diff
@@ -4731,6 +4731,28 @@ void CommandChatPresenceURL::procresult()
     }
 }
 
+CommandRegisterPushNotification::CommandRegisterPushNotification(MegaClient *client, int deviceType, const char *token)
+{
+    this->client = client;
+    cmd("spt");
+    arg("p", deviceType);
+    arg("t", token);
+    notself(client);
```
javiserrano: Is it needed a `notself` here?
I mean, does this command generate an any action packet?

src/megaapi_impl.cpp
```diff
@@ -2657,6 +2657,7 @@ const char *MegaRequestPrivate::getRequestString() const
         case TYPE_CHAT_TRUNCATE: return "CHAT_TRUNCATE";
         case TYPE_SET_MAX_CONNECTIONS: return "SET_MAX_CONNECTIONS";
         case TYPE_CHAT_PRESENCE_URL: return "CHAT_PRESENCE_URL";
+        case TYPE_REGISTER_PUSH_NOTIFICATION: return "TYPE_REGISTER_PUSH_NOTIFICATION";
```
javiserrano: Remove the `TYPE_` prefix.

include/megaapi.h
```diff
@@ -1812,7 +1812,7 @@ class MegaRequest
             TYPE_GET_CANCEL_LINK, TYPE_CONFIRM_CANCEL_LINK,
             TYPE_GET_CHANGE_EMAIL_LINK, TYPE_CONFIRM_CHANGE_EMAIL_LINK,
             TYPE_CHAT_UPDATE_PERMISSIONS, TYPE_CHAT_TRUNCATE, TYPE_CHAT_SET_TITLE, TYPE_SET_MAX_CONNECTIONS,
-            TYPE_PAUSE_TRANSFER, TYPE_MOVE_TRANSFER, TYPE_CHAT_PRESENCE_URL,
+            TYPE_PAUSE_TRANSFER, TYPE_MOVE_TRANSFER, TYPE_CHAT_PRESENCE_URL, TYPE_REGISTER_PUSH_NOTIFICATION,
```
javiserrano: It seems a bit strange to not add the new value at the end of the enum, but since the iOS bindings seems to follow the same order, I think that it's OK.

sergiohs84: Yes, I think the position not at the end was due to a merge with develop, which already had a new type of request, IIRC.

src/megaapi_impl.cpp
sergiohs84: I'll ammend it.

src/commands.cpp
sergiohs84: I don't think so. I'll remove the `notself`

javiserrano:
Perfect!

PR #526, opened 2017-01-13
WIP: Change nodepath returned for root nodes other than /

polmr:


polmr:
this cannot be merged yet. full consistency with getNodeByPath should be provided

src/megaapi_impl.cpp
```diff
@@ -11218,8 +11218,12 @@ char* MegaApiImpl::getNodePath(MegaNode *node)
 	if (n->nodehandle == client->rootnodes[0])
 	{
 		path = "/";
-        sdkMutex.unlock();
-        return stringToArray(path);
+                if (strcmp("CRYPTO_ERROR",n->displayname()))
```
sergiohs84: In order to detect if client is logged in into a folder link, it would be better to use:
`if (!ISUNDEF(publichandle))`
Even better, add `isLoggedInFolder()` and replace it where appropriate (whereever the publichandle is checked to be undefined)

PR #529, opened 2017-01-21
Updating Zlib URL and MD5 to latest version

SkyrisBactera:
To avoid error (https://github.com/meganz/MEGAsync/issues/61) in which build_sdk.sh cannot  download Zlib, because a newer version has been created which moved 1.2.8 into http://zlib.net/fossils/


polmr:
thanks for the contribution @SkyrisBactera

Issue #201, opened 2015-12-26
error compiling sdk D:


latot:
Hi, here a little problem, Logs!

```
make -j9
make  all-recursive
make[1]: se entra en el directorio '/home/pipe/Documentos/git/sdk'
Making all in include
make[2]: se entra en el directorio '/home/pipe/Documentos/git/sdk/include'
make[2]: No se hace nada para 'all'.
make[2]: se sale del directorio '/home/pipe/Documentos/git/sdk/include'
make[2]: se entra en el directorio '/home/pipe/Documentos/git/sdk'
  CXX      src/src_libmega_la-megaclient.lo
  CXX      src/src_libmega_la-commands.lo
  CXX      src/src_libmega_la-http.lo
  CXX      src/src_libmega_la-json.lo
  CXX      src/src_libmega_la-node.lo
  CXX      src/src_libmega_la-pubkeyaction.lo
  CXX      src/src_libmega_la-serialize64.lo
  CXX      src/src_libmega_la-request.lo
  CXX      src/src_libmega_la-share.lo
src/commands.cpp: En la función miembro ‘virtual void mega::CommandGetUA::procresult()’:
src/commands.cpp:1967:22: error: ‘priv’ no se declaró en este ámbito
                 && ((priv && strncmp(attributename.c_str(), "prEd255", 7))
                      ^
  CXX      src/src_libmega_la-sharenodekeys.lo
  CXX      src/src_libmega_la-sync.lo
  CXX      src/src_libmega_la-transfer.lo
Makefile:1881: fallo en las instrucciones para el objetivo 'src/src_libmega_la-commands.lo'
make[2]: *** [src/src_libmega_la-commands.lo] Error 1
make[2]: *** Se espera a que terminen otras tareas....
src/megaclient.cpp: En la función miembro ‘int mega::MegaClient::inited25519()’:
src/megaclient.cpp:6810:76: error: no se encontró una función coincidente para la llamada a ‘mega::MegaClient::putua(const char [8], const byte*, unsigned int, int)’
     putua("prEd255", (const byte*)signkey.keySeed, crypto_sign_SEEDBYTES, 1);
                                                                            ^
src/megaclient.cpp:6810:76: nota: el candidato es:
src/megaclient.cpp:6037:6: nota: void mega::MegaClient::putua(const char*, const byte*, unsigned int)
 void MegaClient::putua(const char* an, const byte* av, unsigned avl)
      ^
[...]

PR #534, opened 2017-01-25
Minor adjustments

javiserrano:


src/posix/net.cpp
```diff
@@ -162,13 +162,20 @@ CurlHttpIO::CurlHttpIO()
     curl_multi_setopt(curlm[GET], CURLMOPT_SOCKETDATA, this);
     curl_multi_setopt(curlm[GET], CURLMOPT_TIMERFUNCTION, download_timer_callback);
     curl_multi_setopt(curlm[GET], CURLMOPT_TIMERDATA, this);
+#ifdef _WIN32
+    curl_multi_setopt(curlm[GET], CURLMOPT_MAXCONNECTS, 200);
```
sergiohs84: What happens if the SDK attempts to create the connection #201?

javiserrano: cURL just discards the oldest cached connection.
I have increased the limit on Windows because I noticed that it was already happening after few transfers. I haven't increased the limit for other platforms because limits of the operating system could conflict (specially OS X).

PR #541, opened 2017-01-31
thread: fix building with POSIX threads in MinGW

shinchiro:


polmr:
trigger compilation

polmr:
thanks for the contribution

PR #542, opened 2017-01-31
Assign a notification number to each callback related to transfers

javiserrano:


src/megaapi_impl.cpp
```diff
@@ -10347,6 +10360,8 @@ void MegaApiImpl::fireOnRequestTemporaryError(MegaRequestPrivate *request, MegaE
 void MegaApiImpl::fireOnTransferStart(MegaTransferPrivate *transfer)
 {
 	activeTransfer = transfer;
```
sergiohs84: The line above seems it has the wrong indentation

src/megaapi_impl.cpp
```diff
@@ -17372,6 +17397,7 @@ MegaTransferDataPrivate::MegaTransferDataPrivate(const MegaTransferDataPrivate *
     this->uploadTags = transferData->uploadTags;
     this->downloadPriorities = transferData->downloadPriorities;
     this->uploadPriorities = transferData->uploadPriorities;
+    this->notificationNumber = transferData->notificationNumber;
```
sergiohs84: In these cases, we usually use the setter:
`this->setNotificationNumber(transferData->notificationNumber)`

javiserrano: I have made it that way to keep it consistent with the lines above. Since it should be a bit more efficient and I can't see any strong reason to change all those assignments now, I think that we can continue with that approach there for now.

PR #543, opened 2017-02-01
Transfer management fixes

javiserrano:


src/megaapi_impl.cpp
```diff
@@ -17369,24 +17381,37 @@ vector<Node *> &PublicLinkProcessor::getNodes()
...
+        {
+            File *file = (*fit);
+            uploadTags.push_back(file->tag);
+            uploadPriorities.push_back(transfer->priority);
+        }
     }
-    this->notificationNumber = notificationNumber;
+    numUploads = uploadTags.size();
+
+    notificationNumber = notificationNumber;
```
sergiohs84: Add `this` to the first variable

PR #545, opened 2017-02-01
Add creation date to chat data

sergiohs84:


sergiohs84:
trigger compilation

src/utils.cpp
```diff
@@ -190,6 +193,9 @@ TextChat* TextChat::unserialize(class MegaClient *client, string *d)
     ou = MemAccess::get<handle>(ptr);
     ptr += sizeof ou;
 
+    ts = (uint32_t)MemAccess::get<time_t>(ptr);
```
javiserrano: I don't understand the types here.
`ts` is `m_time_t`. Shouldn't that type be used instead of `uint32_t` and `time_t` in this line and the line below?

src/megaclient.cpp
```diff
@@ -4814,6 +4818,10 @@ void MegaClient::sc_chatupdate()
                     chat->url = ""; // not received in action packets
                     chat->ou = ou;
                     chat->title = title;
+                    if (ts != -1)
```
javiserrano: I have a doubt about this.
In this case, `chat->ts` is not updated if it's not received. However, in other two places in this pull request, `chat->ts` is set to 0 when it's not received.
I think that it's OK, but just to be sure that it's the intended.

polmr:
trigger compilation

src/utils.cpp
sergiohs84: Yes, definitely. I was influenced by the datatype used internally by the chat library to represent this timestamp. I'll fix that.

src/megaclient.cpp
sergiohs84: Yes, that is made in purpose actually. Here, we are processing the AP. The AP will update the title, the privilege... but not the creation timestamp, which should not be reset to 0, but keep the existing one.
There are 2 cases where the action packet will effectively create the chat object: when the chat is a new one and when you are invited to an existing chat (you don't know about it before). In those cases, a new instance of `TextChat` will be created (by default, `ts=0` at ctor). The AP for those cases should include the `ts`, but other APs may not include it (currently all APs do, but it's prone to change after André's review) and we don't want to lose the existing `ts`

PR #566, opened 2017-02-15
Review of the logic to open and close files during uploads

javiserrano:


src/filesystem.cpp
```diff
@@ -309,44 +310,47 @@ AsyncIOContext *FileAccess::asyncfopen(string *f)
 
 bool FileAccess::asyncopenf()
 {
+    numAsyncReads++;
     if (!localname.size())
     {
         return true;
```
sergiohs84: I guess the answer is *no*, but... would it be possible to reach this point with `isAsyncOpened = false`??

src/filesystem.cpp
```diff
@@ -309,44 +310,47 @@ AsyncIOContext *FileAccess::asyncfopen(string *f)
 
 bool FileAccess::asyncopenf()
 {
+    numAsyncReads++;
```
sergiohs84: I'd increment this counter only when we're sure the file has been opened. If this function returns `false`, you probably don't want the counter to be incremented. Or maybe I'm wrong and it's possible to have _async reads_ with a file that is not opened?

polmr: this counter registers the number of attempts, regardless the success or failure

sergiohs84:
trigger compilation
It failed because: `Contact request creation not received after 300 seconds`

src/filesystem.cpp
javiserrano: Yes. The meaning is the one commented by @polmr , to control that `asyncopenf` is called the same number of times than `asyncclosef'  before the deletion of the `FileAccess` object.

To control if the file has been already opened or not, there is a new flag (`isAsyncOpened`)

src/filesystem.cpp
javiserrano: The answer is _yes_ 😄 
The first time that `asyncopenf` is called for a `FileAccess` object, the flag must be _false_ so the file can be opened. Next times the flag should be already _true_ to prevent that the file is opened several times (few lines below).

PR #567, opened 2017-02-15
Improved the detection and management of files modified during uploads

javiserrano:
- Update _mtime_ and _size_ of _FileAccess_ objects when changes are detected
- Correctly set the _retry_ flag when a read fails due to a file change
- Set the correct error code when individual files of a multifile transfer fail
- Detect modifications of the first file in a transfer at startup and during the upload
- Detect modifications of any file in a transfer at the end of the upload
- Never update the fingerprint of a _Transfer_ object
- Cancel the upload of modified files and complete others, if modifications are detected at the end of the upload.


javiserrano:
Needs testing

src/transfer.cpp
```diff
@@ -795,19 +817,51 @@ void Transfer::complete()
...
+                    slot->retrying = true;
+                    slot->retrybt.backoff(11);
+                    delete fa;
+                    return;
+                }
+            }
+
+            if (!isOpen || f->genfingerprint(fa))
+            {
+                LOG_warn << "Modification detected after upload";
```
polmr: Does this warning make sense if fopen failed? it could be clearer to have
if (!isOpen) LOG_warn << "Unable to check modification after upload" << endl;
else LOG_warn << "Modification detected after upload";

src/transfer.cpp
```diff
@@ -795,19 +817,51 @@ void Transfer::complete()
...
+#ifdef ENABLE_SYNC
+                if (f->syncxfer)
+                {
+                    client->syncdownrequired = true;
+                }
+#endif
+                client->filecachedel(f);
+                files.erase(it++);
```
polmr: For clarifying, I'd take this it++ out and remove the else, and do always it++ down

src/transfer.cpp
javiserrano: The deletion could be considered as a "modification" for the upload 😄 but I will change the log so it provides more information 👍

src/transfer.cpp
javiserrano: The iterator can't be incremented before the erase, nor after because it would be invalid. That's why it's there. Another option would be to create a copy of the iterator (to be used inside the `if`), increment it (before the `if`). But I'm not sure if that approach is more clear than the current one.

src/megaclient.cpp
```diff
@@ -2600,7 +2600,12 @@ bool MegaClient::dispatch(direction_t d)
 
                 if (d == PUT)
                 {
-                    nexttransfer->size = ts->fa->size;
+                    if (ts->fa->mtime != nexttransfer->mtime || ts->fa->size != nexttransfer->size)
+                    {
+                        LOG_warn << "Modification detected starting upload";
```
polmr: Minor issue:
"Modification detected" is warned twice for changes detected on startup (one for starting & one "during"):
15:59:47 (warn):  Modification detected starting upload (megaclient.cpp:2601)
15:59:47 (debug): Transfer failed with error -21 (transfer.cpp:359)
16:00:24 (warn):  Modification detected during active upload (transfer.cpp:395)

PR #570, opened 2017-02-16
Update cache and notify chats

sergiohs84:
- Added the notification of chats updated via commands (it was totally missing)
 - Update chats in cache according to the commands result
 - Allow to differentiate between own changes and external changes in `MegaTextChat` notified objects

### DIFF
--- a/src/commands.cpp
+++ b/src/commands.cpp
@@ -4467,7 +4467,6 @@ void CommandChatCreate::procresult()
     }
     else
     {
-        string url;
         handle chatid = UNDEF;
         int shard = -1;
         bool group = false;
@@ -4476,10 +4475,6 @@ void CommandChatCreate::procresult()
         {
             switch (client->json.getnameid())
             {
-                case MAKENAMEID3('u','r','l'):
-                    client->json.storeobject(&url);
-                    break;
-
                 case MAKENAMEID2('i','d'):
                     chatid = client->json.gethandle(MegaClient::CHATHANDLE);
                     break;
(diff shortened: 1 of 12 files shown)


src/commands.cpp
```diff
@@ -4503,12 +4498,14 @@ void CommandChatCreate::procresult()
                         TextChat *chat = client->chats[chatid];
                         chat->id = chatid;
                         chat->priv = PRIV_MODERATOR;
-                        chat->url = url;
                         chat->shard = shard;
                         delete chat->userpriv;  // discard any existing `userpriv`
                         chat->userpriv = this->chatPeers;
                         chat->group = group;
 
+                        chat->setTag(tag ? tag : -1);
```
javiserrano: I think that this check and other similar ones in this pull request could be removed because the tag `0` shouldn't be used (except for commands generated internally by the SDK without any interest for client apps and libraries), but I understand that they ensure that `isOwnChange()` never returns `0` for own changes, even if a wrong tag is set.

src/commands.cpp
```diff
@@ -4814,16 +4807,28 @@ void CommandChatUpdatePermissions::procresult()
...
                 if (upvit->first == uh)
                 {
                     chat->userpriv->erase(upvit);
                     chat->userpriv->push_back(userpriv_pair(uh, priv));
+                    found = true;
                     break;
                 }
             }
+
+            if (!found)
```
javiserrano: Is this valid for changes in the own privileges?

polmr:
trigger compilation (failed due to a exceded contact invitations threshold)

src/commands.cpp
sergiohs84: Exactly. Currently, the `tag` is never set to `0` since the SDK does not generate chat-related commands by itself. However, if one day we start doing it, we are already protected.

polmr:
trigger compilation

src/commands.cpp
sergiohs84: Damn!! Actually updating your own privilege will not work properly. Fixing...

polmr:
trigger compilation

PR #571, opened 2017-02-19
Megacmd

be47liberty:


polmr:
This will be merged in `develop` in time first so as not to break our development cycle.

PR #574, opened 2017-02-20
Fix orphan users and `purgenodesuserabortsc()`

sergiohs84:


src/megaclient.cpp
```diff
@@ -6667,17 +6667,16 @@ void MegaClient::mapuser(handle uh, const char* email)
         // yes: add email reference
         u = &users[hit->second];
 
-        if (strcmp(u->email.c_str(), email))
-        { 
-            if (u->email.size())
-            {
-                umindex.erase(u->email);
```
javiserrano: I think that the new approach doesn't take into account email changes and wouldn't remove the element related to the previous email from `umindex`.

sergiohs84: You're right! Fixing...

polmr:
trigger compilation (failed due to a exceded contact invitations threshold)

polmr:
trigger compilation

sergiohs84:
trigger compilation, Jenkins :)

src/megaclient.cpp
```diff
@@ -6667,17 +6667,27 @@ void MegaClient::mapuser(handle uh, const char* email)
...
-            Node::copystring(&u->email, email);
-            umindex[nuid] = hit->second;
+        string useremail;
+        Node::copystring(&useremail, u->email.c_str());
+        transform(useremail.begin(), useremail.end(), useremail.begin(), ::tolower);
+
+        // if mapping a different email, remove old index
+        if (strcmp(useremail.c_str(), nuid.c_str()))
+        {
+            umindex.erase(u->email);
```
javiserrano: If we assume that `u->email` can have uppercase characters, this would be wrong because keys in `umindex` seems to be always lowercase. I think that the same applies to `MegaClient::discarduser`. 

Instead of fixing it (or additionally), another option would be to ensure that `u->email` is always lowercase. In that case, it would be needed to modify `MegaClient::finduser`. But even in that case, it could be appropriate to use `useremail` instead of `u->email` here for consistency.

polmr:
trigger compilation

PR #599, opened 2017-03-28
Avoid reset asymmkey when it's not set

sergiohs84:
If key is reset more than once while logged out (no logins performed in
between), the second attempt will result in a segfault.


sergiohs84:
Rejected! this is an issue from karere library, which deletes the AsymmCipher object which belongs to the SDK

PR #602, opened 2017-04-03
Feature5445 send contacts files

alber2510:
Send contacts and files. Feature #5445
- Add new MegaNode to MegaNodeList
- Return MegaClient from MegaApi

trigger compilation


alber2510:
trigger compilation

polmr:
trigger compilation

alber2510:
trigger compilation

sergiohs84:
This PR needs to be created for `develop` instead of `master`

PR #608, opened 2017-04-04
Effectively delete inactive users

sergiohs84:
since keeping them as inactive users may cause conflicts when a new user
is registered under the same email address than the cancelled account.


sergiohs84:
I tested to create a folder link, to create outgoing and incoming shares, to send a file to a contact's Inbox and to share a folder with a non-contact. Everything works as expected

PR #613, opened 2017-04-07
Add funtionality to feature #5445 - Add support to share contacts & f…

alber2510:
…iles

### DIFF
--- a/include/megaapi.h
+++ b/include/megaapi.h
@@ -71,7 +71,6 @@ class MegaContactRequestList;
 class MegaShareList;
 class MegaTransferList;
 class MegaApi;
-class MegaClient;
 
 class MegaSemaphore;
 
@@ -9446,7 +9445,7 @@ class MegaApi
          */
         MegaTextChatList *getChatList();
 
-        virtual std::string getFileAttribute(MegaHandle handle);
+        virtual const char* getFileAttribute(MegaHandle handle);
 
 #endif
 
--- a/include/megaapi.h
+++ b/include/megaapi.h
@@ -9445,6 +9445,12 @@ class MegaApi
          */
         MegaTextChatList *getChatList();
 
+        /**
+         * @brief Get files attributes from a node
+         * You take the ownership of the returned value
+         * @param handle handle from node
+         * @return char array with files attributes from the node.
+         */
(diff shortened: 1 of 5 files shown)


include/megaapi.h
```diff
@@ -71,6 +71,7 @@ class MegaContactRequestList;
 class MegaShareList;
 class MegaTransferList;
 class MegaApi;
+class MegaClient;
```
javiserrano: `MegaClient` belongs to the low level API. It can't be used in the public interface of the intermediate layer.

include/megaapi.h
```diff
@@ -9406,6 +9419,9 @@ class MegaApi
          * @return A list of MegaTextChat objects with detailed information about each chatroom.
          */
         MegaTextChatList *getChatList();
+
+        virtual std::string getFileAttribute(MegaHandle handle);
```
javiserrano: Please use `char *` or `const char *` if possible for a better compatibility with SWIG.

include/megaapi_impl.h
```diff
@@ -1673,6 +1675,8 @@ class MegaApiImpl : public MegaApp
         MegaClient *getMegaClient();
         static FileFingerprint *getFileFingerprintInternal(const char *fingerprint);
 
+        virtual std::string getFileAttribute(MegaHandle handle);
```
javiserrano: If the corresponding function in the public object is changed. Change this one accordingly.

src/megaapi_impl.cpp
```diff
@@ -2916,7 +2916,27 @@ MegaNode *MegaNodeListPrivate::get(int i)
...
+    return s;
+}
+
+void MegaNodeListPrivate::addNode(MegaNode *node)
+{
+    MegaNodeList *duplicatedList = copy();
+
+    s = 0;
+    for(int i=0; i<s; i++)
+        delete list[i];
```
javiserrano: This is very dangerous. According to the documentation, internal MegaNode objects are supposed to be valid until the deletion of the MegaNodeList and this breaks that behavior. It should be avoided (for example by creating a new list of pointers but reusing the pointers from the previous list) or, if there isn't any other possible solution, the documentation should be updated to say that the usage of this function invalidate all internal MegaNode objects.

PR #617, opened 2017-04-10
Fix memory leak

jnavarrom:


sergiohs84:
This issue is already solved in the branch `feature5545-send-contacts-files` that is about to be merged.
Thanks for reporting anyway

PR #630, opened 2017-04-17
Don't consider the own user as inactive

javiserrano:
Despite the API provides that visibility for it

### DIFF
--- a/src/megaclient.cpp
+++ b/src/megaclient.cpp
@@ -3438,7 +3438,7 @@ void MegaClient::updatesc()
             for (user_vector::iterator it = usernotify.begin(); it != usernotify.end(); it++)
             {
                 char base64[12];
-                if ((*it)->show == INACTIVE)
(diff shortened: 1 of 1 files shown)


src/megaclient.cpp
```diff
@@ -6728,7 +6728,7 @@ void MegaClient::mapuser(handle uh, const char* email)
         u = &users[hit->second];
 
         um_map::iterator mit = umindex.find(nuid);
-        if (mit != umindex.end() && mit->second != hit->second && users[mit->second].show != INACTIVE)
+        if (mit != umindex.end() && mit->second != hit->second && (users[mit->second].show != INACTIVE || users[mit->second].userhandle == me))
```
polmr: Shouldn't this be +        if (mit != umindex.end() && mit->second != hit->second && (users[mit->second].show != INACTIVE **&&** users[mit->second].userhandle **!=** me))

PR #635, opened 2017-04-22
URL Changed on server!

lundell:
URL Changed on server!


polmr:
This has already been taken care in `https://github.com/meganz/sdk/pull/586/files`. It was not the `.12` version that was moved to old/ but the one used before `.8`. Thanks anyway

PR #637, opened 2017-04-25
v3.1.0

javiserrano:


RayPylant:
Thank you for getting those kinks out of the way it helps a lot thanks again.

PR #640, opened 2017-04-26
Scan after own changes in the account

javiserrano:


RayPylant:
i have made the changes and reloaded its ok ty.

PR #642, opened 2017-04-27
Update WP & UWP bindings

javiergm1983:
- Check if the temporal bandwidth usage is valid after an overquota error.
- Allow change the API URL.
- Allow to set the language code.


RayPylant:
viewed awesome ty.

PR #644, opened 2017-04-28
Check if the available transfer quota is enough

javiserrano:


javiergm1983:
trigger compilation

include/mega/megaclient.h
```diff
@@ -277,6 +277,9 @@ class MEGA_API MegaClient
     // retrieve user details
     void getaccountdetails(AccountDetails*, bool, bool, bool, bool, bool, bool);
 
+    // check if the available bandwidth quota is enough to transfer an amount of bytes
+    void querybandwidthquota(m_off_t size);
+
```
sergiohs84: Maybe we should refactorize all these to be _transfer quota_, which is the actual name of the quota and how it's displayed to users.

src/commands.cpp
```diff
@@ -3229,6 +3229,26 @@ void CommandGetUserQuota::procresult()
...
+    tag = client->reqtag;
+}
+
+void CommandQueryBandwidthQuota::procresult()
+{
+    if (!client->json.isnumeric())
+    {
+        LOG_err << "Unexpected response: " << client->json.pos;
+        client->json.storeobject();
+        return client->app->querybandwidthquota_result(0);
```
sergiohs84: Shouldn't we return `API_EINTERNAL` in this case??

src/megaapi_impl.cpp
```diff
@@ -10197,6 +10206,16 @@ void MegaApiImpl::account_details(AccountDetails*, error e)
     fireOnRequestFinish(request, megaError);
 }
 
+void MegaApiImpl::querybandwidthquota_result(int code)
+{
+    if(requestMap.find(client->restag) == requestMap.end()) return;
+    MegaRequestPrivate* request = requestMap.at(client->restag);
+    if(!request || (request->getType() != MegaRequest::TYPE_QUERY_BANDWIDTH_QUOTA)) return;
+
```
sergiohs84: If we return in the `Command::procresult()` an `API_EINTERNAL` (or any other error code), I think we should not set the flag in the request unless the code is `API_OK`

bindings/wp8/MegaSDK.cpp
```diff
@@ -1843,6 +1843,16 @@ void MegaSDK::getExtendedAccountDetails(bool sessions, bool purchases, bool tran
     megaApi->getExtendedAccountDetails(sessions, purchases, transactions);
 }
 
+void MegaSDK::queryBandwidthQuota(int64 size, MRequestListenerInterface^ listener)
+{
+    megaApi->queryBandwidthQuota(size, createDelegateMRequestListener(listener));
```
sergiohs84: This MegaApi function has been refactored to `queryTransferQuota()`. You need to change this call in the binding.

bindings/wp8/MegaSDK.cpp
```diff
@@ -1843,6 +1843,16 @@ void MegaSDK::getExtendedAccountDetails(bool sessions, bool purchases, bool tran
...
+}
+
+void MegaSDK::queryBandwidthQuota(int64 size)
+{
+    megaApi->queryBandwidthQuota(size);
```
sergiohs84: Same, need to refactor.

javiserrano:
I think that it's OK now

PR #653, opened 2017-05-04
Allow to get previews/thumbnails for attached nodes in chats

sergiohs84:
Already tested by @jnavarrom

### DIFF
--- a/include/megaapi.h
+++ b/include/megaapi.h
@@ -5683,6 +5683,8 @@ class MegaApi
          * - MegaRequest::getText - Returns the file attribute string if \c node is an attached node from chats. NULL otherwise
          * - MegaRequest::getFile - Returns the destination path
          * - MegaRequest::getParamType - Returns MegaApi::ATTR_TYPE_THUMBNAIL
+         * - MegaRequest::getBase64Key - Returns the nodekey in Base64 (only when node has file attributes)
+         * - MegaRequest::getPrivateKey - Returns the file-attribute string (only when node has file attributes)
          *
          * @param node Node to get the thumbnail
          * @param dstFilePath Destination path for the thumbnail.
@@ -5706,6 +5708,8 @@ class MegaApi
          * - MegaRequest::getText - Returns the file attribute string if \c node is an attached node from chats. NULL otherwise
          * - MegaRequest::getFile - Returns the destination path
          * - MegaRequest::getParamType - Returns MegaApi::ATTR_TYPE_PREVIEW
+         * - MegaRequest::getBase64Key - Returns the nodekey in Base64 (only when node has file attributes)
+         * - MegaRequest::getPrivateKey - Returns the file-attribute string (only when node has file attributes)
          *
          * @param node Node to get the preview
          * @param dstFilePath Destination path for the preview.
--- a/include/megaapi.h
+++ b/include/megaapi.h
@@ -886,6 +886,19 @@ class MegaNode
         virtual std::string* getAttrString();
 
         /**
+         * @brief Returns a string that contains the file attributes related to the node
+         *
+         * The return value is only valid for nodes attached in a chatroom. In all other cases this function
+         * will return an empty string.
+         *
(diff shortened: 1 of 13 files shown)


include/megaapi.h
```diff
@@ -886,6 +886,19 @@ class MegaNode
         virtual std::string* getAttrString();
 
         /**
+         * @brief Returns a string that contains the file attributes related to the node
+         *
+         * The return value is only valid for nodes attached in a chatroom. In all other cases this function
+         * will return an empty string.
```
sergiohs84: The above is not always true, since there are more cases where the file attribute string will be copied from the original `Node`. However, only for attachments it's guaranteed that fileattribute is not empty.

src/megaclient.cpp
```diff
@@ -3719,7 +3724,7 @@ bool MegaClient::slotavail() const
 bool MegaClient::moretransfers(direction_t d)
 {
     m_off_t r = 0;
-    int total = 0;
+    unsigned int total = 0;
```
sergiohs84: My compiler throws a warning: comparing signed with unsigned value :)

src/megaapi_impl.cpp
```diff
@@ -12905,11 +12937,38 @@ void MegaApiImpl::sendPendingRequests()
...
+                e = API_EARGS;
+                break;
+            }
 
-            e = client->getfa(node, (fatype)type);
+            string fileattrstring;
+            string key;
+            if (!fa)
+            {
+                fileattrstring = node->fileattrstring;
```
polmr: does it ever get there?

src/megaapi_impl.cpp
```diff
@@ -85,10 +85,11 @@ MegaNodePrivate::MegaNodePrivate(const char *name, int type, int64_t size, int64
     this->nodehandle = nodehandle;
     this->parenthandle = parentHandle;
     this->attrstring.assign(attrstring->data(), attrstring->size());
+    this->fileattrstring = fileattrstring;
     this->nodekey.assign(nodekey->data(), nodekey->size());
     this->changed = 0;
-    this->thumbnailAvailable = false;
-    this->previewAvailable = false;
+    this->thumbnailAvailable = (Node::hasfileattribute(&fileattrstring, 0) != 0);
+    this->previewAvailable = (Node::hasfileattribute(&fileattrstring, 1) != 0);
```
sergiohs84: Fixed there and other places where 0-1 was used instead of define

src/megaapi_impl.cpp
```diff
@@ -68,7 +68,7 @@
 using namespace mega;
 
 MegaNodePrivate::MegaNodePrivate(const char *name, int type, int64_t size, int64_t ctime, int64_t mtime, uint64_t nodehandle,
-                                 string *nodekey, string *attrstring, const char *fingerprint, MegaHandle parentHandle,
+                                 string *nodekey, string *attrstring, string fileattrstring, const char *fingerprint, MegaHandle parentHandle,
```
sergiohs84: Agree

src/megaapi_impl.cpp
sergiohs84: Yes. Some MegaNode may not have been initialized with the `fileattrstring` because it's not really needed. In those cases, the request is created without the `setText()`, so `fa == NULL` and both `fileattrstring` and `nodekey` need to be copied from the node retrieved via `nodebyhandle()`.

src/megaapi_impl.cpp
```diff
@@ -5072,7 +5084,17 @@ void MegaApiImpl::getNodeAttribute(MegaNode *node, int type, const char *dstFile
...
 
     request->setParamType(type);
-    if(node) request->setNodeHandle(node->getHandle());
+    if(node)
+    {
+        request->setNodeHandle(node->getHandle());
+        if (!node->getFileAttrString()->empty())
+        {
+            request->setText(node->getFileAttrString()->c_str());
+            const char *nodekey = node->getBase64Key();
```
sergiohs84: Done!

PR #654, opened 2017-05-09
Fix out-of-sync `User::notified`

sergiohs84:
The `usernotify` vector is cleared but the object corresponding to the
own user is kept in memory with the `notified` flag to true, so it never
gets notified again.

Already tested by @carolzato


sergiohs84:
trigger compilation

PR #655, opened 2017-05-09
Manage the list of users with granted access to a node in a chatroom 

sergiohs84:


sergiohs84:
The changes in this branch add the `mcna` data to the cached data, keeping backwards compatibility. However, users who attached files via chat with previous versions will not know the `mcna` list. As a consequence, the revoke of access to old attachments may actually not work as expected until they do a full reload.

src/utils.cpp
```diff
@@ -173,14 +197,62 @@ TextChat* TextChat::unserialize(class MegaClient *client, string *d)
...
+    if (hasAttachments)
+    {
+        unsigned short numNodes = 0;
+        if (ptr + sizeof numNodes > end)
+        {
+            delete userpriv;
+            return NULL;
+        }
+
+        numNodes = MemAccess::get<short>(ptr);
```
javiserrano: The type of `numNodes` is `unsigned short`.

src/utils.cpp
```diff
@@ -173,14 +197,62 @@ TextChat* TextChat::unserialize(class MegaClient *client, string *d)
...
+            if (ptr + sizeof h + sizeof numUsers > end)
+            {
+                delete userpriv;
+                return NULL;
+            }
+
+            h = MemAccess::get<handle>(ptr);
+            ptr += sizeof h;
+
+            numUsers = MemAccess::get<short>(ptr);
```
javiserrano: The type of `numUsers` is `unsigned short`.

src/utils.cpp
```diff
@@ -226,6 +300,32 @@ void TextChat::resetTag()
...
+    {
+        attachments_map::iterator uhit = attachedNodes.find(h);
+        if (uhit != attachedNodes.end())
+        {
+            uhit->second.erase(uh);
+            if (uhit->second.empty())
+            {
+                attachedNodes.erase(h);
+                changed.attachments = true;
+                return true;
```
javiserrano: Shouldn't the function return true if the `erase` succeeded even if other users have still access to the node?

src/megaclient.cpp
```diff
@@ -4932,6 +4937,79 @@ void MegaClient::sc_chatupdate()
...
+                {
+                    if (chats.find(chatid) == chats.end())
+                    {
+                        chats[chatid] = new TextChat();
+                    }
+
+                    TextChat *chat = chats[chatid];
+                    if (r)  // access revoked
+                    {
+                        if(!chat->setNodeUserAccess(h, uh, true))
```
javiserrano: Please check the comment in `setNodeUserAccess()` related to the return value of the function.

src/utils.cpp
sergiohs84: You're very right. Thanks

src/megaclient.cpp
```diff
@@ -7825,6 +7904,66 @@ void MegaClient::procmcf(JSON *j)
...
+                        readingNode = false;
+                    }
+                    break;
+                }
+            }
+            j->leaveobject();
+        }
+    }
+
+    j->leavearray();
```
sergiohs84: Fixed!

src/utils.cpp
```diff
@@ -82,7 +84,27 @@ bool TextChat::serialize(string *d)
     d->append((char*)&ou, sizeof ou);
     d->append((char*)&ts, sizeof(ts));
 
-    d->append("\0\0\0\0\0\0\0\0\0", 10); // additional bytes for backwards compatibility
+    char hasAttachments = attachedNodes.size();
```
sergiohs84: Correct. It would be very bad luck to overflow to zero, but it could happen. Fixed!

PR #658, opened 2017-05-11
Serialization of MegaNode

sergiohs84:
required by Android to pass MegaNodes from the app to the download service

### DIFF
--- a/src/megaapi_impl.cpp
+++ b/src/megaapi_impl.cpp
@@ -378,6 +378,20 @@ MegaNode *MegaNodePrivate::copy()
     return new MegaNodePrivate(this);
 }
 
+char *MegaNodePrivate::serialize()
+{
+    string d;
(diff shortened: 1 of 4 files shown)


src/megaapi_impl.cpp
```diff
@@ -378,6 +378,20 @@ MegaNode *MegaNodePrivate::copy()
...
 
+char *MegaNodePrivate::serialize()
+{
+    string d;
+    if (!serialize(&d))
+    {
+        return NULL;
+    }
+
+    char *ret = new char[d.size()*4/3+3];
```
polmr: I believe it to be safer with `+ 4`:
`char *ret = new char[d.size() * 4 / 3 + 4];`

sergiohs84: Checking the code, 3 is more than enough for the `btoa()` implementation :)

PR #664, opened 2017-05-19
Add webclient-compatibility fix for `sigPubk`

sergiohs84:
Signatures of public RSA keys generated by weclient use 4 bytes for
serialization of exponent, while the SDK uses the minimum required
bytes, which results on verification failure for accounts confirmed in
webclient.

### DIFF
--- a/src/crypto/cryptopp.cpp
+++ b/src/crypto/cryptopp.cpp
@@ -466,12 +466,12 @@ void AsymmCipher::resetkey()
 
 void AsymmCipher::serializekeyforjs(string& d, bool fixedSize)
 {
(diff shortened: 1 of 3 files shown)


src/crypto/cryptopp.cpp
```diff
@@ -464,9 +464,37 @@ void AsymmCipher::resetkey()
...
+        c = 0;
+        for (unsigned j = 0; j < 4 - sizeE; j++)
+        {
+            d.append(&c, sizeof c);
+        }
+    }
+
+    for (int j = sizeE; j--;)
+    {
+        c = key[1].GetByte(j);  // returns 0 if out-of-range
```
polmr: better `key[PUB_E]`

PR #673, opened 2017-05-31
Allow to disable logs in console

sergiohs84:


include/megaapi.h
```diff
@@ -5479,6 +5479,8 @@ class MegaApi
          */
         static void setLogLevel(int logLevel);
 
+        static void setLogToConsole(bool enable);
```
polmr: missing docstrings saying that this determines if log is printed in console in the absence of a logger object. By default true. And it is likely to change in the future.

PR #681, opened 2017-06-07
Version 3.1.3

javiserrano:


include/mega/version.h
```diff
@@ -5,5 +5,5 @@
 #define MEGA_MINOR_VERSION 1
 #endif
 #ifndef MEGA_MICRO_VERSION
-#define MEGA_MICRO_VERSION 0
+#define MEGA_MICRO_VERSION 3
```
sergiohs84: Ooops, sorry. I forgot to update this number when prepared the v3.1.2

src/megaclient.cpp
```diff
@@ -10692,8 +10689,7 @@ bool MegaClient::startxfer(direction_t d, File* f, bool skipdupes)
             if (overquotauntil && overquotauntil > Waiter::ds)
             {
                 dstime timeleft = overquotauntil - Waiter::ds;
-                app->transfer_failed(t, API_EOVERQUOTA, timeleft);
-                t->bt.backoff(timeleft);
+                t->failed(API_EOVERQUOTA, timeleft);
```
sergiohs84: This block is also used for `Transfers` not found in the map of transfers, the `if` block of this `else`. We probably should change both.

javiserrano: Fixed!

PR #683, opened 2017-06-14
Account confirmation AP including email address

sergiohs84:
Pending for changes in API (include the email, not the uh, in the AP)


bindings/qt/QTMegaGlobalListener.cpp
```diff
@@ -43,6 +43,22 @@ void QTMegaGlobalListener::onReloadNeeded(MegaApi *api)
...
 }
 
+void QTMegaGlobalListener::onEvent(MegaApi *api, MegaEvent *e)
+{
+    switch(e->getType())
+    {
+        case MegaEvent::EVENT_ACCOUNT_CONFIRMATION:
+        {
+            QTMegaEvent *event = new QTMegaEvent(api, (QEvent::Type)QTMegaEvent::OnEvent);
+            event->setEvent(e ? e->copy() : NULL);
```
javiserrano: This doesn't make sense because if `e` is NULL the code would crash some lines above.

bindings/qt/QTMegaListener.cpp
```diff
@@ -111,6 +111,22 @@ void QTMegaListener::onReloadNeeded(MegaApi *api)
...
 }
 
+void QTMegaListener::onEvent(MegaApi *api, MegaEvent *e)
+{
+    switch(e->getType())
+    {
+        case MegaEvent::EVENT_ACCOUNT_CONFIRMATION:
+        {
+            QTMegaEvent *event = new QTMegaEvent(api, (QEvent::Type)QTMegaEvent::OnEvent);
+            event->setEvent(e ? e->copy() : NULL);
```
javiserrano: This doesn't make sense because if `e` is NULL the code would crash some lines above.

examples/megacli.cpp
```diff
@@ -4740,6 +4740,14 @@ void DemoApp::account_updated()
     }
 }
 
+void DemoApp::notify_confirmation(const char *email)
+{
+    if (client->loggedin() == EPHEMERALACCOUNT)
+    {
+        LOG_debug << "Account has been confirmed with email " + string(email) + ". Proceed to login with credentials.";
```
javiserrano: The `string()` constructor could be removed if `<<`  is used instead of `+`. That would also prevent a crash if `email` is NULL.

PR #690, opened 2017-06-20
Support archive chats

alber2510:


include/megaapi.h
```diff
@@ -1966,10 +1966,10 @@ class MegaRequest
             TYPE_GET_RECOVERY_LINK, TYPE_QUERY_RECOVERY_LINK, TYPE_CONFIRM_RECOVERY_LINK,
             TYPE_GET_CANCEL_LINK, TYPE_CONFIRM_CANCEL_LINK,
             TYPE_GET_CHANGE_EMAIL_LINK, TYPE_CONFIRM_CHANGE_EMAIL_LINK,
-            TYPE_CHAT_UPDATE_PERMISSIONS, TYPE_CHAT_TRUNCATE, TYPE_CHAT_SET_TITLE, TYPE_SET_MAX_CONNECTIONS,
-            TYPE_PAUSE_TRANSFER, TYPE_MOVE_TRANSFER, TYPE_CHAT_PRESENCE_URL, TYPE_REGISTER_PUSH_NOTIFICATION,
-            TYPE_GET_USER_EMAIL, TYPE_APP_VERSION, TYPE_GET_LOCAL_SSL_CERT, TYPE_SEND_SIGNUP_LINK,
-            TOTAL_OF_REQUEST_TYPES
+            TYPE_CHAT_UPDATE_PERMISSIONS, TYPE_CHAT_TRUNCATE, TYPE_CHAT_SET_TITLE, TYPE_CHAT_ARCHIVE,
```
sergiohs84: Unfortunately, we need to respect the order of the enumeration. Just add this new type at the end of it.

src/commands.cpp
```diff
@@ -5129,6 +5129,52 @@ void CommandRegisterPushNotification::procresult()
         client->app->registerpushnotification_result(API_EINTERNAL);
     }
 }
+
+CommandArchiveChat::CommandArchiveChat(MegaClient *client, handle chatid, bool archive)
+{
+    this->client = client;
```
sergiohs84: No need to assign the client here, since it's already set at the parent class `Command`

src/commands.cpp
```diff
@@ -5129,6 +5129,52 @@ void CommandRegisterPushNotification::procresult()
...
+void CommandArchiveChat::procresult()
+{
+    if (client->json.isnumeric())
+    {
+        error e = (error) client->json.getint();
+        if (e == API_OK)
+        {
+            if (client->chats.find(mChatid) == client->chats.end())
+            {
+                // the invitation succeed for a non-existing chatroom
```
sergiohs84: Update the comment here: "invitation" by "archive"

src/utils.cpp
```diff
@@ -84,9 +84,12 @@ bool TextChat::serialize(string *d)
     d->append((char*)&ou, sizeof ou);
     d->append((char*)&ts, sizeof(ts));
 
+    char isArchive = archive ? 1 : 0;
+    d->append((char*)&isArchive, 1);
+
```
sergiohs84: Better to place this new flag after the byte of `hasAttachments` (and before the empty bytes for compatibility), since very likely the _attachment code_ will be deployed before than the _archive code_

src/utils.cpp
```diff
@@ -197,10 +201,12 @@ TextChat* TextChat::unserialize(class MegaClient *client, string *d)
     ts = MemAccess::get<m_time_t>(ptr);
     ptr += sizeof(m_time_t);
 
+    archive = MemAccess::get<char>(ptr);
+    ptr += sizeof(char);
```
sergiohs84: If the byte is moved at serialization, move it here too

src/utils.cpp
```diff
@@ -277,6 +283,14 @@ TextChat* TextChat::unserialize(class MegaClient *client, string *d)
     chat->ou = ou;
     chat->resetTag();
     chat->ts = ts;
+    if (archive == 0)
```
sergiohs84: There's implicit casting for integers to boolean. No need the if-else

PR #695, opened 2017-06-22
Add class MEGAEvent, on event callback and fast send signup link meth…

jnavarrom:
…od (Obj-C)

### DIFF
--- a/bindings/ios/Private/DelegateMEGAGlobalListener.mm
+++ b/bindings/ios/Private/DelegateMEGAGlobalListener.mm
@@ -22,6 +22,7 @@
 #import "MEGAUserList+init.h"
 #import "MEGANodeList+init.h"
 #import "MEGAContactRequestList+init.h"
+#import "MEGAEvent+init.h"
 
 using namespace mega;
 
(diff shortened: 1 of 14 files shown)


bindings/ios/Private/DelegateMEGAGlobalListener.mm
```diff
@@ -97,3 +97,24 @@
         });
     }
 }
+
+void DelegateMEGAGlobalListener::onEvent(mega::MegaApi *api, mega::MegaEvent *event) {
+    switch (event->getType()) {
+        case MegaEvent::EVENT_ACCOUNT_CONFIRMATION:            
+            if (listener != nil && [listener respondsToSelector:@selector(onEvent:event:)]) {
+                MegaEvent *tempEvent = NULL;
+                if(event) {
```
sergiohs84: FYI, you don't really need to check `event`, you always receive a valid instance, like the `MegaError` or the `MegaRequest` at `onRequestFinish()`

bindings/ios/Private/DelegateMEGAListener.mm
```diff
@@ -193,3 +194,25 @@
...
     }
 }
+
+
+void DelegateMEGAListener::onEvent(mega::MegaApi *api, mega::MegaEvent *event) {
+    switch (event->getType()) {
+        case MegaEvent::EVENT_ACCOUNT_CONFIRMATION:            
+            if (listener != nil && [listener respondsToSelector:@selector(onEvent:event:)]) {
+                MegaEvent *tempEvent = NULL;
+                if(event) {
```
sergiohs84: Same here, `event` is always valid.

PR #697, opened 2017-06-22
(obsolete) New features related to MEGAchat

javiserrano:
- Support for GET requests and DNS queries
- Support for timeout for individual network requests
- Support to select the number of retries allowed for a network request
- Support for timeout for network requests (including retries)
- Public key pinning for MEGAchat servers
- Support to download files via HTTP or HTTPS
- Support for GeLB requests
- Support to send chat stats
- Documentation for the new requests

### DIFF
--- a/src/megaapi_impl.cpp
+++ b/src/megaapi_impl.cpp
@@ -2711,6 +2711,7 @@ const char *MegaRequestPrivate::getRequestString() const
         case TYPE_APP_VERSION: return "APP_VERSION";
         case TYPE_GET_LOCAL_SSL_CERT: return "GET_LOCAL_SSL_CERT";
         case TYPE_SEND_SIGNUP_LINK: return "SEND_SIGNUP_LINK";
+        case TYPE_QUERY_DNS: return "QUERY_DNS";
     }
     return "UNKNOWN";
 }
@@ -7434,6 +7435,14 @@ void MegaApiImpl::getLocalSSLCertificate(MegaRequestListener *listener)
     waiter->notify();
 }
 
+void MegaApiImpl::queryDNS(char *hostname, MegaRequestListener *listener)
+{
+    MegaRequestPrivate *request = new MegaRequestPrivate(MegaRequest::TYPE_QUERY_DNS, listener);
+    request->setName(hostname);
+    requestQueue.push(request);
+    waiter->notify();
+}
+
 const char *MegaApiImpl::getUserAgent()
 {
     return client->useragent.c_str();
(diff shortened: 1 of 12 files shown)


src/megaapi_impl.cpp
```diff
@@ -9893,6 +9932,65 @@ void MegaApiImpl::notify_confirmation(const char *email)
...
+    request->setTotalBytes(size);
+    if (request->getType() == MegaRequest::TYPE_QUERY_GELB
+            || request->getType() == MegaRequest::TYPE_CHAT_STATS
+            || request->getType() == MegaRequest::TYPE_QUERY_DNS)
+    {
+        string result;
+        result.assign((const char *)data, size);
+        request->setText(result.c_str());
+    }
+    else if (request->getType() == MegaRequest::TYPE_DOWNLOAD_FILE)
```
sergiohs84: This `else` can only be because of this type or request, so we don't really to check it here.

src/megaapi_impl.cpp
```diff
@@ -9893,6 +9932,65 @@ void MegaApiImpl::notify_confirmation(const char *email)
...
+    {
+        return;
+    }
+
+    request->setNumber(httpCode);
+    request->setTotalBytes(size);
+    if (request->getType() == MegaRequest::TYPE_QUERY_GELB
+            || request->getType() == MegaRequest::TYPE_CHAT_STATS
+            || request->getType() == MegaRequest::TYPE_QUERY_DNS)
+    {
```
sergiohs84: Shouldn't we add a check for the error code here?

src/megaapi_impl.cpp
```diff
@@ -9893,6 +9932,65 @@ void MegaApiImpl::notify_confirmation(const char *email)
...
+
+            if (size)
+            {
+                if (!f->fwrite((const byte*)data, size, 0))
+                {
+                    delete f;
+                    fireOnRequestFinish(request, MegaError(API_EWRITE));
+                    return;
+                }
+            }
```
sergiohs84: In case the response has `size=0`, we return `API_OK` but no file would be written

src/megaclient.cpp
```diff
@@ -993,6 +999,73 @@ void MegaClient::exec()
...
+            while (it != pendinghttp.end())
+            {
+                GenericHttpReq *req = it->second;
+                switch (req->status)
+                {
+                case REQ_FAILURE:
+                    if (!req->httpstatus && (!req->maxretries || (req->numretry + 1) < req->maxretries))
+                    {
+                        req->numretry++;
+                        req->status = REQ_PREPARED;
```
sergiohs84: If a request is prepared for a retry, could be execute it without a new iteration of the `exec()` loop?

src/megaapi_impl.cpp
javiserrano: I agree. Initially, I just passed the httpcode to the callback and for the DNS query is irrelevant what's the code. But with the current code, I think that it would be good to check it.

src/megaapi_impl.cpp
javiserrano: Yes, I have put the check just in case we add more request types related to this callbacks in the future, to not forget about it. But it can be removed now, of course.

src/megaapi_impl.cpp
javiserrano: mmm, in that case the file is deleted, it's open for writing and closed, so a zero-byte file should be created, shouldn't it?

sergiohs84: Correct. The `fopen()` will create the 0-byte file. No problems here then.

src/megaclient.cpp
javiserrano: The backoff timer will wake up the SDK to execute the retry when needed, even if it's immediately.

PR #708, opened 2017-06-29
New features related to MEGAchat

sergiohs84:


include/megaapi.h
```diff
@@ -9004,6 +9005,61 @@ class MegaApi
...
+         *
+         * Valid data in the MegaRequest object received in onRequestFinish when the error code
+         * is MegaError::API_OK:
+         * - MegaRequest::getNumber - Return the HTTP status code from the GeLB server
+         * - MegaRequest::getText - Returns the JSON response from the GeLB server
+         * - MegaRequest::getTotalBytes - Returns the number of bytes in the response
+         *
+         * @param service Service to check
+         * @param timeoutms Timeout for the request, including all possible retries
+         * @param listener MegaRequestListener to track this request
```
polmr: missing @param maxretries

src/megaclient.cpp
```diff
@@ -993,6 +999,73 @@ void MegaClient::exec()
...
+                                req->post(this);
+                                break;
+                            case METHOD_NONE:
+                                req->dns(this);
+                                break;
+                        }
+                        it++;
+                        break;
+                    }
+                    // no retry -> fall through
```
polmr: missing it++ & break here?

javiserrano: It is on purpose to check if the maximum time available for the request has expired during the time between retries

bindings/ios/3rdparty/different_address.patch
```diff
@@ -0,0 +1,182 @@
...
+-    }
+-
+   if (channel->sock_create_cb)
+     {
+       int err = channel->sock_create_cb(s, SOCK_DGRAM,
+@@ -1244,34 +1227,6 @@
+   return 1;
+ }
+ 
+-static int same_address(struct sockaddr *sa, struct ares_addr *aa)
```
javiserrano: I agree, but it helps to check that it's not used elsewhere.

include/megaapi.h
javiserrano: I'm going to add it.

PR #721, opened 2017-07-04
Ensure that JNI doesn't generate CESU-8 characters

javiserrano:


bindings/megaapi.i
```diff
@@ -92,25 +84,41 @@ JNIEXPORT jint JNICALL JNI_OnLoad(JavaVM *jvm, void *reserved)
...
+        $1[$1_size] = '\0';
+    }
+%}
+
+%typemap(freearg) char*
+%{
+    if ($1)
+    {
+        delete [] $1;
+        jenv->DeleteLocalRef($1_array);
```
sergiohs84: This free can be moved above.

PR #724, opened 2017-07-06
Protection against cloned Volume IDs on Windows

javiserrano:
Improved the management of file overwrites

### DIFF
--- a/src/sync.cpp
+++ b/src/sync.cpp
@@ -782,7 +782,7 @@ LocalNode* Sync::checkpath(LocalNode* l, string* localpath, string* localname)
                         || (it->second->sync == parent->sync)
                         || ((fp1 = it->second->sync->dirnotify->fsfingerprint())
                             && (fp2 = parent->sync->dirnotify->fsfingerprint())
-                            && (fp1 == fp2))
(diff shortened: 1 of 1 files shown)


polmr:
trigger compilation

src/sync.cpp
```diff
@@ -661,24 +661,32 @@ LocalNode* Sync::checkpath(LocalNode* l, string* localpath, string* localname)
...
-                                        fsfp_t fp1, fp2;
-                                        if (l->sync == it->second->sync
-                                            || ((fp1 = l->sync->dirnotify->fsfingerprint())
-                                                && (fp2 = it->second->sync->dirnotify->fsfingerprint())
-                                                && (fp1 == fp2)))
-                                        {
-                                            delete it->second;
-                                        }
+                                        l->mtime = -1;  // trigger change detection
+                                        delete it->second;   // delete old LocalNode
```
javiserrano: If a false `fsid` match is detected, not only delete the previous `LocalNode` as before, but also trigger a change detection for the current `LocalNode`, because the `fsid` doesn't match so it's needed to, at least, update it.

PR #730, opened 2017-07-10
Create avatar

javiserrano:


src/megaapi_impl.cpp
```diff
@@ -6371,12 +6371,34 @@ bool MegaApiImpl::createPreview(const char *imagePath, const char *dstPath)
...
 }
 
+bool MegaApiImpl::createAvatar(const char *imagePath, const char *dstPath)
+{
+    if (!gfxAccess)
+    {
+        return false;
+    }
+    
+    string utf8ImagePath = imagePath;
```
sergiohs84: Should we protect `imagePaht` / `dstPath` being `NULL` ? it's directly received from the app.

src/megaapi_impl.cpp
```diff
@@ -6371,12 +6371,34 @@ bool MegaApiImpl::createPreview(const char *imagePath, const char *dstPath)
     fsAccess->path2local(&utf8DstPath, &localDstPath);
 
     sdkMutex.lock();
-    bool result = gfxAccess->savefa(&localImagePath, GfxProc::PREVIEW1000x1000, &localDstPath);
+    bool result = gfxAccess->savefa(&localImagePath, 1000, 1000, &localDstPath);
```
sergiohs84: It would be better to use `GfxProc::dimensions[GfxProc::PREVIEW1000x1000][0]` and `GfxProc::dimensions[GfxProc::PREVIEW1000x1000][1]` instead of hardcoded dimensions. Or maybe adapt the size declarations to something more usable with the new prototype of `savefa()`

src/megaapi_impl.cpp
```diff
@@ -6371,12 +6371,34 @@ bool MegaApiImpl::createPreview(const char *imagePath, const char *dstPath)
...
+    string utf8ImagePath = imagePath;
+    string localImagePath;
+    fsAccess->path2local(&utf8ImagePath, &localImagePath);
+    
+    string utf8DstPath = dstPath;
+    string localDstPath;
+    fsAccess->path2local(&utf8DstPath, &localDstPath);
+    
+    sdkMutex.lock();
+    bool result = gfxAccess->savefa(&localImagePath, 250, 0, &localDstPath);
```
sergiohs84: Same as before, replace the hardcoded value by a definition

javiserrano:
The last commits should fix the problems

PR #737, opened 2017-07-13
Get children files and folders independently

sergiohs84:
A new `MegaChildrenLists` wrapper has been created in order to avoid two passes through the children vector, which would be required in case of two different APIs for files and folders.

### DIFF
--- a/bindings/wp8/MChildrenLists.cpp
+++ b/bindings/wp8/MChildrenLists.cpp
@@ -1,0 +1,52 @@
+/**
+* @file MChildrenLists.cpp
+* @brief Lists of file and folder children MegaNode objects.
+*
+* (c) 2013-2014 by Mega Limited, Auckland, New Zealand
+*
+* This file is part of the MEGA SDK - Client Access Engine.
+*
+* Applications using the MEGA API must present a valid application key
+* and comply with the the rules set forth in the Terms of Service.
+*
+* The MEGA SDK is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+*
+* @copyright Simplified (2-clause) BSD License.
+*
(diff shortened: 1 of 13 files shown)


bindings/wp8/MChildrenLists.cpp
```diff
@@ -0,0 +1,52 @@
...
+}
+
+MChildrenLists^ MChildrenLists::copy()
+{
+    return childrenLists ? ref new MChildrenLists(childrenLists->copy(), true) : nullptr;
+}
+
+MNodeList^ MChildrenLists::getFolderList()
+{
+    return childrenLists ? ref new MNodeList(childrenLists->getFolderList(), true) : nullptr;
```
javiserrano: `getFolderList()` doesn't give the ownership of the returned list, so the second parameter should be `false` (and the app should take care to not use the returned `MNodeList` after deleting the `MChildrenList`) or a  copy of the list should be created to pass it to the constructor. The second is more inefficient but also more secure.

bindings/wp8/MChildrenLists.cpp
```diff
@@ -0,0 +1,52 @@
...
+}
+
+MNodeList^ MChildrenLists::getFileList()
+{
+    return childrenLists ? ref new MNodeList(childrenLists->getFileList(), true) : nullptr;
```
javiserrano: `getFileList()` doesn't give the ownership of the returned list, so the second parameter should be `false` (and the app should take care to not use the returned `MNodeList` after deleting the `MChildrenList`) or a  copy of the list should be created to pass it to the constructor. The second is more inefficient but also more secure.

bindings/wp8/MChildrenLists.cpp
javiergm1983: Requested changes done.

bindings/wp8/MChildrenLists.cpp
javiergm1983: Requested changes done.

PR #739, opened 2017-07-22
Fix android example

javiserrano:


sergiohs84:
Tested on Ubuntu machine, everything working as expected. I built the APK, it loads and launches the app in Android phone :)

PR #752, opened 2017-08-07
Sync fixes

javiserrano:


src/megaclient.cpp
```diff
@@ -9913,6 +9913,15 @@ bool MegaClient::syncup(LocalNode* l, dstime* nds)
                                 RegCloseKey(hKey);
                             }
                         }
+
+                        lpath.append((char *)L":OECustomProperty", 34);
```
sergiohs84: No idea about the purpose of this tailing string for the filename

src/node.cpp
```diff
@@ -1073,8 +1073,11 @@ void LocalNode::setnameparent(LocalNode* newparent, string* newlocalpath)
         }
     }
 
-    LocalTreeProcUpdateTransfers tput;
-    sync->client->proclocaltree(this, &tput);
+    if (newlocalpath)
```
sergiohs84: Why this new restriction?

src/megaclient.cpp
javiserrano: Windows Search (enabled by default on Windows) adds this ADS to .eml and .pst files just after they are downloaded. Unfortunately that changes the modification time of the files too and Windows Search doesn't restore it.

src/node.cpp
javiserrano: If the path of the LocalNode hasn't changed, there is no need to update related Transfers. Also, that can cause crashes in some corner cases.

src/win32/fs.cpp
```diff
@@ -660,9 +663,35 @@ bool WinFileSystemAccess::istransientorexists(DWORD e)
...
+    int r = 0;
+#ifndef WINDOWS_PHONE
+    for (set<WinDirNotify*>::iterator it = dirnotifys.begin(); it != dirnotifys.end(); it++)
+    {
+        if ((*it)->enabled)
+        {
+            DWORD bytes = 0;
+            if (GetOverlappedResult((*it)->hDirectory, &((*it)->overlapped), &bytes, FALSE))
+            {
+                r = 1;
```
sergiohs84: Use `Waiter::NEEDEXEC` better than `1`

PR #756, opened 2017-08-09
Exclusion lists for synced folders

javiserrano:
This pull request is intended to add support to exclude relative paths from the synchronization instead of just file / folder names.
Additionally, it introduces a new exclusion method based on regular expressions. The new method can be enabled with `USE_PCRE`, but it's not expected to be 100% stable yet and it's very problable that the interface of the object `MegaRegExp` changes in few time to support different types of regular expressions and a whitelist.


src/megaapi_impl.cpp
```diff
@@ -15993,6 +16209,230 @@ void MegaSyncPrivate::setState(int state)
...
+    return regExps.size();
+}
+
+const char * MegaRegExpPrivate::getRegExp(int index)
+{
+    return regExps.at(index).c_str();
+}
+
+/**
+ * @brief Checks if the given regular expression is
```
polmr: *is correct

include/megaapi.h
```diff
@@ -7945,6 +8044,36 @@ class MegaApi
...
+        MegaSync *getSyncByNode(MegaNode *node);
+
+        /**
+         * @brief getSyncByPath Get the syncrhonization associated with a local path
+         * @param localPath Root local path of the synchronization
+         * @return Synchronization with the specified root local path
+         */
+        MegaSync *getSyncByPath(const char *localPath);
+
+#ifdef USE_PCRE
```
javiserrano: Name changed 👍

javiserrano: I agree, it's already planned to add support for wildcards in MegaRegExp.

src/megaapi_impl.cpp
```diff
@@ -6401,6 +6437,14 @@ bool MegaApiImpl::isSyncable(const char *path, long long size)
 
     bool result = false;
     sdkMutex.lock();
+    if (size >= 0)
+    {
+        if (!is_syncable(size))
+        {
+            return false;
```
polmr: missing mutex unlock

PR #766, opened 2017-08-18
Check if it's possible to start syncrhronizing a folder node

javiserrano:


include/megaapi.h
```diff
@@ -7956,6 +7956,22 @@ class MegaApi
...
+         * - MegaError::API_ENOENT if the node doesn't exist in the account
+         * - MegaError::API_EARGS if the node is NULL or is not a folder
+         * - MegaError::API_EACCESS if the node doesn't have full access
+         * - MegaError::API_EEXIST if there is a conflicting synchronization (nodes can't be synced twice)
+         * - MegaError::API_EINCOMPLETE if the SDK hasn't been built with support for synchronization
+         *
+         * @param Folder node to check
+         * @return MegaError::API_OK if the node is syncable, otherwise it returns an error.
+         */
+        int isNodeSyncable(MegaNode *node);
```
sergiohs84: Why not to call it `isFolderSyncable()`?

src/megaclient.cpp
```diff
@@ -9592,6 +9588,30 @@ error MegaClient::addsync(string* rootpath, const char* debris, string* localdeb
...
+    }
+    return API_OK;
+#else
+    return API_EINCOMPLETE;
+#endif
+}
+
+// check sync path, add sync if folder
+// disallow nested syncs (there is only one LocalNode pointer per node), return
+// EEXIST otherwise
```
sergiohs84: The errorcode in the comment is outdated with this changes. It can be EACCESS, EARGS, EINCOMPLETE, EEXIST.

include/megaapi.h
javiserrano: It could be confused with a local folder, but it would be a valid name too.

PR #767, opened 2017-08-18
Support for MEGA achievements

sergiohs84:


javiserrano:
There is a missing `delete achievementsDetails` in `~MegaRequestPrivate()`

javiserrano:
It's needed to implement `MegaAchievementsDetailsPrivate::copy`.

src/megaapi_impl.cpp
```diff
@@ -19445,3 +19529,158 @@ MegaChildrenListsPrivate::MegaChildrenListsPrivate()
...
+                strcpy(buf, it->c_str());
+                data.push_back(buf);
+
+                it++;
+            }
+            return new MegaStringListPrivate(data.data(), data.size());
+        }
+    }
+
+    return NULL;
```
javiserrano: It would be good to document that this function can return NULL in megaapi.h, or return an empty MegaStringList here,

src/megaapi_impl.cpp
```diff
@@ -19445,3 +19529,158 @@ MegaChildrenListsPrivate::MegaChildrenListsPrivate()
...
+    if (index < details.awards.size())
+    {
+        if (details.awards.at(index).achievement_class == MEGA_ACHIEVEMENT_INVITE)
+        {
+            vector<char*> data;
+            char *buf;
+            vector<string>::iterator it = details.awards.at(index).emails_invited.begin();
+            while (it != details.awards.at(index).emails_invited.end())
+            {
+                buf = new char[it->size()];
```
javiserrano: It shouldn't be (size() +1) for the `\0`.
Anyway, it's probably better to use `MegaApi::strdup()`.

src/megaapi_impl.cpp
```diff
@@ -9083,6 +9140,21 @@ void MegaApiImpl::getlocalsslcertificate_result(m_time_t ts, string *certdata, e
...
+{
+    MegaError megaError(e);
+    if(requestMap.find(client->restag) == requestMap.end()) return;
+    MegaRequestPrivate* request = requestMap.at(client->restag);
+    if(!request || (request->getType() != MegaRequest::TYPE_GET_ACHIEVEMENTS)) return;
+
+    if (!e)
+    {
+
+    }
```
javiserrano: This doesn't seem to be needed.

src/megaapi_impl.cpp
```diff
@@ -3942,6 +3973,17 @@ char *MegaApiImpl::getMyXMPPJid()
...
 }
 
+bool MegaApiImpl::isAchievementsEnabled()
+{
+    bool ret;
+
+    sdkMutex.lock();
+    ret = client->achievements_enabled;
+    sdkMutex.unlock();
+
```
javiserrano: I think that we don't need the mutex here.

sergiohs84: Removed the mutex lock/unlock

src/megaapi_impl.cpp
sergiohs84: True! :)

src/megaapi_impl.cpp
sergiohs84: Changed to `strdup()`

src/megaapi_impl.cpp
sergiohs84: Changed to an empty `MegaStringList`

PR #775, opened 2017-08-22
Delayed notifications for buggy network filesystems

javiserrano:
Some shared folders shared with Samba provide notifications *before* files are changed, instead of *after* when files are modified several times in a row. These changes create filesystem notifications to check new or modified files 15 seconds after they are detected to get possible newer changes.


src/filesystem.cpp
```diff
@@ -201,7 +201,7 @@ void DirNotify::notify(notifyqueue q, LocalNode* l, const char* localpath, size_
         delete fa;
     }
 
-    if (q == DirNotify::DIREVENTS)
+    if (q <= DirNotify::DIREVENTS)
```
sergiohs84: For readability and to avoid future conflicts with new queues we may add in the future, I'd change this condition to:
```if (q == DirNotify::DIREVENTS || q == DirNotify::EXTRA)```

javiserrano: I agree :+1:

PR #785, opened 2017-08-26
Added MegaApi constructor with multiple interfaces.

lemourin:
MegaApi(const char* appKey, Waiter* waiter, HttpIO* httpio,
        FileSystemAccess* fileAccess,
        DbAccess* dbAccess, MegaGfxProcessor* processor,
        const char *basePath = NULL, const char *userAgent = NULL);

The new constructor allows to supply the custom interfaces for http io,
file access and data base access.


polmr:
trigger compilation

polmr:
trigger compilation

PR #798, opened 2017-09-05
New getters for MEGA achievements

javisanesp:


src/megaapi_impl.cpp
```diff
@@ -19842,6 +19842,16 @@ int MegaAchievementsDetailsPrivate::getRewardsCount()
...
 }
 
+int MegaAchievementsDetailsPrivate::getRewardAwardId(unsigned int index)
+{
+    if (index < details.rewards.size())
+    {
+        return details.rewards.at(index).award_id;
+    }
+    
+    return 0;
```
sergiohs84: Better to return -1 to signal the index was not found, so you cannot use the returned value as a valid award id

javisanesp: 👍

PR #799, opened 2017-09-06
Delay the update of user attr until fetchnodes completion

sergiohs84:


polmr:
```
  CXX      src/src_libmega_la-commands.lo
src/commands.cpp: In member function 'virtual void mega::CommandPutUA::procresult()':
src/commands.cpp:2487:13: error: 'app' was not declared in this scope
             app->putua_result(API_EACCESS);
```

PR #801, opened 2017-09-06
Supporting LibreSSL

jbgg:
LibreSSL doesn't have RSA_get0_key function.


javiserrano:
Thank you very much! I have merged your changes into `develop`, that is our development branch.
In normal circumstances, we never merge any pull request directly into `master`.
These changes will reach `master` and soon as we release a new SDK version (usually every two weeks or so).

PR #809, opened 2017-09-12
Save memory avoiding to save short names if it's not needed

javiserrano:
Also, when it's not needed, a string is not allocated.

### DIFF
--- a/src/node.cpp
+++ b/src/node.cpp
@@ -937,9 +937,11 @@ void LocalNode::setnameparent(LocalNode* newparent, string* newlocalpath)
         // remove existing child linkage
         parent->children.erase(&localname);
 
-        if (slocalname.size())
+        if (slocalname)
         {
-            parent->schildren.erase(&slocalname);
+            parent->schildren.erase(slocalname);
+            delete slocalname;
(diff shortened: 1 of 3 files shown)


src/node.cpp
```diff
@@ -1047,9 +1049,18 @@ void LocalNode::setnameparent(LocalNode* newparent, string* newlocalpath)
         // (we don't construct a UTF-8 or sname for the root path)
         parent->children[&localname] = this;
 
-        if (sync->client->fsaccess->getsname(newlocalpath, &slocalname))
+        if (!slocalname)
         {
-            parent->schildren[&slocalname] = this;
+            slocalname = new string();
+        }
+        if (sync->client->fsaccess->getsname(newlocalpath, slocalname) && slocalname->size() && *slocalname != localname)
```
sergiohs84: It seems that `slocalname->size()` is always true if `getsname()` returned true. Maybe the check is not needed.

javiserrano: I agree. I added it basically to protect the SDK against a malfunction of `getsname` that is platform-specific.

PR #810, opened 2017-09-13
Fix jpeg images preview/thumbnail generation

javiergm1983:
Fix the preview/thumbnail generation of jpeg images when some rotation information is present in the image's metadata.


javiergm1983:
trigger compilation

sergiohs84:
trigger compilation

PR #813, opened 2017-09-14
WIP: Update bindings (Obj-C)

javisanesp:


javisanesp:
Please, @sergiohs84 and @jnavarrom review and mergue, whenever you can.

javiserrano:
`this` in the callback belongs to the intermediate layer and is deleted in `removeMEGAXXXDelegate()` functions (it's not managed by the garbage collector). So keeping `this` instead of `this->listener` would cause problems if the pointer has been already deleted. A possible solution would be to queue the deletion of those listeners using `dispatch_async`.

brookskhan:
Rigan can

javisanesp:
Somehow deprecated, so closing it 🙂

Issue #552, opened 2017-02-06
Python, Error: AttributeError: module 'mega' has no attribute 'MegaApi'


skar404:
When you build the lib for Python 3 does not work `import` from the example. But it works if you use `from mega import mega` then everything works.

The original text:

> При сборке lib для Python 3 не работает `import` из примера. Но при этом если использовать `from mega import mega` то все работает. 

### Error: 
```Python
>>> import mega
>>> mega.MegaApi('test')
AttributeError: module 'mega' has no attribute 'MegaApi'
```

### Correct: 
```Python
>>> from mega import mega
>>> mega.MegaApi('test')
<mega.mega.MegaApi; proxy of <Swig Object of type 'mega::MegaApi *' at 0x7fb3051fe060> >
```
---

OC Ubuntu 16.04.1 x64, Python v3.5.2,  MegaLib v3.0.8

Building: 
```shell
git clone https://github.com/meganz/sdk
cd ./sdk/

sh autogen.sh
./configure --disable-silent-rules \
--enable-python \
--disable-examples \
--with-python3

make
sudo make install
cd bindings/python/
python3 setup.py bdist_wheel
cd dist/
pip3 install megasdk-2.8.0-py2.py3-none-any.whl
```

polmr:
I belive this branch solves that: https://github.com/meganz/sdk/tree/fixpython3

polmr:
In the end this was solved with: https://github.com/meganz/sdk/pull/832 . Already in `master`. Python 3 should work now as python 2.7

PR #821, opened 2017-09-21
fix python3 import

polmr:
This seems to solve the issue: https://github.com/meganz/sdk/issues/552
However it will break the usage for users doing what's described in that issue.
Tested ok for python 2.7 (no changes in usage). Still it should be reviewed


polmr:
This caused `ImportError: No module named '_module'` in swig >= 3.0.9. Discarded in favour of: https://github.com/meganz/sdk/pull/832

PR #836, opened 2017-10-03
Added support to EventChangeToHttps on Objective-C bindings

javitrujillo:


javitrujillo:
trigger compilation

bindings/ios/Private/DelegateMEGAGlobalListener.mm
```diff
@@ -99,22 +99,15 @@
...
-                    [tempListener onEvent:tempMegaSDK event:(tempEvent ? [[MEGAEvent alloc] initWithMegaEvent:tempEvent cMemoryOwn:YES] : nil)];
-                });
-            }
-            break;
-            
-        default:
-            break;
+    if (listener != nil && [listener respondsToSelector:@selector(onEvent:event:)]) {
+        MegaEvent *tempEvent = NULL;
+        if(event) {
```
sergiohs84: No need to check it. `event` is always a valid object

PR #847, opened 2017-10-10
Return API_ENOENT instead of API_EKEY if the password can't be verified

javiserrano:


javiserrano:
Shouldn't we also call `localloout()` upon a failed log in?

PR #850, opened 2017-10-10
Import Welcome PDF upon account creation

sergiohs84:


examples/megacli.cpp
```diff
@@ -4583,6 +4608,8 @@ void DemoApp::ephemeral_result(handle uh, const byte* pw)
     Base64::btoa(pw, SymmCipher::KEYLENGTH, buf);
     cout << buf << endl;
 
+    pdf_to_import = true;
```
javiserrano: It seems that this flag is enabled inconditionally, so it would affect the usage of `resumeephemeral()`. Maybe it could be enabled just before the call `createephemeral()`. In that case, the other callback `ephemeral_result(error)` should be updated to clear the flag.

PR #857, opened 2017-10-18
Better management of local file updates using a temporary backup

javiserrano:
It allows the sync engine to create file versions in those cases

### DIFF
--- a/src/sync.cpp
+++ b/src/sync.cpp
@@ -868,22 +868,25 @@ LocalNode* Sync::checkpath(LocalNode* l, string* localpath, string* localname, d
                                         LOG_warn << "File checked in the future";
                                     }
 
-                                    if (currentsecs >= prevfa->mtime)
+                                    if (!waitforupdate)
                                     {
-                                        if (currentsecs - prevfa->mtime < (FILE_UPDATE_DELAY_DS / 10))
+                                        if (currentsecs >= prevfa->mtime)
                                         {
-                                            LOG_verbose << "currentsecs = " << currentsecs << "  mtime = " << prevfa->mtime;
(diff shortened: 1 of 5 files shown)


src/sync.cpp
```diff
@@ -813,6 +815,114 @@ LocalNode* Sync::checkpath(LocalNode* l, string* localpath, string* localname)
...
+                                    else
+                                    {
+                                        LOG_debug << "There isn't anything in the origin path";
+                                    }
+                                }
+
+                                if (waitforupdate)
+                                {
+                                    LOG_debug << "Possible file update detected.";
+                                    *backoffds += FILE_UPDATE_DELAY_DS;
```
sergiohs84: It should be `backoffds = FILE_UPDATE_DELAY_DS`.

src/sync.cpp
```diff
@@ -813,6 +815,114 @@ LocalNode* Sync::checkpath(LocalNode* l, string* localpath, string* localname)
...
+                                {
+                                    LOG_debug << "File detected in the origin of a move";
+
+                                    if (currentsecs >= client->updatedfilets)
+                                    {
+                                        if ((currentsecs - client->updatedfilets) < (FILE_UPDATE_DELAY_DS / 10))
+                                        {
+                                            LOG_verbose << "currentsecs = " << currentsecs << "  lastcheck = " << client->updatedfilets
+                                                      << "  currentsize = " << prevfa->size << "  lastsize = " << client->updatedfilesize;
+                                            LOG_debug << "The file was checked too recently. Waiting...";
```
polmr: this could be LOG_verbose

PR #863, opened 2017-10-25
Use a shared SymmCipher for all transfers

javiserrano:
to reduce the memory usage

### DIFF
--- a/src/megaclient.cpp
+++ b/src/megaclient.cpp
@@ -6111,10 +6111,15 @@ int MegaClient::readnodes(JSON* j, int notify, putsource_t source, NewNode* nn, 
                     {
                         app->reload("Node inconsistency");
 
-                        int creqtag = reqtag;
-                        reqtag = 0;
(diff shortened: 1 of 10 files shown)


src/megaclient.cpp
```diff
@@ -6108,10 +6111,15 @@ int MegaClient::readnodes(JSON* j, int notify, putsource_t source, NewNode* nn,
                     {
                         app->reload("Node inconsistency");
 
-                        int creqtag = reqtag;
-                        reqtag = 0;
-                        sendevent(99437, "Node inconsistency");
-                        reqtag = creqtag;
+                        static bool reloadnotified = false;
```
sergiohs84: Good catch! I guess we send event for every node detected as inconsistent before actually start a reload. But... after a reload, will we stop notifying future inconsistencies?

sergiohs84: Reloads are done after the app restarts, so there's no such issue :)

PR #866, opened 2017-10-31
Feature/backups

polmr:


src/gfx/qt.cpp
```diff
@@ -706,6 +706,22 @@ QImageReader *GfxProcQT::readbitmapFfmpeg(int &w, int &h, int &orientation, QStr
     //Allocate video frames
```
Satoshi2: @javiserrano

polmr:
trigger compilation

PR #872, opened 2017-11-10
Patch for RSA public key with padding

sergiohs84:
It prevents to create invalid `sigPubk` when RSA keypair is generated by
Webclient and the signature is created by the SDK.


sergiohs84:
trigger compilation

PR #878, opened 2017-11-15
Skip node copies over the the same node and version trees over others

javiserrano:


sergiohs84:
Everythings seems ok till this commit. Waiting for the push of the pending commits :)

PR #880, opened 2017-11-16
Changes for video-calls

sergiohs84:


polmr:
trigger compilation

src/megaapi_impl.cpp
```diff
@@ -10536,8 +10536,8 @@ void MegaApiImpl::notify_confirmation(const char *email)
 
 void MegaApiImpl::notify_disconnect()
 {
-    MegaEventPrivate *event = new MegaEventPrivate(MegaEvent::EVENT_DISCONNECT);
-    fireOnEvent(event);
+    //MegaEventPrivate *event = new MegaEventPrivate(MegaEvent::EVENT_DISCONNECT);
+    //fireOnEvent(event);
```
sergiohs84: Once @alber2510 changes MEGAchat to not disconnect on this event if there's an on-going call, I think this callback is suitable and should be kept here.

src/megaclient.cpp
```diff
@@ -9871,6 +9871,12 @@ void MegaClient::queueread(handle h, bool p, SymmCipher* key, int64_t ctriv, m_o
     else
     {
         it->second->enqueue(offset, count, reqtag, appdata);
+        if (overquotauntil && overquotauntil > Waiter::ds)
```
sergiohs84: @javiserrano are you aware of this change? it's not in `develop` yet, so I wonder how it ended up here in this branch

javiserrano: Yes, it's to fix the management of the overquota state on streaming transfers.
The problem was reported by Android developers and the fix has been already confirmed to be working fine.

PR #894, opened 2017-12-11
Creation of previews and thumbnails using ffmpeg

javiserrano:
Needs testing


src/gfx/qt.cpp
```diff
@@ -521,19 +574,293 @@ QImage GfxProcQT::resizebitmapQT(QImageReader *image, int orientation, int w, in
...
+        avformat_close_input(&formatContext);
+        return NULL;
+    }
+
+    // Find first video stream type
+    AVStream *videoStream = NULL;
+    int videoStreamIdx = 0;
+    for (int i = 0; i < formatContext->nb_streams; i++)
+    {
+        if (formatContext->streams[i]->codec->codec_type == AVMediaType::AVMEDIA_TYPE_VIDEO)
```
sergiohs84: I would add the checkup for `codec == NULL`

PR #915, opened 2018-01-12
Improvements for the completion of downloads

javiserrano:
- Restoration of missing fingerprints for non-sync transfers
- Do not modify fields of FileFIngerprint of the Transfer object
- Prevent persistent hangs in transfers caused by persistent errors accessing the filesystem
- Prevent failed transfers due to temporary errors during the completion of some files
- Restore missing file attributes once per file (instead of once per node)


javiserrano:
trigger compilation

PR #951, opened 2018-02-13
Update the bindings and SDK projects for WP and UWP to add onEvent callback

javiergm1983:


bindings/wp8/MEvent.cpp
```diff
@@ -0,0 +1,67 @@
...
+}
+
+MEvent^ MEvent::copy()
+{
+    return megaEvent ? ref new MEvent(megaEvent->copy(), true) : nullptr;
+}
+
+MEventType MEvent::getType()
+{
+    return (MEventType) (megaEvent ? megaEvent->getType() : 0);
```
sergiohs84: I'd use -1 by default, since 0 is a valid type of event. Anyway, the callback `onEvent(e)` always provide a valid object `e` (not like `onUserUpdate()` and similar, which are called with `NULL` upon fetchnodes completion)

javiergm1983: 👍 Ok, makes sense. Changed the default value to -1.

PR #965, opened 2018-03-05
These changes are to support buildling MEGAcmd for Synology and QNAP …

mattw-mega:
…NAS drives, but should also be useful for other environments.

Added new flags to build_sdk.sh to support cross-compiling: -I -R -X -C -O -0 (described in the help output)
Only building OpenSSL as static, added some lib options to curl to enable that, though there is now also a way to pass OpenSSL specific flags now if some builds still need that.
Addressed some issues with 32bit and old compilers (mostly in preparing FreeImage for building)
Echo the configure parameters for each project
Added a new constructor for MegaApi() which enables logging during construction - vital if there is a problem as otherwise the app exits on startup with no clue as to why.

### DIFF
--- a/contrib/build_sdk.sh
+++ b/contrib/build_sdk.sh
@@ -46,8 +46,16 @@ enable_cares=0
 enable_curl=0
 enable_libuv=0
 android_build=0
+readline_build=0
 enable_cryptopp=0
 disable_mediainfo=0
+incremental=0
+no_optimisation=0
+extra_openssl_params=""
+cross_compiling=0
+configure_cross_options=""
+openssl_cross_option=""
+status_dir=""
 
 on_exit_error() {
     echo "ERROR! Please check log files. Exiting.."
@@ -115,7 +123,6 @@ package_download() {
         return
     fi
 
-    echo "Downloading $name"
 
     if [ -f $file ]; then
         rm -f $file || true
@@ -123,14 +130,19 @@ package_download() {
 
     # use packages previously downloaded in /tmp/megasdkbuild folder
     # if not present download from URL specified
-    # if wget fail, try curl
+    # if wget fails, try curl
     mkdir -p /tmp/megasdkbuild/
-    
+
 #    cp /srv/dependencies_manually_downloaded/$3 $file 2>/dev/null || \
 
-    cp /tmp/megasdkbuild/$3 $file || \
-    wget --no-check-certificate -c $url -O $file --progress=bar:force -t 2 -T 30 || \
-    curl -k $url > $file || exit 1
+    if [ -e /tmp/megasdkbuild/$3 ] ; then
+        echo "Using cached file /tmp/megasdkbuild/$3"
+        cp /tmp/megasdkbuild/$3 $file 
+    else
+        echo "Downloading $name to get $3"
+        wget --no-check-certificate -c $url -O $file --progress=bar:force -t 2 -T 30 || \
+        curl -k $url > $file || exit 1
+    fi
     
     echo "Checking MD5SUM for $file"
     if ! echo $md5sum \*$file | md5sum -c - ; then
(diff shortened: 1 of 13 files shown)


contrib/build_sdk.sh
```diff
@@ -550,13 +636,20 @@ curl_pkg() {
...
+    fi
+
     package_download $name $curl_url $curl_file $curl_md5
     if [ $download_only -eq 1 ]; then
         return
     fi
 
     package_extract $name $curl_file $curl_dir
-    package_configure $name $curl_dir $install_dir "$curl_params"
+    package_configure $name $curl_dir $install_dir "$curl_params" "-ldl -lpthread"
```
polmr: I'm not 100% sure this has no consequences. does curl always need libpthread? how could it be working?

contrib/build_sdk.sh
```diff
@@ -656,6 +779,7 @@ freeimage_pkg() {
         # manually copy header and library
         cp $freeimage_dir/Dist/FreeImage.h $install_dir/include || exit 1
         cp $freeimage_dir/Dist/libfreeimage* $install_dir/lib || exit 1
+        echo $? > $status_dir/$name.success
```
polmr: this is actually equal to `echo 0 > $status_dir/$name.success`, but that's ok

contrib/build_sdk.sh
polmr: After testing some compilations: As a matter of fact, this breaks some cross compilation for Android in a system where libpthread is not available. If this is required for QNAP/SYNOLOGY compilation should be configured as parameter

src/filefingerprint.cpp
```diff
@@ -26,6 +26,12 @@
 #include "mega/utils.h"
 
 namespace mega {
+long long abs(long long n)
```
polmr: Good catch, but this might better go to utils.h & utils.cpp

src/transfer.cpp
```diff
@@ -30,6 +30,11 @@
 #include "megawaiter.h"
 
 namespace mega {
+static long long abs(long long n)
```
polmr: remove this one after moved to utils

src/megaapi_impl.cpp
```diff
@@ -3832,6 +3837,8 @@ void MegaApiImpl::init(MegaApi *api, const char *appKey, MegaGfxProcessor* proce
     syncLowerSizeLimit = 0;
     syncUpperSizeLimit = 0;
 
+    addLoggerClass(initLogger);
```
polmr: actually this does not make much sense. that function is static. You can simply use the static `MegaApi::addLoggerObject` before creating any MegaApi object. I've already updated `develop` branch in MEGAcmd project to proceed that way.

contrib/build_sdk.sh
mattw-mega: I ran into a problem linking curl at some point after removing "shared" from the OpenSSL static configuration, and found this solution (probably on stackoverflow).  However I'm not able to recreate the problem now, so perhaps this was addressed by other config fixes later on.  I'll remove these extra libs for this submission.

src/filefingerprint.cpp
mattw-mega: moved, thanks.  I was worried it would eclipse other overloads in the mega namespace, but it turns out this is the only overload of the function that we use.  If we need others later we can bring them into the namespace with 'using'

src/transfer.cpp
mattw-mega: done, thanks

src/megaapi_impl.cpp
mattw-mega: Agreed, I should have seen that the loggers were ultimately static.  I've undone those changes.  thanks!

contrib/build_sdk.sh
mattw-mega: cool, left as is

PR #1003, opened 2018-03-27
Add rich link user atrribute

alber2510:


src/megaapi_impl.cpp
```diff
@@ -5095,6 +5104,29 @@ void MegaApiImpl::setUserAttribute(int type, const MegaStringMap *value, MegaReq
...
+}
+
+void MegaApiImpl::setRichLinkWarningCounterValue(int value, MegaRequestListener *listener)
+{
+    MegaStringMap *stringMap = new MegaStringMapPrivate();
+    std::ostringstream oss;
+    oss << value;
+    string base64value;
+    Base64::btoa(oss.str(), base64value);
+    stringMap->set("c", base64value.c_str());
```
javiserrano: Since this atribute has several values ("c" and "num"), it seems that using this function would reset "num" and using the function above would reset "c". In that case, it would be needed to merge both functions to have only one that allows to correctly set both values of the attribute at once.

sergiohs84: @javiserrano in fact, the "c" should be set only when there's no value at "num". If there's "num" (no matters if `0` or `1`), the "c" should be ignored (and even overwritten). We can discuss if the effort of updating the existing value is worthy or not

PR #1018, opened 2018-04-13
Sync adjustments

javiserrano:


src/megaclient.cpp
```diff
@@ -10955,6 +10955,72 @@ bool MegaClient::syncup(LocalNode* l, dstime* nds)
...
                 if (ll->type == FILENODE)
                 {
+                    if (ll->node != rit->second)
+                    {
+                        ll->sync->statecacheadd(ll);
+                    }
+                    ll->setnode(rit->second);
+
+                    // check if file is likely to be identical
+                    if (*ll == *(FileFingerprint*)rit->second)
```
sergiohs84: May it happen that `!rit->second->isValid` at this point? It's a condition that used to be and now it's not ported to this check.
Additionally, there use to be a check for the size that has not been ported either:
```if (ll->size == rit->second->size)```

File: src/megaapi_impl.cpp [hunk 2]
```diff
@@ -16985,10 +16982,17 @@ void MegaApiImpl::update()
 int MegaApiImpl::isWaiting()
 {
 #ifdef ENABLE_SYNC
-    if (waiting || client->syncfslockretry)
+    if (client->syncfslockretry || client->syncfsopsfailed)
+    {
+        LOG_debug << "SDK waiting for a blocked file: " << client->blockedfile;
```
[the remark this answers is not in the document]
javiserrano: The SDK will keep retrying on both cases. `syncfslockretry` means that there are files that can't be scanned at the moment, while `syncfsopsfailed` means that there are external changes that can't be applied to the local filesystem at the moment. Both should be temporary problems, usually because the affected file is blocked by another app.

src/megaclient.cpp
javiserrano: Yes, it could happen. The value of `isValid` is taken into account in the overloaded operator `==` of `FileFingerprint`, resulting on a check similar to the one that has been replaced, but taking also into account that the modification times can be slightly different due to its resolution on some filesystems like FAT.

PR #1028, opened 2018-04-20
Set flags correctly for rich link attribute request

alber2510:


src/megaapi_impl.cpp
```diff
@@ -11482,22 +11482,31 @@ void MegaApiImpl::getua_result(error e)
...
+                request->setFlag(true); // the password reminder dialog should be shown
+            }
         }
-        else if (request->getType() == MegaRequest::TYPE_GET_ATTR_USER
-                 && (time(NULL) - client->accountsince) > User::PWD_SHOW_AFTER_ACCOUNT_AGE)
+        else if (request->getParamType() == MegaApi::USER_ATTR_RICH_PREVIEWS &&
+                 request->getType() == MegaRequest::TYPE_GET_ATTR_USER)
         {
-            request->setFlag(true); // the password reminder dialog should be shown
+            request->setNumber(0);
```
sergiohs84: This line is not needed at all :)

PR #1051, opened 2018-05-03
Support for chat-links

sergiohs84:


examples/megacli.cpp
```diff
@@ -771,11 +841,7 @@ void DemoApp::printChatInformation(TextChat *chat)
     }
     if (!chat->title.empty())
     {
-        char *tstr = new char[chat->title.size() * 4 / 3 + 4];
-        Base64::btoa((const byte *)chat->title.data(), int(chat->title.size()), tstr);
-
-        cout << "\tTitle: " << tstr << endl;
-        delete [] tstr;
+        cout << "\tTitle: " << chat->title.c_str() << endl;
```
javiserrano: The old code managed `chat->title` as a binary string.

examples/megacli.cpp
```diff
@@ -4730,6 +4876,150 @@ static void process_line(char* l)
 
                         }
                     }
+                    else if (words[0] == "chatlu")
+                    {
+                        if (words.size() == 2)
+                        {
+                            handle publichandle;
```
javiserrano: If `handle` is not initialized to zero, `Base64::atob`, depending on the last parameter, it could not set all bytes. There are more cases of this possible problem above and below.

include/mega/command.h
```diff
@@ -273,10 +273,11 @@ class MEGA_API CommandPutUA : public Command
 class MEGA_API CommandGetUA : public Command
 {
     string uid;
-    attr_t at;  // attribute type
+    attr_t at;      // attribute type
+    const char *ph; // public handle for preview mode
```
javiserrano: Saving the `const char *` directly would require to take the ownership of the buffer received in the constructor or at least to know and manage the lifetime of that buffer. I think that it's much better to copy the value to a `std::string` like we do in other commands.

include/mega/types.h
```diff
@@ -227,6 +227,13 @@ typedef enum { TYPE_UNKNOWN = -1, FILENODE = 0, FOLDERNODE, ROOTNODE, INCOMINGNO
 const int FILENODEKEYLENGTH = 32;
 const int FOLDERNODEKEYLENGTH = 16;
 
+
+//handle lengths
+static const int PUBLICHANDLE = 6;
+
+//handle lengths
+static const int USERHANDLE = 8;
+
```
javiserrano: It would be probably better to use the already existing `MegaClient::USERHANDLE` and `MegaClient::NODEHANDLE` or, if `PUBLICHANDLE` is not related to nodes, add the new value there. Anyway, it seems that these definitions are not being used.

include/megaapi.h
```diff
@@ -12118,9 +12195,46 @@ class MegaApi
...
          * @param group Flag to indicate if the chat is a group chat or not
          * @param peers MegaTextChatPeerList including other users and their privilege level
+         * @param title Byte array that contains the chat topic if exists. NULL if no custom title is required.
          * @param listener MegaRequestListener to track this request
          */
-        void createChat(bool group, MegaTextChatPeerList *peers, MegaRequestListener *listener = NULL);
+        void createChat(bool group, MegaTextChatPeerList *peers, const char *title = NULL, MegaRequestListener *listener = NULL);
+
+        /**
+         * @brief Creates an public chatroom for multiple participants (groupchat)
```
javiserrano: an -> a

include/megaapi.h
```diff
@@ -12151,9 +12266,11 @@ class MegaApi
          * - MegaTextChatPeerList::PRIV_MODERATOR = 3
          * @param title Byte array representing the title that wants to be set, already encrypted and
          * converted to Base64url encoding (optional).
+         * @param unifiedKey Byte array that contains the unified key, already encrypted and
+         * converted to Base64url encoding.
          * @param listener MegaRequestListener to track this request
          */
-        void inviteToChat(MegaHandle chatid, MegaHandle uh, int privilege, const char *title = NULL, MegaRequestListener *listener = NULL);
+        void inviteToChat(MegaHandle chatid, MegaHandle uh, int privilege, const char *unifiedKey = NULL, const char *title = NULL, MegaRequestListener *listener = NULL);
```
javiserrano: I'm not sure but, if `unifiedKey` is only related to public chats, it could be better to create a new function like `inviteToPublicChat` with this parameter mandatory and keeping `inviteToChat` in the previous state.

examples/megacli.cpp
sergiohs84: I've checked and the change is correct. The `TextChat::title` string stores the B64url encoded version of the corresponding array of bytes that represents the `ct`

examples/megacli.cpp
sergiohs84: Nice catch. Thanks for the notice. It's also interesting for @jgandres. I've already fixed it.

src/file.cpp
```diff
@@ -217,13 +229,36 @@ File *File::unserialize(string *d)
...
+        if (ptr + chatauthlen + sizeof(unsigned short) > end)
+        {
+            LOG_err << "File unserialization failed - chat auth too long";
+            delete fp;
+            return NULL;
+        }
+        const char *chatauth = ptr;
+        ptr += chatauthlen;
+
+        file->chatauth = strdup(chatauth);
```
javiserrano: I think that there are several problems in serialization / unserialization of the chat auth.
- `MemAccess::get<unsigned short>(ptr);` could read after the end of the buffer. I think that it's needed to add a check before that line.
- The check after that line contains `+ sizeof(unsigned short)`, and it doesn't seem correct because only `chatauthlen` bytes are going to be read after that.
- The usage of `strdup` is not correct there because, during the serialization, the trailing NULL character of the string is not stored (only `strlen` bytes). So it's needed to save that character or take advantage of `chatauthlen` to know the length of the string. Also, it's preferred to not use `strdup` because that memory should be freed using `free()` and we use `delete []` to free all those strings.

src/megaapi_impl.cpp
```diff
@@ -595,20 +608,37 @@ MegaNodePrivate *MegaNodePrivate::unserialize(string *d)
...
+    {
+        unsigned short chatauthlen = MemAccess::get<unsigned short>(ptr);
+        ptr += sizeof(chatauthlen);
+        if (ptr + chatauthlen + sizeof(unsigned short) > end)
+        {
+            LOG_err << "MegaNodePrivate unserialization failed - chat auth too long";
+            return NULL;
+        }
+        chatauth = ptr;
+        ptr += chatauthlen;
```
javiserrano: The same comments in `File::unserialize` can be applied here.

src/megaapi_impl.cpp
```diff
@@ -9395,7 +9539,24 @@ void MegaApiImpl::authorizeMegaNodePrivate(MegaNodePrivate *node)
...
+}
+
+MegaNode *MegaApiImpl::authorizeChatNode(MegaNode *node, const char *cauth)
+{
+    if (!node)
+    {
+        return NULL;
+    }
+
+    sdkMutex.lock();
```
javiserrano: I think that it's not needed to lock/unlock the mutex in this function because it doesn't access anything from the SDK

src/file.cpp
sergiohs84: You are very correct. I added checkups to not read after the end of the buffer, fixed the checkup with the wrong condition `(ptr + chatauthlen + sizeof(unsigned short))` and replaced the `strdup` by a simple `memcpy` using the `chatauthlen` (allocating memory with `new` previously).

sergiohs84: You are very correct. I added checkups to not read after the end of the buffer, fixed the checkup with the wrong condition `(ptr + chatauthlen + sizeof(unsigned short))` and replaced the `strdup` by a simple `memcpy` using the `chatauthlen` (allocating memory with `new` previously).

sergiohs84: You are very correct. I added checkups to not read after the end of the buffer, fixed the checkup with the wrong condition `(ptr + chatauthlen + sizeof(unsigned short))` and replaced the `strdup` by a simple `memcpy` using the `chatauthlen` (allocating memory with `new` previously).

sergiohs84: You are very correct. I added checkups to not read after the end of the buffer, fixed the checkup with the wrong condition `(ptr + chatauthlen + sizeof(unsigned short))` and replaced the `strdup` by a simple `memcpy` using the `chatauthlen` (allocating memory with `new` previously).

src/megaapi_impl.cpp
sergiohs84: Agree. Removing the lock.

PR #1053, opened 2018-05-07
New registration and login procedure

javiserrano:


polmr:
trigger compilation

polmr:
trigger compilation

src/commands.cpp
```diff
@@ -1429,8 +1429,55 @@ void CommandLogout::procresult()
...
+}
+
+void CommandPrelogin::procresult()
+{
+    if (client->json.isnumeric())
+    {
+        return client->app->prelogin_result(0, NULL, NULL, (error)client->json.getint());
+    }
+
+    int v = 1;
```
sergiohs84: Maybe we can default it to `0`. Then, in the `EOO`, we can really check whether a `v=1` has been received. Otherwise, return error.

PR #1064, opened 2018-05-15
Support to use libraw in the gfx layer with QT

javiserrano:


src/gfx.cpp
```diff
@@ -90,6 +90,155 @@ const char* GfxProc::supportedvideoformats()
...
+        delete job;
+    }
+
+    while (job = responses.pop())
+    {
+        for (int i = 0; i < job->imagetypes.size(); i++)
+        {
+            delete job->images[i];
+        }
+        delete job;
```
polmr: couldn't be this deletion be called before the response is processed in GfxProc::checkevents?

src/gfx/qt.cpp
```diff
@@ -147,10 +151,12 @@ int GfxProcQT::getExifOrientation(QString &filePath)
             default:
                 // Skip any other sections.
                 delete data;
+                data = NULL;
                 break;
         }
     }
 
+    delete data;
```
polmr: Shouldn't we move this `delete data` to the end of `case M_EXIF:`?

src/gfx/qt.cpp
```diff
@@ -586,10 +614,175 @@ const char *GfxProcQT::supportedformatsQT()
...
+    libraw_processed_image_t *output = NULL;
+    const libraw_data_t &imgdata = libRaw.imgdata;
+    LOG_debug << "Processing RAW image: " << imagePath.toUtf8().constData()
+              << " " << imgdata.sizes.width
+              << " " << imgdata.sizes.height
+              << " " << imgdata.thumbnail.twidth
+              << " " << imgdata.thumbnail.theight
+              << " " << imgdata.sizes.flip;
+
+    if (imgdata.thumbnail.twidth > 0 && imgdata.thumbnail.theight > 0)
```
polmr: Should we consider discarding libraw's preview image if it is too small?

src/gfx/qt.cpp
```diff
@@ -586,10 +614,175 @@ const char *GfxProcQT::supportedformatsQT()
...
+        int pixelSize = output->colors * colorSize;
+
+        unscaled = QImage(output->width, output->height, QImage::Format_RGB32);
+        uchar *pixels = unscaled.bits();
+        uchar *data = output->data;
+        for (int i = 0; i < numPixels; i++, data += pixelSize)
+        {
+            if (output->colors == 3)
+            {
+                pixels[i * 4] = data[2 * colorSize];
```
polmr: minor performance enhancement (in case the compiler don't do it automatically): save i*4

src/gfx/qt.cpp
javiserrano: In that case, there would be a double `delete` because that case finishes on the `default` case. Anyway, I will improve the code to make it more readable.

src/gfx/qt.cpp
javiserrano: I did it in my first approach, but I changed it to follow the same approach as the webclient with these images (always use the thumbnail if available).

src/gfx/qt.cpp
javiserrano: I will change it 👍

PR #1097, opened 2018-06-18
Nas merge

mattw-mega:


contrib/build_sdk.sh
```diff
@@ -477,14 +477,15 @@ libuv_pkg() {
 
     package_extract $name $libuv_file $libuv_dir
 
+    export OLD_CFLAGS="$CFLAGS"
```
polmr: why export?

mattw-mega: thanks - I've created a much cleaner version of this PR and included the fix for this.

PR #1107, opened 2018-06-26
Update WP and UWP bindings

javiergm1983:
- Update `MRequest` types enumeration.
- Allow to get the last error from `MTransfer` objects.
- Update documentation of the `contactLinkQuery` request.
- Restoration of nodes moved to the Rubbish Bin.
- New function to know if an outgoing shared folder is pending.


polmr:
trigger compilation

PR #1121, opened 2018-07-12
double-checked locking for CurlHttpIO::locking_function

mattw-mega:


src/posix/net.cpp
```diff
@@ -35,13 +35,19 @@ MUTEX_CLASS CurlHttpIO::curlMutex(false);
 #if defined(USE_OPENSSL) && !defined(OPENSSL_IS_BORINGSSL)
 
 MUTEX_CLASS **CurlHttpIO::sslMutexes = NULL;
+static MUTEX_CLASS lock_init_mutex;
```
javiserrano: The constructor needs a parameter to initialize the mutex. Otherwise, most implementations of `MUTEX_CLASS` expect a call to `init()` and crash in the first `lock()` if it's not called. The only exception is the implementation for Win32.
Changing this line by `static MUTEX_CLASS lock_init_mutex(false);` (or `true` but the mutex doesn't need to be recursive in this case) the fix should work for all platforms.

src/posix/net.cpp
```diff
@@ -35,13 +35,19 @@ MUTEX_CLASS CurlHttpIO::curlMutex(false);
...
 void CurlHttpIO::locking_function(int mode, int lockNumber, const char *, int)
 {
     MUTEX_CLASS *mutex = sslMutexes[lockNumber];
     if (mutex == NULL)
     {
-        mutex = new MUTEX_CLASS(true);
-        sslMutexes[lockNumber] = mutex;
+        // we still have to be careful about multiple threads getting to this point simultaneously
+        lock_init_mutex.lock();
+        while (!(mutex = sslMutexes[lockNumber]))
```
javiserrano: I understand that the double check is needed because another thread could have created the mutex a thread passes the `if` above but is there any reason to have a `while` here instead of an `if`?

src/posix/net.cpp
mattw-mega: thanks for catching that - I've added the parameter, cheers

src/posix/net.cpp
mattw-mega: the `while` was just to get that `mutex = sslMutexes[lockNumber]` executed a second time after the mutex was constructed and added to the array, so that we can use the value in the code below as usual.  But you're right, it's clearer for the reader to use an `if`, especially as we can easily do the assignment in the existing line.  Adjusted - cheers

javiserrano: Great, thank you!

PR #1123, opened 2018-07-17
Support for the new PSA System

javiserrano:


include/mega/types.h
```diff
@@ -440,7 +440,8 @@ typedef enum {
     ATTR_PWD_REMINDER = 15,     // private, non-encrypted - char array in B64 - non-versioned
     ATTR_DISABLE_VERSIONS = 16, // private, non-encrypted - char array in B64 - non-versioned
     ATTR_CONTACT_LINK_VERIFICATION = 17,  // private, non-encrypted - char array in B64 - non-versioned
-    ATTR_RICH_PREVIEWS = 18      // private - byte array
+    ATTR_RICH_PREVIEWS = 18,     // private - byte array
+    ATTR_LAST_PSA = 19      // private - char array
```
sergiohs84: We will need to adjust types for attributes. The 19 is also used in the https://github.com/meganz/sdk/pull/1190 for the RubbishTime.

src/megaapi_impl.cpp
```diff
@@ -16729,6 +16776,16 @@ void MegaApiImpl::sendPendingRequests()
 
                     client->putua(type, (byte *)value, 1);
                 }
+                else if (type == ATTR_LAST_PSA)
+                {
+                    if (!value || !atoll(value))
```
sergiohs84: If the PSA's id 0 were a valid one, it could block the app to set that PSA as seen. Not sure if ids start from 1?

PR #1139, opened 2018-08-02
auto-test app for sync functionality

mattw-mega:
Added a test app for sync.  It's not finished yet but has some useful tests and may be good to have available in other branches.  I've used some c++17 features just in that (without affecting the main SDK - mostly std::filesystem) which prompted this next one:

Resolved some issues when building with a c++17 compliant compiler, mainly that it defines std::byte which clashes with the byte we were defining, and probably with others.  The main change is removing 'using namespace std' from headers (even inside the mega namespace it still caused clashes).  Just 'using std::map' to pull the specific items we need into the mega namespace works pretty well.


mattw-mega:
I see cryptopp have made an adjustment for std::byte also, they too have stopped relying on a global definition and defined their own in their own namespace.   I see that in version 610 which I've been building against.  From their config.h: ```// Originally in global namespace to avoid ambiguity with other byte typedefs.
// Moved to Crypto++ namespace due to C++17, std::byte and potential compile problems. Also see
// http://www.cryptopp.com/wiki/std::byte and http://github.com/weidai11/cryptopp/issues/442
// typedef unsigned char byte;
#define CRYPTOPP_NO_GLOBAL_BYTE 1

NAMESPACE_BEGIN(CryptoPP)

typedef unsigned char byte;```

PR #1170, opened 2018-08-29
check for empty arrays in cr_response

mattw-mega:
also send an event if we detect a request for inshare keys


sergiohs84:
trigger compilation

PR #1187, opened 2018-09-07
Hotfix: Call notifypurge() before a possible commit to the local database.

javiserrano:
Force a reload to fix possible inconsistencies caused by this bug.


polmr:
trigger compilation

polmr:
trigger compilation

polmr:
trigger compilation

PR #1199, opened 2018-09-20
Support for MEGAcmd to not use `readline` on WIN32

mattw-mega:
Instead it can use our unicode-friendly WinConsole
Added MutexGuard, a very simple RAII way to lock one of our mutexes on
construction, and unlock automatically on exit from the scope
Added an option to specify a function to fetch additional completion
options for a command line, and merge them in our autocomplete code.


polmr:
trigger compilation

PR #1221, opened 2018-10-08
Always stream the last generated link

bubbajr:


javiserrano:
Despite it's a valid patch to always stream the last generated link, it would break other usages of the HTTP server.

PR #1246, opened 2018-10-31
MEGAchat has clashing enums #defined in winhttp.h

mattw-mega:
avoid #including winhttp.h until it's really needed as it #defines some symbols that are then used in an enum declaration in MEGAchat.


mattw-mega:
trigger compilation

javiserrano:
trigger compilation

PR #1256, opened 2018-11-07
Moved autocomplete.h/cpp from win32 to main folders

mattw-mega:
Also enabled building for linux, provided the compiler supports c++11 or later.
This enables autocomplete use for megaclc in the MEGAchat project.


mattw-mega:
trigger compilation

src/autocomplete.cpp
```diff
@@ -503,7 +517,7 @@ bool LocalFS::addCompletions(ACState& s)
         fs::path searchPath(s.word().s + (s.word().s.empty() || (s.word().s.back() == '\\'  || s.word().s.back() == '/' ) ? "*" : ""));
         char sep = (!s.word().s.empty() && s.word().s.find('/') != string::npos ) ?'/':'\\';
         bool relative = !searchPath.is_absolute();
-        searchPath = relative ? fs::current_path().append("\\").append(searchPath) : searchPath;
+        searchPath = relative ? fs::current_path().append("\\").append(searchPath.u8string()) : searchPath;
         std::string cp = relative ? fs::current_path().u8string() + "\\" : "";
```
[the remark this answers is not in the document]
mattw-mega: I think I only added it where one or another compiler insisted on it, perhaps the older experimental version of fs::path (which may differ a little between windows/linux libraries) is not completely consistent with the append() function.   Thanks very much for the review!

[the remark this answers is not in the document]
mattw-mega: I looked a bit further, and the path append() function really just converts the argument(s) to a path, and uses operator /= to do the actual joining.   So using that directly is shorter and more efficient, since we already have a path object to append.   I've made the change, and also enabled its use in megacli so that we can test it on linux (megaclc doesn't do anything with paths) (and provided the c++ compiler supports c++11, which mega::autocomplete needs).   cheers!

examples/megacli.cpp
```diff
@@ -6684,12 +6689,59 @@ void DemoApp::userattr_update(User* u, int priv, const char* n)
...
 
+#ifndef NO_READLINE
+char* longestCommonPrefix(ac::CompletionState& acs)
+{
+    string s = acs.completions[0].s;
+    for (int i = acs.completions.size(); i--; )
+    {
+        for (unsigned j = 0; j < acs.completions[i].s.size(); ++j)
+        {
+            if (j < s.size() && s[j] != acs.completions[i].s[j])
```
polmr: Minor improvement (untested):
```
        for (unsigned j = 0; j < acs.completions[i].s.size() && j < s.size(); ++j)
        {
            if (s[j] != acs.completions[i].s[j])
```

mattw-mega: Excellent point, I've made that improvement, and tested it on linux and windows.   I noticed one other thing while testing as well, when trying to iterate a non-existent directory, and that is fixed also.  thanks!

examples/megacli.cpp
```diff
@@ -22,6 +22,7 @@
 #include "mega.h"
 #include "megacli.h"
 #include <fstream>
+#include <mega/autocomplete.h>
```
polmr: this breaks linux compilation in QT if configured to compile with c++11 (autocomplete.cpp is not added)

mattw-mega: Good point.  I have added autocomplete.cpp to the various project files in the SDK, in case there are any other similar cases.  It's good to have it there for consistency, and will only be compiled if the compiler is configured for C++11 or later.

mattw-mega:
trigger compilation

PR #1290, opened 2018-11-28
Backups Improvements

polmr:
- do not store backup transfers in cache
- improved docstrings
- return list of failed transfers
- onBackupFinish return INCOMPLETE/EXPIRED(skipped)/ENOENT(not found) now
- added TransferTemporaryError support for backups
- fix logic that hanged the backups after certain circumstances (pendingTags badly managed)


javiserrano:
I have unified the different `startUpload` functions that were in this pull request because they had the same mandatory parameters and that could cause compilation errors due to ambiguous function calls. 

Additionally, I have made all parameters of the more complex `startUpload` function mandatory to make easier to detect incorrect usages.

Please @polmr, take a look at my changes when possible to check if there is anything wrong. In my opinion, the pull request is ready 👍

PR #1292, opened 2018-12-02
Fix for excess CPU relating to curl timeouts

mattw-mega:
The main changes are making sure the NEEDEXEC flag (and any others occuring) are set on timeout,
and calling curl_multi_action_socket for timeouts.
Also addressed 32/64 bit warnings in those files.


sergiohs84:
These commits have been merged as result of merging PR https://github.com/meganz/sdk/pull/1370, which also include them.
Closing PR and deleting branch

PR #1298, opened 2018-12-11
New overstorage logic

javiserrano:


src/commands.cpp
```diff
@@ -3634,6 +3604,24 @@ void CommandGetUserQuota::procresult()
                 break;
 
             case EOO:
+                if (got_storage)
+                {
+                    if (details->storage_used > details->storage_max)
```
polmr: >= ??

javiserrano: Updated to use the most conservative options to prevent possible problems

src/commands.cpp
```diff
@@ -3633,7 +3604,37 @@ void CommandGetUserQuota::procresult()
...
+            case EOO:                
+                if (uslw <= 0)
+                {
+                    uslw = 9000;
+                    LOG_warn << "Using default almost overstorage threshold";
+                }
+
+                if (got_storage)
+                {
+
```
sergiohs84:
```suggestion
```

PR #1321, opened 2019-01-08
Clear peerlist upon own user's leave

sergiohs84:
When the own user is removed/leaves a room, the API still includes the
peerlist in the actionpacket. However, the peerlist is not included in a
fresh fetchnodes. This commit aims to be consistent with API's state.
Also, it clears the peerlist if inactive chats are stored in cache.


sergiohs84:
This branch has been replaced by the one in PR https://github.com/meganz/sdk/pull/1324

PR #1335, opened 2019-01-16
Recent Actions feature.

mattw-mega:
This feature is being added to the SDK to match the feature in the webclient, available from the clock on the left hand bar.
You will be able to view changes to your MEGA account, as groups of files that were recently uploaded or recently updated.
A different user, folder or significant time difference would mean the files go into a different bucket.


include/mega/megaclient.h
```diff
@@ -1069,6 +1069,21 @@ class MEGA_API MegaClient
...
+    struct recentaction
+    {
+        m_time_t time;
+        handle user;
+        handle parent;
+        bool updated;
+        bool media;
+        node_vector v;
+    };
+    typedef vector<recentaction> recentactions_vector;
```
sergiohs84: Definition of types usually go to the `types.h` file

include/megaapi_impl.h
```diff
@@ -1547,6 +1548,44 @@ class MegaUserAlertListPrivate : public MegaUserAlertList
...
+    MegaRecentActionBucketPrivate(MegaClient::recentaction& ra, MegaClient* mc);
+    MegaRecentActionBucketPrivate(int64_t timestamp, const string& user, handle parent, bool update, bool media, MegaNodeList*);
+    virtual ~MegaRecentActionBucketPrivate();
+    virtual MegaRecentActionBucket *copy() const;
+    virtual int64_t getTimestamp() const;
+    virtual const char* getUserEmail() const;
+    virtual MegaHandle getParentHandle() const;
+    virtual bool getIsUpdate() const;
+    virtual bool getIsMedia() const;
+    MegaNodeList* getNodes();
```
sergiohs84: It misses the virtual, and could be defined as `const`, both the method and the returned pointer :)

include/megaapi_impl.h
```diff
@@ -1547,6 +1548,44 @@ class MegaUserAlertListPrivate : public MegaUserAlertList
...
+
+private:
+    int64_t timestamp;
+    string user;
+    handle parent;
+    bool update, media;
+    MegaNodeList* list;
```
sergiohs84: A meaningful name is always a plus. Better to name it `nodes` here

examples/megacli.cpp
```diff
@@ -5983,6 +5997,22 @@ static void process_line(char* l)
 
                         return;
                     }
+                    else if (words[0] == "recentnodes")
```
sergiohs84: New options should be added to the `help` menu, with their params

examples/megacli.cpp
```diff
@@ -5983,6 +5997,22 @@ static void process_line(char* l)
...
+                        {
+                            node_vector nv = client->getRecentNodes(atoi(words[2].c_str()), m_time() - 60 * 60 * atoi(words[1].c_str()), false);
+                            for (unsigned i = 0; i < nv.size(); ++i)
+                            {
+                                cout << nv[i]->displaypath() << endl;
+                            }
+                        }
+                        else
+                        {
+                            cout << "syntax error" << endl;
```
sergiohs84: As you can see in any other command, in case of syntax error we show the correct syntax. It's very useful.

include/mega/megaclient.h
```diff
@@ -1069,6 +1069,21 @@ class MEGA_API MegaClient
...
+    node_vector getRecentNodes(unsigned maxcount, m_time_t since, bool includerubbishbin);
+
+    struct recentaction
+    {
+        m_time_t time;
+        handle user;
+        handle parent;
+        bool updated;
+        bool media;
+        node_vector v;
```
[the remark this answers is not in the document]
mattw-mega: yes, good point

src/megaapi.cpp
```diff
@@ -200,6 +200,68 @@ int MegaUserAlertList::size() const
...
+}
+
+const char* MegaRecentActionBucket::getUserEmail() const
+{
+    return NULL;
+}
+
+MegaHandle MegaRecentActionBucket::getParentHandle() const
+{
+    return UNDEF;
```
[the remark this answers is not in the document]
mattw-mega: yes, the one from the intermediate layer is more appropriate here, thanks

include/megaapi.h
```diff
@@ -2399,6 +2399,118 @@ class MegaUserAlertList
...
+    * You are the owner of the returned object
+    *
+    * @return Copy of the MegaRecentActionBucket object
+    */
+    virtual MegaRecentActionBucket *copy() const;
+
+    /**
+    * @brief Returns a timestamp reflecting when these changes occurred
+    *
+    * @return Timestamp indicating when the changes occurred
```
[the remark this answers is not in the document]
mattw-mega: yes, good point

include/megaapi.h
```diff
@@ -2399,6 +2399,118 @@ class MegaUserAlertList
...
+    * @brief Returns whether the files are photos or videos
+    *
+    * @return true if the files in this change are media files.
+    */
+    virtual bool getIsMedia() const;
+
+    /**
+    * @brief Returns nodes representing the files chanvged in this bucket
+    *
+    * @return A list of the files in the bucket.  The bucket retains ownership.
```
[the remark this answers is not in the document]
mattw-mega: good point, thanks

include/megaapi.h
```diff
@@ -11611,6 +11723,21 @@ class MegaApi
...
+         * The function that takes no parameters uses the defaults for the MEGA apps
+         * which are (currently) within the last 30 days, and max 10000 nodes.
+         *
+         * @param since      Only return nodes that are more recent than this time.
+         * @param maxnodes   Only return nodes up to this many.
+         *
+         * @return List of buckets containing nodes that were added/modifed as a set
+         */
+        MegaRecentActionBucketList* getRecentActions(int64_t since, unsigned maxnodes);
+        MegaRecentActionBucketList* getRecentActions();
```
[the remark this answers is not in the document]
mattw-mega: thanks, I wasn't aware of that

include/mega/megaclient.h
mattw-mega: thanks, I wasn't aware of that convention, I'll follow it in future.  cheers

include/megaapi_impl.h
mattw-mega: excellent point :-)

include/megaapi_impl.h
mattw-mega: cool

examples/megacli.cpp
mattw-mega: Yes, I agree - will PM you about this

examples/megacli.cpp
mattw-mega: Yes, I agree - will PM you about this

include/megaapi_impl.h
mattw-mega: in the update I've added the override keyword, which is even better since any change in the parent class that causes the function to not override anymore will cause the compiler to alert us to the problem.  cheers!

include/megaapi.h
```diff
@@ -11611,6 +11723,21 @@ class MegaApi
...
+         * Each bucket contains files that were added/modified in a set, by a single user.
+         * The function that takes no parameters uses the defaults for the MEGA apps
+         * which are (currently) within the last 30 days, and max 10000 nodes.
+         *
+         * @param since      Only return nodes that are more recent than this time.
+         * @param maxnodes   Only return nodes up to this many.
+         *
+         * @return List of buckets containing nodes that were added/modifed as a set
+         */
+        MegaRecentActionBucketList* getRecentActions(int64_t since, unsigned maxnodes);
```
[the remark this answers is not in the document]
mattw-mega: cool

include/megaapi_impl.h
sergiohs84: We should start using `overrided` now that C++11 is a must. I'll tell the team ;)

src/megaclient.cpp
```diff
@@ -12971,6 +12972,177 @@ node_vector *MegaClient::nodesbyfingerprint(FileFingerprint* fingerprint)
...
+{
+    MUTEX_CLASS media_check(false);   // when we can use c++11, switch to a lambda that remembers the MegaClient* for compare().  In the meantime, force just one MegaClient at a time instead.
+    MegaClient* mc;
+
+    const static string webclient_is_image_def = ".jpg.jpeg.gif.bmp.png.";
+    const static string webclient_is_image_raw = ".3fr.arw.cr2.crw.ciff.cs1.dcr.dng.erf.iiq.k25.kdc.mef.mos.mrw.nef.nrw.orf.pef.raf.raw.rw2.rwl.sr2.srf.srw.x3f.";
+    const static string webclient_is_image_thumb = "psd.svg.tif.tiff.webp";  // leaving out .pdf
+    const static string webclient_mime_photo_extensions = ".3ds.bmp.btif.cgm.cmx.djv.djvu.dwg.dxf.fbs.fh.fh4.fh5.fh7.fhc.fpx.fst.g3.gif.heic.heif.ico.ief.jpe.jpeg.jpg.ktx.mdi.mmr.npx.pbm.pct.pcx.pgm.pic.png.pnm.ppm.psd.ras.rgb.rlc.sgi.sid.svg.svgz.tga.tif.tiff.uvg.uvi.uvvg.uvvi.wbmp.wdp.webp.xbm.xif.xpm.xwd.";
+    const static string webclient_mime_video_extensions = ".3g2.3gp.asf.asx.avi.dvb.f4v.fli.flv.fvt.h261.h263.h264.jpgm.jpgv.jpm.m1v.m2v.m4u.m4v.mj2.mjp2.mk3d.mks.mkv.mng.mov.movie.mp4.mp4v.mpe.mpeg.mpg.mpg4.mxu.ogv.pyv.qt.smv.uvh.uvm.uvp.uvs.uvu.uvv.uvvh.uvvm.uvvp.uvvs.uvvu.uvvv.viv.vob.webm.wm.wmv.wmx.wvx.";
```
sergiohs84: All the above seems to be the supported formats from Webclient. I don't know whether the SDK supports the same set than web, it should, but I think it's not the case.
Should we use the formats supported by the graphics processor / media-info from the SDK? like the ones here:
https://github.com/meganz/sdk/blob/26c7448f57d2b8d6c85a4c4a985ed814f56b0e2d/src/mediafileattribute.cpp#L561
or here:
https://github.com/meganz/sdk/blob/82080a5c43bc5a1bb6a093f0cbbad2d632aef782/src/gfx/freeimage.cpp#L383

sergiohs84: Alright, just read:
> // evaluate according to the webclient rules, so that we get exactly the same bucketing.

Disregard my previous comment :)

src/megaclient.cpp
```diff
@@ -12971,6 +12972,177 @@ node_vector *MegaClient::nodesbyfingerprint(FileFingerprint* fingerprint)
...
+        string localname, name = n->displayname();
+        mc->fsaccess->path2local(&name, &localname);
+        char ext[10];
+        if (mc->fsaccess->getextension(&localname, ext, sizeof(ext)))
+        {
+            strcat(ext, ".");
+
+            if (webclient_is_image_def.find(ext) != string::npos ||
+                webclient_is_image_raw.find(ext) != string::npos ||
+                (webclient_mime_photo_extensions.find(ext) != string::npos && n->hasfileattribute(GfxProc::PREVIEW)) ||
```
sergiohs84: Why do we skip photos without a preview?

src/megaclient.cpp
```diff
@@ -12971,6 +12972,177 @@ node_vector *MegaClient::nodesbyfingerprint(FileFingerprint* fingerprint)
...
+        {
+            v.push_back(i->second);
+        }
+    }
+
+    // heaps use a 'less' function, and pop_heap returns the largest item stored.
+    std::make_heap(v.begin(), v.end(), nodes_ctime_less);
+
+    node_vector v2;
+    v2.reserve(maxcount);
```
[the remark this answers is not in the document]
mattw-mega: sounds good

src/megaclient.cpp
```diff
@@ -12971,6 +12972,177 @@ node_vector *MegaClient::nodesbyfingerprint(FileFingerprint* fingerprint)
...
+
+        return false;
+    }
+
+    static bool comparetime(const recentaction& a, const recentaction& b)
+    {
+        return a.time > b.time;
+    }
+
+};
```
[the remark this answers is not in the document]
mattw-mega: good point, I'll try to remember to leave those out for namespaces

src/megaclient.cpp
```diff
@@ -12971,6 +12972,177 @@ node_vector *MegaClient::nodesbyfingerprint(FileFingerprint* fingerprint)
...
+        {
+            ++bucketend;
+        }
+
+        std::sort(i, bucketend, action_bucket_compare::compare);
+
+        for (node_vector::iterator j = i; j != bucketend; ++j)
+        {
+            string a = (*i)->displaypath();
+            string b = (*j)->displaypath();
```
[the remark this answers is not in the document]
mattw-mega: yes, good spotting

src/megaclient.cpp
mattw-mega: Yes, I started out using the SDK facillities, but the requirement is to match the webclient (within reason of course), and there were too many discrepancies reported.   Also there may be native clients that don't want to build with those libraries themselves but still be able to show recent account changes.

src/megaclient.cpp
mattw-mega: The reason that's in there is that it's imitating the logic in the webclient - if you search that codebase for `is_image3` you'll see it makes the same test.  Probably due to some case where a common enough file extension is something other than a photo, maybe some program's proprietary format.   The most common photo extensions are covered by the first two tests anyway, so we will mostly not depend on that.  cheers

include/megaapi_impl.h
mattw-mega: 👍

bindings/ios/MEGASdk.mm
```diff
@@ -1786,6 +1789,30 @@ - (MEGANodeList *)nodeListSearchForNode:(MEGANode *)node searchString:(NSString
     return [[MEGANodeList alloc] initWithNodeList:self.megaApi->search((node != nil) ? [node getCPtr] : NULL, (searchString != nil) ? [searchString UTF8String] : NULL, YES) cMemoryOwn:YES];
 }
 
+- (NSMutableArray *)recentActions {
+    MegaRecentActionBucketList *megaRecentActionBucketList = self.megaApi->getRecentActions();
+    NSUInteger count = (NSUInteger)megaRecentActionBucketList->size();
+    NSMutableArray *recentActionBucketMutableArray = NSMutableArray.alloc.init;
```
jnavarrom: It would be better use `initWithCapacity` because we know the size of the bucket list (`count`)

bindings/ios/MEGASdk.mm
```diff
@@ -1786,6 +1789,30 @@ - (MEGANodeList *)nodeListSearchForNode:(MEGANode *)node searchString:(NSString
...
+        [recentActionBucketMutableArray addObject:recentActionBucket];
+    }
+    
+    return recentActionBucketMutableArray;
+}
+
+- (NSMutableArray *)recentActionsSinceDate:(NSDate *)date maxNodes:(NSUInteger)maxNodes {
+    MegaRecentActionBucketList *megaRecentActionBucketList = self.megaApi->getRecentActions((int64_t)date.timeIntervalSince1970, (unsigned)maxNodes);
+    NSUInteger count = (NSUInteger)megaRecentActionBucketList->size();
+    NSMutableArray *recentActionBucketMutableArray = NSMutableArray.alloc.init;
```
jnavarrom: Same can be applied here

bindings/ios/MEGASdk.mm
javisanesp: As discussed we can save some casting by using an int for the variable `count`, and reuse it on the loop.

javisanesp: And only do the casting on the initWithCapacity 🙂

mattw-mega: sounds like a good optimisation, please do make that change.   I think we can merge first though as it will work correctly as is.  cheers

PR #1338, opened 2019-01-17
consider current backup as completed in removal before attr is set

polmr:


polmr:
Should fix this: https://github.com/meganz/MEGAcmd/issues/158

PR #1339, opened 2019-01-17
#12277 Background photo uploads (CUv2)

mattw-mega:
Enable use of the iOS background uploads feature.  iOS will do our uploading for us in the background, in a low power fashion, without the app running.   And iOS aggressively shuts the app down now, so files queued in the SDK don't get much time to be uploaded before iOS decides to stop the app to save power.   So we have added commands to:
	- Perform pre-encryption of files in function MegaBackgroundMediaUpload::encryptFile, which supports encrypting the whole file or pieces of it, without using too much RAM and with options to minimise disk use.  Details in its comments.
	- Get the URL for iOS to upload to.
	- Finish an upload with node creation once all the data has been uploaded, and the upload token received.


mattw-mega:
trigger compilation

src/user.cpp
```diff
@@ -416,6 +416,10 @@ string User::attr2string(attr_t type)
             attrname = "*!geo";
             break;
 
+        case ATTR_UNSHAREABLE_ATTR:
+            attrname = "*unshareable";
```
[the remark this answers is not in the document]
mattw-mega: Yes, excellent point.  I've made that update and pushed it to the branch, thanks

bindings/ios/MEGASdk.h
```diff
@@ -5495,6 +5589,19 @@ typedef NS_ENUM(NSUInteger, StorageState) {
  */
 - (MEGANode *)nodeForFingerprint:(NSString *)fingerprint parent:(MEGANode *)parent;
 
+/**
+ * @brief Returns nodes that have an originalFingerprint equal to the supplied value
+ *
+ * Search the node tree and return a list of nodes that have an originalFingerprint, which
+ * matches the supplied originalfingerprint.
+ *
+ * You take the ownership of the returned value.
```
jnavarrom: I think we can remove this line

bindings/ios/MEGABackgroundMediaUpload.h
```diff
@@ -0,0 +1,70 @@
+/**
+ * @file MEGABackgroundMediaUpload.h
+ * @brief Time zone details
```
jnavarrom: I think this is not the brief of this class.

bindings/ios/MEGABackgroundMediaUpload.h
```diff
@@ -0,0 +1,70 @@
...
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface MEGABackgroundMediaUpload : NSObject
+
+/**
+ * @brief Encrypt the file or a portion of it
+ *
+ * Call this function once with the file to be uploaded.  It uses mediainfo to extract information that will
```
jnavarrom: Extra spaces after .

bindings/ios/MEGABackgroundMediaUpload.h
```diff
@@ -0,0 +1,70 @@
...
+ * help the webclient show or play the file in various browsers.  The information is stored in this object
+ * until the whole operation completes.   The encrypted data is stored in a new file.
```
jnavarrom: Extra spaces after .

bindings/ios/MEGABackgroundMediaUpload.h
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * In order to save space on mobile devices, this function can be called in such a way that the last portion
+ * of the file is encrypted (to a new file), and then that last portion of the file is removed by file truncation.
+ * That operation can be repeated until the file is completely encrypted, and only the encrypted version remains,
+ * and takes up the same amount of space on the device.   The size of the portions must first be calculated by using
```
jnavarrom: Extra spaces after .

bindings/ios/MEGABackgroundMediaUpload.h
```diff
@@ -0,0 +1,70 @@
...
+ * so that RAM usage is not excessive.
+ *
+ * @param inputFilePath The file to encrypt a portion of (and the one that is ultimately being uploaded).
+ * @param start The index of the first byte of the file to encrypt
+ * @param length The number of bytes of the file to encrypt.  The function will round this value up by up to 1MB to fit the
+ *        MEGA internal chunking algorithm.  The number of bytes acutally encrypted and stored in the new file is the updated number.
+ * @param outputFilePath The name of the new file to create, and store the encrypted data in.
+ * @param urlSuffix The function will update the string passed in.  The content of the string must be appended to the URL
+ *        when this portion is uploaded.
+ * @param adjustsSizeOnly If this is set true, then encryption is not performed, and only the length parameter is adjusted.
```
jnavarrom: Usually we used YES/NO instead true/false

bindings/ios/MEGABackgroundMediaUpload.h
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * @brief Retrieves the value of the uploadURL once it has been successfully requested via requestBackgroundUploadURLWithFileSize
+ */
+- (NSString *)uploadURLString;
+
+/**
+ * @brief Turns the data stored in this object into a binary data.
+ *
+ * The object can then be recreated via MegaApi::backgroundMediaUploadResume and supplying the same binary.
```
jnavarrom: Maybe with the [MEGASdk resumeBackgroundMediaUploadBySerializedData:] ?

bindings/ios/MEGABackgroundMediaUpload.mm
```diff
@@ -0,0 +1,71 @@
+/**
+ * @file MEGABackgroundMediaUpload.mm
+ * @brief Time zone details
```
jnavarrom: I think this is not the brief of this class.

bindings/ios/MEGABackgroundMediaUpload.mm
```diff
@@ -0,0 +1,71 @@
...
+    delete _mediaUpload;
+}
+
+- (mega::MegaBackgroundMediaUpload *)getCPtr {
+    return self.mediaUpload;
+}
+
+- (BOOL)encryptFileAtPath:(NSString *)inputFilePath startPosition:(int64_t)start length:(unsigned *)length outputFilePath:(NSString *)outputFilePath urlSuffix:(NSString **)urlSuffix adjustsSizeOnly:(BOOL)adjustsSizeOnly {
+    std::string suffix;
+    BOOL succeed = self.mediaUpload->encryptFile(inputFilePath.UTF8String, start, length, outputFilePath.UTF8String, &suffix, adjustsSizeOnly);
```
jnavarrom: Usually we pass to the intermediate layer an UTF8 string of NULL for const char * arguments, depending if the NSString is `nil` or not. I am not sure if we can/must do the same here, it depends how it is implemented in the c++ intermediate layer.

bindings/ios/MEGASdk.h
```diff
@@ -2585,7 +2585,31 @@ typedef NS_ENUM(NSUInteger, StorageState) {
...
+ * To remove the existing coordinates, set both the latitude and longitude to nil.
+ *
+ * The 'unshareable' variant of this function stores the coordinates with an extra
+ * layer of encryption which only this user can decrypt, so that even if this node is shared
+ * with others, they cannot read the coordinates.
+ *
+ * The associated request type with this request is MEGARequestTypeSetAttrNode
+ * Valid data in the MEGARequest object received on callbacks:
+ * - [MEGARequest nodeHandle] - Returns the handle of the node that receive the attribute
+ * - [MEGARequest flag] - Returns true (official attribute)
```
jnavarrom: YES

bindings/ios/MEGASdk.h
```diff
@@ -5019,6 +5043,81 @@ typedef NS_ENUM(NSUInteger, StorageState) {
...
 
+/**
+ * @brief Initial step to upload a photo/video via iOS low-power background upload feature
+ *
+ * Call ensureMediaInfo() first in order prepare the library to attach file attributes
+ * that enable videos to be identified and played in the web browser.
+ *
+ * @return A pointer to an object that keeps some needed state through the process of
+ *         uploading a media file via iOS low power background uploads (or similar).
+ *         Caller takes ownership of the object.
```
jnavarrom: I think we can remove this line, because we don't need free the MEGABackgroundMediaUpload object in the application code layer.

bindings/ios/MEGASdk.h
```diff
@@ -5019,6 +5043,81 @@ typedef NS_ENUM(NSUInteger, StorageState) {
...
+ */
+- (MEGABackgroundMediaUpload *)backgroundMediaUpload;
+
+/**
+ * @brief Get back the needed MEGABackgroundMediaUpload after the iOS app exited and restarted
+ *
+ * In case the iOS app exits while a background upload is going on, and the app is started again
+ * to complete the operation.  Call this version to recreate the MegaBackgroundMediaUpload object
```
jnavarrom: Extra spaces after .
MEGABackgroundMediaUpload

bindings/ios/MEGASdk.h
```diff
@@ -5019,6 +5043,81 @@ typedef NS_ENUM(NSUInteger, StorageState) {
...
+ * needed for a call to backgroundMediaUploadComplete.   The object must have been serialised
```
jnavarrom: Extra spaces

bindings/ios/MEGASdk.h
```diff
@@ -5019,6 +5043,81 @@ typedef NS_ENUM(NSUInteger, StorageState) {
...
+ * before the app was unloaded.
+ *
+ * @return A pointer to an object that keeps some needed state through the process of
+ *         uploading a media file via iOS low power background uploads (or similar).
+ *         Caller takes ownership of the object.
```
jnavarrom: I think we can remove this line.

bindings/ios/MEGASdk.h
```diff
@@ -5019,6 +5043,81 @@ typedef NS_ENUM(NSUInteger, StorageState) {
...
+ */
+- (MEGABackgroundMediaUpload *)resumeBackgroundMediaUploadBySerializedData:(NSData *)data;
+
+/**
+ * @brief Request the URL suitable for uploading a media file.
+ *
+ * This function requests the URL needed for uploading the file.  The URL will need the urlSuffix
```
jnavarrom: Extra spaces

bindings/ios/MEGASdk.h
```diff
@@ -5019,6 +5043,81 @@ typedef NS_ENUM(NSUInteger, StorageState) {
...
+ * from the MEGABackgroundMediaUpload::encryptFile to be appended before actually sending.
```
jnavarrom: Usually we use the name of the Obj-c binding instead the name of the c++ intermediate layer, for methods, enums, etc
[MEGABackgroundMediaUpload encryptFileAtPath:startPosition:length:outputFilePath:urlSuffix:adjustsSizeOnly:]

bindings/ios/MEGASdk.h
```diff
@@ -5019,6 +5043,81 @@ typedef NS_ENUM(NSUInteger, StorageState) {
...
+ * The result of the request is signalled by the listener onRequestFinsish callback with TYPE_GET_BACKGROUND_UPLOAD_URL.
```
jnavarrom: MEGARequestTypeGetBackgroundUploadURL

bindings/ios/MEGASdk.h
```diff
@@ -5019,6 +5043,81 @@ typedef NS_ENUM(NSUInteger, StorageState) {
...
+ * Provided the error code is API_OK, the URL is available in the MegaBackgroundMediaUpload via its getUploadURL().
```
jnavarrom: MEGAErrorTypeApiOk
MEGABackgroundMediaUpload

bindings/ios/MEGASdk.h
```diff
@@ -5019,6 +5043,81 @@ typedef NS_ENUM(NSUInteger, StorageState) {
...
+ *
+ * Call this function just once (per file) to find out the URL to upload to, and upload all the pieces to the same
+ * URL.   If errors are encountered and the operation must be restarted from scratch, then a new URL should be requested.
```
jnavarrom: Extra spaces

bindings/ios/MEGASdk.h
```diff
@@ -5019,6 +5043,81 @@ typedef NS_ENUM(NSUInteger, StorageState) {
...
+ * A new URL could specify a different upload server for example.
+ *
+ * @param filesize The size of the file
+ * @param state A pointer to the MegaBackgroundMediaUpload object tracking this upload
```
jnavarrom: MEGABackgroundMediaUpload

bindings/ios/MEGASdk.h
```diff
@@ -5019,6 +5043,81 @@ typedef NS_ENUM(NSUInteger, StorageState) {
...
+ */
+- (void)requestBackgroundUploadURLWithFileSize:(int64_t)filesize mediaUpload:(MEGABackgroundMediaUpload *)upload delegate:(id<MEGARequestDelegate>)delegate;
+
+/**
+ * @brief Create the node after completing the background upload of the file.
+ *
+ * Call this function after completing the background upload of all the file data
+ * The node representing the file will be created in the cloud, with all the suitable
+ * attributes and file attributes attached.
+ * The result of the request is signalled by the listener MEGARequestDelegate callback with TYPE_COMPLETE_BACKGROUND_UPLOAD.
```
jnavarrom: MEGARequestTypeCompleteBackgroundUpload

bindings/ios/MEGASdk.h
```diff
@@ -5019,6 +5043,81 @@ typedef NS_ENUM(NSUInteger, StorageState) {
...
+ *
+ * @param mediaUpload The MEGABackgroundMediaUpload object tracking this upload
+ * @param fileName The leaf name of the file, utf-8 encoded
+ * @param parentNode The folder node under which this new file should appear
+ * @param fingerprint  The fingerprint for the uploaded file (use MegaApi::getFingerprint to generate this)
```
jnavarrom: Extra space after the param fingerprint
[MEGASdk fingerprintForFilePath:]

bindings/ios/MEGASdk.h
```diff
@@ -5019,6 +5043,81 @@ typedef NS_ENUM(NSUInteger, StorageState) {
...
+ * @param originalFingerprint If the file uploaded is modified from the original,
+ *        pass the fingerprint of the original file here, otherwise NULL.
```
jnavarrom: nil instead NULL, is more Objective-C friendly.

bindings/ios/MEGASdk.h
```diff
@@ -5019,6 +5043,81 @@ typedef NS_ENUM(NSUInteger, StorageState) {
...
+- (BOOL)completeBackgroundMediaUpload:(MEGABackgroundMediaUpload *)mediaUpload fileName:(NSString *)fileName parentNode:(MEGANode *)parentNode fingerprint:(NSString *)fingerprint originalFingerprint:(NSString *)originalFingerprint binaryUploadToken:(NSData *)token delegate:(id<MEGARequestDelegate>)delegate;
+
+/**
+ * @brief Call this to enable the library to attach media info attributes
+ *
+ * Those attributes enable the web browser to know a file is a video, and
+ * play it with the correct codec.
+ *
+ * @return YES if the library is ready, otherwise the request for media translation
+ *         data is sent to MEGA.  In that case, call again later or use a listener
```
jnavarrom: Extra spaces

bindings/ios/MEGASdk.h
```diff
@@ -5435,32 +5534,27 @@ typedef NS_ENUM(NSUInteger, StorageState) {
...
  *
- * @param assetRepresentation ALAssetRepresentation that provides the data to create the fingerprint
+ * @param data NSData that provides the data to create the fingerprint
  * @param modificationTime Modification time that will be taken into account for the creation of the fingerprint
  * @return Base64-encoded fingerprint
  */
-- (NSString *)fingerprintForAssetRepresentation:(ALAssetRepresentation *)assetRepresentation modificationTime:(NSDate *)modificationTime;
-
-#pragma clang diagnostic pop
+- (NSString *)fingerprintForData:(NSData *)data modificationTime:(NSDate *)modificationTime;
```
jnavarrom: We had `ALAssetRepresentation` because maybe someone wants to use our SDK to develop an application and support iOS 7. But it is not a problem at all if we remove this.

bindings/ios/MEGABackgroundMediaUpload.h
simonwng: changed to "Background media upload". good catch, thanks

bindings/ios/MEGABackgroundMediaUpload.h
simonwng: fixed

bindings/ios/MEGABackgroundMediaUpload.h
simonwng: fixed

bindings/ios/MEGABackgroundMediaUpload.h
simonwng: fixed

bindings/ios/MEGABackgroundMediaUpload.h
simonwng: sure, fixed

bindings/ios/MEGABackgroundMediaUpload.h
simonwng: this is copied from megaapi.h, I have notified @mattw-mega to remove the extra spaces as well. Thanks for catching this.

bindings/ios/MEGABackgroundMediaUpload.h
simonwng: Good catch. I have changed it to 

> resumeBackgroundMediaUploadBySerializedData in MEGASdk

MEGASdk is a class, here I think it is more appropriate to introduce the method rather than giving a real coding syntax.

bindings/ios/MEGABackgroundMediaUpload.mm
simonwng: It was copied across. fixed, thanks!

bindings/ios/MEGABackgroundMediaUpload.mm
simonwng: I think technically, nil and NULL are same and equal to 0. And usually where you can use NULL you can pass nil. Here is the define of nil in MacTypes.h

> #ifndef nil
  #if defined(__has_feature) 
    #if __has_feature(cxx_nullptr)
      #define nil nullptr
    #else
      #define nil __DARWIN_NULL
    #endif
  #else
    #define nil __DARWIN_NULL
  #endif
#endif

However, for consistent and make the code easier to understand, I am happy to convert nil to NULL explicitly

bindings/ios/MEGASdk.h
simonwng: Yes. I have replaced all the true/false to YES/NO in `MEGAsdk.h`

bindings/ios/MEGASdk.h
simonwng: fixed

bindings/ios/MEGASdk.h
simonwng: removed

bindings/ios/MEGASdk.h
simonwng: fixed

bindings/ios/MEGASdk.h
simonwng: removed

bindings/ios/MEGASdk.h
simonwng: fixed

bindings/ios/MEGASdk.h
simonwng: fixed

bindings/ios/MEGASdk.h
simonwng: fixed

bindings/ios/MEGASdk.h
simonwng: removed

bindings/ios/MEGASdk.h
simonwng: fixed

bindings/ios/MEGASdk.h
simonwng: fixed

bindings/ios/MEGASdk.h
simonwng: I removed it is because the `ALAssetRepresentation` is already deprecated in iOS quite a time ago.

bindings/ios/MEGASdk.h
simonwng: true, removed

bindings/ios/MEGABackgroundMediaUpload.h
jnavarrom: It is ok, but the Objective-C method instead the c++ method

bindings/ios/MEGABackgroundMediaUpload.mm
simonwng: Yes, they are same. Thanks Navarro!

polmr:
trigger compilation

polmr:
trigger compilation

polmr:
trigger compilation

1488777:
> Aktivera användningen av iOS-bakgrundsuppladdningsfunktionen. iOS kommer att göra vår uppladdning för oss i bakgrunden, på ett lågt sätt, utan att appen körs. Och iOS stänger aggressivt appen nu, så filer i kö i SDK får inte mycket tid att ladda upp innan IOS beslutar att stoppa appen för att spara ström. Så har vi lagt till kommandon för att: 
> - Utför förkryptering av filer i funktionen MegaBackgroundMediaUpload :: encryptFile, som stöder krypterar hela filen eller delar av den, utan att använda för mycket RAM och med alternativ för att minimera diskanvändning. Detaljer i sina kommentarer. 
> - Hämta URL för iOS att ladda upp till. 
> - Avsluta en uppladdning med node skapelse när alla data har laddats upp och uppladdningstoken mottogs.

Duplicate of #

include/megaapi.h
```diff
@@ -477,6 +478,21 @@ class MegaNode
          */
         virtual const char* getFingerprint();
 
+        /**
+         * @brief Returns the original fingerprint (Base64-encoded) of the node
+         *
+         * In the case where a file was modified before uploaded (eg. resized photo or gps coords removed),
+         * it may have an original fingerprint set (by setOriginalFingerprint() or backgroundMediaUploadComplete()),
```
sergiohs84: In MegaApi's documentation, methods are written without the parenthesis usually. Something like `MegaApi::setOriginalFingerprint`. I'll fix it

include/megaapi.h
```diff
@@ -5874,6 +5914,77 @@ class MegaListener
...
+     * @param startPos The index of the first byte of the file to encrypt
+     * @param length The number of bytes of the file to encrypt.  The function will round this value up by up to 1MB to fit the
+     *        MEGA internal chunking algorithm.  The number of bytes acutally encrypted and stored in the new file is the updated number.
+     * @param outputFilepath The name of the new file to create, and store the encrypted data in.
+     * @param urlSuffix The function will update the string passed in.  The content of the string must be appended to the URL
+     *        when this portion is uploaded.
+     * @param adjustsizeonly If this is set true, then encryption is not performed, and only the length parameter is adjusted.
+     *        This feature is to enable precalculating the exact sizes of the file portions for upload.
+     */
+    virtual bool encryptFile(const char* inputFilepath, int64_t startPos, unsigned int* length, const char* outputFilepath, std::string* urlSuffix, bool adjustsizeonly);
```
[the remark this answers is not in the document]
mattw-mega: The urlSuffix is an output parameter, and the pattern in the SDK seems to be to use std::string* in those cases?  Whereas the `const char*` parameters are input parameters.

include/megaapi.h
```diff
@@ -12529,6 +12684,103 @@ class MegaApi
...
+         *         Caller takes ownership of the object.
+         */
+        MegaBackgroundMediaUpload* backgroundMediaUploadResume(const std::string* serialised);
+
+        /**
+         * @brief Request the URL suitable for uploading a media file.
+         *
+         * This function requests the URL needed for uploading the file.  The URL will need the urlSuffix
+         * from the MegaBackgroundMediaUpload::encryptFile to be appended before actually sending.
+         * The result of the request is signalled by the listener onRequestFinsish callback with TYPE_GET_BACKGROUND_UPLOAD_URL.
```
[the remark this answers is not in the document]
mattw-mega: thanks, updated

include/megaapi.h
```diff
@@ -12529,6 +12684,103 @@ class MegaApi
...
+         * In case the iOS app exits while a background upload is going on, and the app is started again
+         * to complete the operation.  Call this version to recreate the MegaBackgroundMediaUpload object
+         * needed for a call to backgroundMediaUploadComplete.   The object must have been serialised
+         * before the app was unloaded.
+         *
+         * @return A pointer to an object that keeps some needed state through the process of
+         *         uploading a media file via iOS low power background uploads (or similar).
+         *         Caller takes ownership of the object.
+         */
+        MegaBackgroundMediaUpload* backgroundMediaUploadResume(const std::string* serialised);
```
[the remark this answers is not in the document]
mattw-mega: I started to make this change, and discovered why I did it that way - MegaBackgroundMediaUpload needs to store a MegaApiImpl* for future use, and immediate use.   That pointer is only available via the MegaApi.   It's not the first class to store one of those pointers, but it's the first to do that and support serialization.   What's the best way to resolve that?   thanks!

src/commands.cpp
```diff
@@ -393,6 +393,57 @@ void CommandPutFile::procresult()
     }
 }
 
+// request upload target URL for application to upload photo to using eg. iOS background upload feature
+CommandPutFileBackgroundURL::CommandPutFileBackgroundURL(m_off_t size, int putmbpscap, int ctag)
+{
+    cmd("u");
+    arg("ssl", 2);   // always SSL for background uploads
```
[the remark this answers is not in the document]
mattw-mega: This was decided early on with Simon, Mathias, and Bram.  I'll send you the details - perhaps we should revisit it.

include/megaapi.h
```diff
@@ -12529,6 +12684,103 @@ class MegaApi
...
+         * attributes and file attributes attached.
+         * The result of the request is signalled by the listener onRequestFinsish callback with TYPE_COMPLETE_BACKGROUND_UPLOAD.
+         *
+         * @param state The MegaBackgroundMediaUpload object tracking this upload
+         * @param utf8Name The leaf name of the file, utf-8 encoded
+         * @param parent The folder node under which this new file should appear
+         * @param fingerprint  The fingerprint for the uploaded file (use MegaApi::getFingerprint to generate this)
+         * @param fingerprintoriginal If the file uploaded is modified from the original,
+         *        pass the fingerprint of the original file here, otherwise NULL.
+         * @param binaryUploadToken The N binary bytes of the token returned from the file upload (of the last portion). N=36 currently.
```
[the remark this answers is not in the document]
mattw-mega: About this one, I was concerned that it would be difficult to specify the encoding to use, and felt that straight binary would be the least difficult to get the client app to supply correctly.   We don't provide any binary-to-64 functions for the client to use (that I can see?), and I presume that is deliberate?  thanks

src/commands.cpp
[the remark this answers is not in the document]
sergiohs84: It seems Apple push apps to use SSL for all transfers since 2016, so let's keep it on.

include/megaapi.h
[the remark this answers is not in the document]
sergiohs84: Well, there are methods to convert handles from/to B64, but it's true that till now there was no need to provide converters for any size. Maybe it's time to do it?

include/megaapi.h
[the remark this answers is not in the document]
sergiohs84: Correct me if I'm wrong, but the only method in `MegaApi` that uses a pointer to `std::string` is a converter UTF8<-->UTF16 used by WP8. It's only doable in Windows, it will not work in Android.
If we want to provide this feature only for iOS, it should be fine. If you want to make it compatible with other languages (specially those not C-friendly like Java or others), I recommend to change it. Otherwise, just mark this conversation as _Resolved_ :)

include/megaapi.h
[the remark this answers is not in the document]
sergiohs84: The best way to solve it would be to add the pointer to `MegaApi` to the params, to both of them:
```
static MegaBackgroundMediaUpload* MegaBackgroundMediaUpload::unserialize(const char* data, MegaApi* api);
```
```
static MegaBackgroundMediaUpload* MegaBackgroundMediaUpload::createInstance(MegaApi *api)
```
You can see the example above for the `createInstance()`, for which I provided a possible implementation it my comment.

include/megaapi.h
```diff
@@ -5874,6 +5914,77 @@ class MegaListener
...
+     * so that RAM usage is not excessive.
+     *
+     * @param inputFilepath The file to encrypt a portion of (and the one that is ultimately being uploaded).
+     * @param startPos The index of the first byte of the file to encrypt
+     * @param length The number of bytes of the file to encrypt.  The function will round this value up by up to 1MB to fit the
+     *        MEGA internal chunking algorithm.  The number of bytes acutally encrypted and stored in the new file is the updated number.
+     * @param outputFilepath The name of the new file to create, and store the encrypted data in.
+     * @param urlSuffix The function will update the string passed in.  The content of the string must be appended to the URL
+     *        when this portion is uploaded.
+     * @param adjustsizeonly If this is set true, then encryption is not performed, and only the length parameter is adjusted.
```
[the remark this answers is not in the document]
sergiohs84: Done!

src/megaapi_impl.cpp
```diff
@@ -20316,7 +20330,7 @@ void MegaApiImpl::sendPendingRequests()
             }
             std::string uploadTokenStr(uploadToken);
             std::string binaryUploadToken;
-            Base64::atob(binaryUploadToken, uploadTokenStr);
+            Base64::atob(uploadTokenStr, binaryUploadToken);
```
sergiohs84: Ouch, my fault! Thanks for spotting it :+1:

src/megaapi_impl.cpp
```diff
@@ -882,6 +930,194 @@ MegaHandle MegaNodePrivate::getOwner() const
...
+    w.serializeexpansionflags(false, false);
+    return s;
+}
+
+
+MegaBackgroundMediaUploadPrivate::~MegaBackgroundMediaUploadPrivate()
+{
+}
+
+class MEGA_API EncryptFilePieceByChunks : public EncryptByChunks
```
sergiohs84: It's usually a good practice to keep declarations in header files and definitions in source files. Why is this one all together here? I'll fix it

src/megaapi_impl.cpp
```diff
@@ -882,6 +930,194 @@ MegaHandle MegaNodePrivate::getOwner() const
...
+    // specialisation for encrypting a piece of a file without using too much RAM
+    FileAccess* fain;
+    FileAccess* faout;
+    m_off_t inpos, outpos;
+    string buffer;
+    unsigned lastsize;
+
+public:
+
+    EncryptFilePieceByChunks(FileAccess* f1, m_off_t f1pos, FileAccess* f2, m_off_t f2pos, SymmCipher* k, chunkmac_map* m, uint64_t c)
```
sergiohs84: Despite there's a lot of legacy code not following this rules, we try to name variable members with a prefix `m`, and `c` for params in the ctor, specially in MEGAchat and in the new objects of the intermediate layer in SDK. I'll fix it

src/megaclient.cpp
```diff
@@ -13190,6 +13192,43 @@ recentactions_vector MegaClient::getRecentActions(unsigned maxcount, m_time_t si
     return rav;
 }
 
+
+void MegaClient::nodesbyoriginalfingerprint(const char* originalfingerprint, Node* parent, node_vector *nv)
+{
+    if (parent)
```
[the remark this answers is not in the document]
sergiohs84: Fixed at `MegaApiImpl::getNodesByOriginalFingerprint()`

include/megaapi.h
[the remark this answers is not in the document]
mattw-mega: Good point, I had not realised that about the java bindings.   I've made the update, those return a string now instead, with an empty string indicating failure.  thanks!

include/megaapi.h
```diff
@@ -12529,6 +12684,103 @@ class MegaApi
...
+         * @brief Call this to enable the library to attach media info attributes
+         *
+         * Those attributes enable the web browser to know a file is a video, and
+         * play it with the correct codec.
+         *
+         * @return True if the library is ready, otherwise the request for media translation
+         *         data is sent to MEGA.  In that case, call again later or use a listener
+         *         to find out when the translation data is available.
+         */
+        bool ensureMediaInfo();
```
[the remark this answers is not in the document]
mattw-mega: Good spotting on the procresult().  That code was from the transition from no version to having a version at the front, and tried to cope with either from the API.  Now we can be sure a version is expected first, I'll update it.
About isMediaInfoReady, if we arrange it like that we could end up with two independent parts of a client app both requesting the `mc` data, and actually fetch it twice even though we only need one copy.   Should we be concerned about that?  thanks

src/commands.cpp
```diff
@@ -3064,6 +3132,96 @@ void CommandDelUA::procresult()
...
 
+
+CommandUnshareableUA::CommandUnshareableUA(MegaClient* client, bool fetch, int triesleft)
+{
+    maxtries = triesleft;
+    fetching = fetch;
+    if (fetching)
+    {
+        cmd("uga");
+        arg("u", client->finduser(client->me, 1)->uid.c_str());
```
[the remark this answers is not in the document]
mattw-mega: good point, I had not realised `finduser()` had that side effect, thanks for that.  I've switched it to use `client->uid` instead, which I added later on in this project.  It is set at the same time `client->me` is.

src/commands.cpp
```diff
@@ -3064,6 +3132,96 @@ void CommandDelUA::procresult()
...
+    else
+    {
+        byte newunshareablekey[SymmCipher::BLOCKSIZE];
+        client->rng.genblock(newunshareablekey, sizeof(newunshareablekey));
+
+        cmd("up");
+        arg(User::attr2string(ATTR_UNSHAREABLE_ATTR).c_str(), newunshareablekey, sizeof(newunshareablekey));
+        notself(client);
+    }
+    tag = 0;
```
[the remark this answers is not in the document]
mattw-mega: so it is, nice one.  fixed

src/commands.cpp
```diff
@@ -393,6 +393,60 @@ void CommandPutFile::procresult()
...
+
+// set up file transfer with returned target URL
+void CommandPutFileBackgroundURL::procresult()
+{
+    string url;
+
+    if (client->json.isnumeric())
+    {
+        error e = (error)client->json.getint();
+        if (!canceled)
```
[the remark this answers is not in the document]
mattw-mega: I see `Command` does have an implementation for `cancel()`, and it does set the flag?   Perhaps that was changed fairly recently?

src/megaapi_impl.cpp
```diff
@@ -882,6 +930,194 @@ MegaHandle MegaNodePrivate::getOwner() const
...
+    api->client->rng.genblock(filekey, sizeof filekey);
+}
+
+MegaBackgroundMediaUploadPrivate::MegaBackgroundMediaUploadPrivate(const string& serialised, MegaApiImpl* capi)
+    : api(capi)
+{
+    CacheableReader r(serialised);
+    string mediapropertiesstr;
+    unsigned char expansions[8];
+    string fileattrstring; // fileattrstring is not serialized
```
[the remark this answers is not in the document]
mattw-mega: I know Simon is keen for that - initially we had too much else to get working and using the existing functions for attaching those to an existing node kept it a bit simpler, with less to go wrong.   I think we could do it by providing a version of those functions like setThumbnail but without taking a Node, and just do the metad uploading (using the existing framework) and return a file attribute to the listener.  Those file attributes could then be supplied to the backgroundMediaUploadComplete() for the "atomic" completion that Simon is looking to achieve- where the concern is that once the user sees the node appear in the browser, they may close or delete the iOS app, without all attributes or file attributes being attached yet.

src/megaapi_impl.cpp
```diff
@@ -882,6 +930,194 @@ MegaHandle MegaNodePrivate::getOwner() const
...
+
+std::string MegaBackgroundMediaUploadPrivate::serialize()
+{
+    std::string s;
+    CacheableWriter w(s);
+    w.serializebinary(filekey, sizeof(filekey));
+    w.serializechunkmacs(chunkmacs);
+    w.serializestring(mediaproperties.serialize());
+    w.serializestring(url);
+    w.serializeexpansionflags(false, false);
```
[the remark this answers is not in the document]
mattw-mega: MegaNodePrivate has that one - here I passed two flags with the default, which isn't clear at all.  I've removed those, and commented.  thanks!

include/megaapi.h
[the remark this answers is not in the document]
sergiohs84: Thanks. I would have preferred the old-fashion `char *`/`const char *` to easy the bindings (IIRC, those types don't need anything special, while `string` will need specific declaration for memory management), but it's fine. We may need to change in a future though

include/megaapi.h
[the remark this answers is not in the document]
sergiohs84: Well, fetch the same data twice for no strong reason is definitely not the way to go :) But usually apps are aware of having queued a request, and they wait for the result. That's the approach we've been using for more than 5 years. That's why I didn't like too much this approach. However, now that apps can listen to the `onEvent(EVENT_MEDIA_INFO_READY)`, I think we are all good.
Also, I tried to convert the `ensureMediaInfo()` into a request, but it would require a bunch of changes and philosophy, since the command is launched from different places internally, as well as in result of app's calls, and it started to be complicated to decouple it. On top of that, it's specific to the `MediaInfo` and that's already legacy code for this feature, so I preferred to live with this `call and silently wait`, but it's not a desired approach, since it leaves black holes where things may start to fail

src/commands.cpp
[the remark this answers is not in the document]
sergiohs84: My bad, sorry. I looked at:
```void CommandPutFile::cancel()
{
    Command::cancel();
    tslot = NULL;
}

// set up file transfer with returned target URL
void CommandPutFile::procresult()
{
    if (tslot)
    {
        tslot->pendingcmd = NULL;
    }
    else
    {
        canceled = true;
    }```
and thought the flag was set thanks to that override. But you are right, indeed.
Closing discussion...

File: src/megaapi_impl.cpp (hunk 42 above)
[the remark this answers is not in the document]
Comment (sergiohs84) [reply src/megaapi_impl.cpp]: True. It's a tricky use-case. From other point of view, we could say the _media-upload_ does not finishes until its metadata are also uploaded and attached to the node.
I assume that Simon, in case he's working on that improvement, it's going to be done in a different branch than the main feature's branch. Hence, we can leave the required changes in the SDK for another branch too. But please, if so, create a ticket with the introduction to the issue, the link to the expected behavior explained by Mathias, and the suggested workaround.
Thanks

File: src/commands.cpp [hunk 44]
```diff
@@ -3064,6 +3132,96 @@ void CommandDelUA::procresult()
...
+    if (client->json.isnumeric())
+    {
+        error e = (error)client->json.getint();
+
+        if (e == API_ENOENT && fetching && maxtries > 0)
+        {
+            // we can't get it because it doesn't exist yet, so make it now
+            client->reqs.add(new CommandUnshareableUA(client, false, maxtries - 1));
+        }
+        return;
```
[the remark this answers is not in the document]
mattw-mega: good point, I have added those.

src/megaapi_impl.cpp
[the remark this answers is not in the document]
mattw-mega: Hi Sergio, good point.  We're so close to completing CU2 (the SDK part anyway), I thought it would be great to finalize that, so the interface is stable when it gets to `develop`.   I've pushed a change so the thumbnail and preview can be pushed with the initial putnodes, please take a look when you get a chance.  thanks!

src/megaapi_impl.cpp
```diff
@@ -882,6 +931,215 @@ MegaHandle MegaNodePrivate::getOwner() const
...
+            return nullptr;
+        }
+        else
+        {
+            // make sure we load to a chunk boundary
+            m_off_t endPos = ChunkedHash::chunkceil(startPos + *length, fain->size);
+            *length = endPos - startPos;
+            if (adjustsizeonly)
+            {
+                return "1";  // nonempty to indicate success
```
sergiohs84: ISO C++ forbids converting a string constant to ‘char*’. I'll fix it

mattw-mega:
trigger compilation

sergiohs84:
trigger compilation

PR #1341, opened 2019-01-23
Apply hotfixes from master

sergiohs84:
Hotfix/Increase the minimum backoff to enable the retrying state


sergiohs84:
There's no changes to apply

PR #1345, opened 2019-01-25
Do not ignore IN_CREATE for files

polmr:
given that certain ocassions (e.g: newer implementations of QFile::copy)
may not produce any other event for a file copied into a synched folder


polmr:
Replaced by this: https://github.com/meganz/sdk/pull/1350

PR #1348, opened 2019-01-31
Fix bonus upon creation of account

sergiohs84:
The ephemeral session needs to send the `si` to API in order to trigger the Invitation bonus. Otherwise, the new account will not trigger the MEGA Achievement for Invitation/referral.
Related ticket: #11431


sergiohs84:
Issue closed, patched in the API.

PR #1349, opened 2019-01-31
add methods to convert from attr number to name/long name

polmr:


src/megaapi_impl.cpp
```diff
@@ -4987,86 +4987,17 @@ void MegaApiImpl::addEntropy(char *data, unsigned int size)
...
-    }
+string MegaApiImpl::userAttributeToLongName(int type)
+{
+    return User::attr2longname((::mega::attr_t) type);
+}
 
-    return attrname;
+int MegaApiImpl::userAttributeFromString(const char *name)
+{
+    return User::string2attr(name);
```
sergiohs84: If `name` is NULL, it will cause undefined behavior
```suggestion
    if (!name)
    {
        return USER_ATTR_UNKNOWN;
    }
    return User::string2attr(name);
```

PR #1354, opened 2019-02-04
WIP: 16343: do always notify in case of IN_MOVED_FROM

polmr:
prevents case when haste execution of "mv a b; rm b"
lead to unnoticed removal of a

### DIFF
--- a/include/mega/posix/megafs.h
+++ b/include/mega/posix/megafs.h
@@ -118,6 +118,9 @@ public:
 
     PosixFileSystemAccess(int = -1);
     ~PosixFileSystemAccess();
+#ifdef ENABLE_SYNC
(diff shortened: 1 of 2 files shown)


cb-mega:
Hey Pablo, I had a look at this problem and your PR. There are some issues I see:
1. Your fix only addresses Posix but the two sync tests are failing on Windows as well
2. Your fix only passes `BasicSync.moveAndDeleteLocalFile` but `BasicSync.MoveSeveralExistingIntoDeepNewLocalFolders` is still failing

So I guess we could accept your PR and do a separate implementation for Windows later? Although not ideal I think.

What do you think would be required to pass `BasicSync.MoveSeveralExistingIntoDeepNewLocalFolders`?`

PR #1357, opened 2019-02-07
Ability for `MegaNode::getOwner()`

sergiohs84:
Not always the owner of the node is the own user, even for nodes in the
own cloud (ie. files/folders created by another user in an outgoing
folder).

### DIFF
--- a/src/megaapi_impl.cpp
+++ b/src/megaapi_impl.cpp
@@ -631,7 +631,7 @@ MegaNodePrivate *MegaNodePrivate::unserialize(string *d)
         LOG_err << "MegaNodePrivate unserialization failed - invalid version";
         return NULL;
     }
-    ptr += 7;
+    ptr += 6;
 
     string chatauth;
     if (hasChatAuth)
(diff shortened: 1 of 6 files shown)


src/megaapi_impl.cpp
```diff
@@ -615,7 +623,10 @@ MegaNodePrivate *MegaNodePrivate::unserialize(string *d)
     char hasChatAuth = MemAccess::get<char>(ptr);
     ptr += sizeof(char);
 
-    if (memcmp(ptr, "\0\0\0\0\0\0", 7))
+    char hasOwner = MemAccess::get<handle>(ptr);
+    ptr += sizeof(char);
+
+    if (memcmp(ptr, "\0\0\0\0\0", 6))
```
polmr: following `ptr += 7` shouldn't be `ptr += 6` ?

sergiohs84: Totally agree. Good catch, thanks

polmr:
trigger compilation

PR #1361, opened 2019-02-11
Release/v3.4.7

sergiohs84:
- Fix build of SQLite in Windows with `qmake`.
- Fix login into 2FA protected accounts in `megacli`.
- Public API to retrieve the owner of a node.


jnavarrom:
The iOS app version 4.4.4 is in the App Store. We can merge this PR.

PR #1363, opened 2019-02-13
Refactoring serialization for maintainability

mattw-mega:
Added classes CacheableReader and CacheableWriter, which wrap up the techniques already used for serialization and deserialisation.
The change means the code can be much more compact, enabling writing, maintaining, and reviewing those functions much more easily.

### DIFF
--- a/src/mediafileattribute.cpp
+++ b/src/mediafileattribute.cpp
@@ -420,6 +420,37 @@ MediaProperties::MediaProperties()
 {
 }
 
+MediaProperties::MediaProperties(const std::string& deserialize)
+{
+    CacheableReader r(deserialize);
+    r.unserializebyte(shortformat);
+    r.unserializeu32(width);
+    r.unserializeu32(height);
+    r.unserializeu32(fps);
+    r.unserializeu32(containerid);
+    r.unserializeu32(videocodecid);
+    r.unserializeu32(audiocodecid);
+    r.unserializebool(is_VFR);
+    r.unserializebool(no_audio);
+}
+
+std::string MediaProperties::serialize()
+{
+    std::string s;
+    CacheableWriter r(s);
+    r.serializebyte(shortformat);
+    r.serializeu32(width);
+    r.serializeu32(height);
+    r.serializeu32(fps);
+    r.serializeu32(containerid);
+    r.serializeu32(videocodecid);
+    r.serializeu32(audiocodecid);
+    r.serializebool(is_VFR);
+    r.serializebool(no_audio);
+    r.serializeexpansionflags();
+    return s;
+}
+
 bool MediaProperties::operator==(const MediaProperties& o) const
 { 
     return shortformat == o.shortformat && width == o.width && height == o.height && fps == o.fps && playtime == o.playtime &&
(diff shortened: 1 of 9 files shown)


src/mediafileattribute.cpp
```diff
@@ -420,6 +420,37 @@ MediaProperties::MediaProperties()
...
+    r.unserializebyte(shortformat);
+    r.unserializeu32(width);
+    r.unserializeu32(height);
+    r.unserializeu32(fps);
+    r.unserializeu32(containerid);
+    r.unserializeu32(videocodecid);
+    r.unserializeu32(audiocodecid);
+    r.unserializebool(is_VFR);
+    r.unserializebool(no_audio);
+}
```
polmr: I'm not sure if there will be a case where we wan this to receive `string *deserialize` ?  and then do `r.eraseused(deserialize);` by the end

tests/tests.cpp
```diff
@@ -50,6 +50,127 @@ TEST(Serialize64, serialize)
...
+    ASSERT_EQ(expansions[4], 0);
+    ASSERT_EQ(expansions[5], 0);
+    ASSERT_EQ(expansions[6], 1);
+    ASSERT_EQ(expansions[7], 1);
+
+    r.eraseused(readstring);
+    ASSERT_EQ(readstring, "abc");
+}
+
+
```
polmr: We might want to have some tests for serialize -> deserialize -> compare and unserialize -> serialize -> compare for different stored objects. But that can go out of the scope of this PR

src/utils.cpp
```diff
@@ -48,6 +48,227 @@ Cachable::Cachable()
...
+
+    memcpy(data, ptr, len);
+    ptr += len;
+    fieldnum += 1;
+    return true;
+}
+
+bool CacheableReader::unserializechunkmacs(chunkmac_map& m)
+{
+    return m.unserialize(ptr, end);   // ptr is adjusted by reference
```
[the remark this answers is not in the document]
mattw-mega: good spotting!  added

src/megaapi_impl.cpp
```diff
@@ -451,234 +451,75 @@ char *MegaNodePrivate::serialize()
...
-    char hasChatAuth = MemAccess::get<char>(ptr);
-    ptr += sizeof(char);
-
-    char hasOwner = MemAccess::get<handle>(ptr);
-    ptr += sizeof(char);
-
-    if (memcmp(ptr, "\0\0\0\0\0", 6))
-    {
-        LOG_err << "MegaNodePrivate unserialization failed - invalid version";
-        return NULL;
```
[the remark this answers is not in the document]
mattw-mega: yes, that was an oversight - I've added it in now.   Thanks for spotting that!

src/mediafileattribute.cpp
mattw-mega: Good point.  There isn't currently for this particular class, and if we do need it in the future then we can easily add it unconditionally, or with a boolean parameter on the function.   However I think in future we could move to the serialization/deserialization functions receiving CacheableWriter/Reader objects instead of strings, which would mean we could process member objects without any need for eraseused() at all, which would improve efficiency - but that would be too much to introduce in this PR.

tests/tests.cpp
mattw-mega: Yes - I've added one now for MediaProperties, and we can add more in the future when we use these in particular classes also. cheers

PR #1370, opened 2019-02-14
CloudRAID support

sergiohs84:
Initial implementation of CloudRAID(tm) client code in the SDK


src/raid.cpp
```diff
@@ -0,0 +1,911 @@
...
+        assert(connectionNum < tempurls.size());
+        return tempurls[connectionNum];
+    }
+    else if (!tempurls.empty())
+    {
+        return tempurls[0];
+    }
+    else
+    {
+        return emptyReturnString;
```
sergiohs84: Is it really safe to return an empty string here? Could HttpReqXfer::prepare() be called with an empty string? Not sure about it, just wondering.

src/raid.cpp
```diff
@@ -0,0 +1,911 @@
...
+        // start with 6 connections and drop the slowest to respond
+        // (unless we know one of the 6 URLs is on a dodgy server, in which case start with the other 5)
+       // unsigned worstindex = RAIDPARTS;
+        unsigned worstindex = rand() % RAIDPARTS;
+
+        MutexGuard g(m);
+        if (!recentFails.empty())
+        {
+            m_time_t now = m_time();
+            m_time_t worsttime = now - 10 * 3600;
```
sergiohs84: I suggest to use `NEVER`. Not sure if possible, but if any of the `url`s received as input is a hit in `recentFails` and the `ts` associated is older than 10 hours (likely to happen in MEGAsync), it would be excluded from the search of the worst index.
```suggestion
            m_time_t worsttime = NEVER;
```

mattw-mega: I was thinking that if a server is down, or temporarily can't serve that file, etc, it will probably be fixed on some timescale, maybe within a few hours and then we should resume normal operation rather than having the program avoid that server forever (in the absence of any other server getting into the data structure).   10 hours is probably not too frequent to try and see if problems have been resolved, and not too slow - what do you think?  thanks

src/raid.cpp
mattw-mega: the empty string there is just so that the function can return something legal (by reference) if the vector is empty.  That should never occur since the function shouldn't be used unless we got a reply for URLs to upload/download to/from.  I've added an assert in to clarify (and verify) that.  The return is by reference for efficiency, no need to copy strings in the normal case.  I've also changed it to a normal class member instead of a static, just in case of that initialization sequence issue.   thanks!

src/raid.cpp
sergiohs84: It's a good idea, indeed, to avoid using a faulty server for 10h and after that period of time to give it another chance. However, in that case, I think the:
```if (j != recentFails.end() && j->second > worsttime) ```
should be:
```if (j != recentFails.end() && j->second <= worsttime) ```
in order to select the faulty URL as the worst one, unless more than 10h have already elapsed. After 10h, the `worstindex` may randomly be assigned to the faulty one again.

Comment (sergiohs84) [reply src/raid.cpp]: This is the commit with the comments: https://github.com/meganz/sdk/commit/bc999f5d68ad34944cd2a681982d1bd4c786cc97

File: src/transfer.cpp [hunk 3]
```diff
@@ -1192,8 +1213,21 @@ void DirectReadNode::cmdresult(error e, dstime timeleft)
...
+
+            if (dr->drbuf.tempUrlVector().empty())
+            {
+                // DirectRead starting
+                dr->drbuf.setIsRaid(dr->drn->tempurls, dr->offset, dr->offset + dr->count, dr->drn->size, 2097152);  // 2 MB max buffer usage approx for streaming
+            }
+            else
+            {
+                // URLs have been re-requested, eg. due to temp URL expiry.  Keep any parts downloaded already
+                dr->drbuf.updateUrlsAndResetPos(dr->drn->tempurls);
```
sergiohs84: This is the only place in the code where URLs are updated (by keeping downloaded parts). I wonder if there's any other codepath where it should be done :thinking:

mattw-mega: For transfers a similar thing can happen when a transfer is rescheduled when the known tempurls have been cleared (eg. after a certain number of errors, or the URLs are too old after loading from cache).   I've kept the same logic for exactly when we fetch new URLs, except that they are stored in the RaidBufferManager now (to abstract away the use of 1 or 6 URLs).  Hopefully that makes it a bit easier to review, by keeping changes to a minimum.  thanks

src/transferslot.cpp
```diff
@@ -1115,6 +1138,33 @@ void TransferSlot::doio(MegaClient* client)
     }
 }
 
+
+bool TransferSlot::tryRaidRecoveryFromHttpGetError(unsigned connectionNum)
+{
+    // If we are downloding a cloudraid file then we may be able to ignore one connection and download from the other 5.
+    if (transferbuf.isRaid())
```
sergiohs84: Since `RadiBufferManager::tryRaidHttpGetErrorRecovery()` is only used here and it already checks the `isRaid()`, I'd suggest to convert the checkup in the `RadiBufferManager::tryRaidHttpGetErrorRecovery()` to an assert.

src/raid.cpp
```diff
@@ -0,0 +1,923 @@
...
+{
+    if (!isRaid())
+    {
+        return false;
+    }
+
+    // see if one connection is stalled or running much slower than the others, in which case try the other 5 instead
+    for (unsigned j = RAIDPARTS; j--; )
+    {
+        if (j != slowConnection && j != unusedRaidConnection && !connectionPaused[j])
```
sergiohs84: If the goal is to check whether the other *5* connections (excluding `slowConnection`) are paused in order to continue with those 5 instead of 6... I think it could happen that there's already one `unusedRaidConnection`, so already using 5 connections. If others than `slowConnection` are all paused, it may return `true` and, in consequence, would we start using 4 connections?
The case is as follows. Imagine you're checking `slowConnection = 0`, and the `unusedRaidConnection = 1`. Connections 2, 3, 4 and 5 are all paused. This code would return `true`.

sergiohs84: I reply myself: the `unusedRaidConnection` will be used again, and the `slowConnection` would be the new `unusedRaidConnection` :smiley:

src/transferslot.cpp
```diff
@@ -370,10 +389,36 @@ void TransferSlot::doio(MegaClient* client)
     {
         if (reqs[i])
         {
+            unsigned slowestConnection;
+            if (transfer->type == GET && reqs[i]->contentlength == reqs[i]->size)
+                if (transferbuf.detectSlowestRaidConnection(i, slowestConnection))
```
sergiohs84: I think this `if` should be replaced by one `&&` in the previous `if`. It makes it more readable and less prone to error if touched in the future.

mattw-mega: agreed, and done.  I had arranged it like that temporarily so I could breakpoint the last condition, and had not put it back :-)  cheers

src/raid.cpp
mattw-mega: :-)

src/transferslot.cpp
mattw-mega: Yes, good point, and done.  cheers

src/raid.cpp
```diff
@@ -0,0 +1,925 @@
...
+m_off_t& TransferBufferManager::transferPos(unsigned connectionNum)
+{
+    return isRaid() ? RaidBufferManager::transferPos(connectionNum) : transfer->pos;
+}
+
+m_off_t TransferBufferManager::nextTransferPos()
+{
+    assert(!isRaid());
+    chunkmac_map& chunkmacs = transfer->chunkmacs;
+    while (chunkmacs.find(ChunkedHash::chunkfloor(transfer->pos)) != chunkmacs.end())
```
sergiohs84: I don't know if the `chunkmacs` map may contain a lot of items, but since the item resulting from the `find()` is used inside the block perhaps is a good idea to declare an iterator and use it afterwords, avoiding 2 searches instead of one. I'm adding some comments, so I'll change it unless you prefer to keep it as it is currently ;)

mattw-mega: Hi Sergio, this is actually code that existed before (in transferslot) but I've moved it here as part of the abstraction.   As transfer->pos is being updated in the loop, and that's what find() is searching for, I think it's best left as is in this case (though, generally that is a good idea of course!)  thanks

sergiohs84: I've added a commit (not pushed yet) like this:
```
    chunkmac_map::const_iterator it = chunkmacs.find(ChunkedHash::chunkfloor(transfer->pos));
    while (it != chunkmacs.end())
    {
        if (it->second.finished)
        {
            transfer->pos = ChunkedHash::chunkceil(transfer->pos);
        }
        else
        {
            transfer->pos += it->second.offset;
            break;
        }
        it = chunkmacs.find(ChunkedHash::chunkfloor(transfer->pos));
    }```
But I can revert it, if you feel it's better to not change it :smiley:

Comment (mattw-mega) [reply src/raid.cpp]: Ah, I see what you mean now - yes that looks like a good improvement, thanks

**Comment by sergiohs84** [comment]
trigger compilation

**Comment by sergiohs84** [comment]
trigger compilation

File: src/commands.cpp [hunk 8]
```diff
@@ -443,14 +447,39 @@ void CommandDirectRead::procresult()
...
+                    }
+                    else
+                    {
+                        std::string tu;
+                        if (client->json.storeobject(&tu))
+                        {
+                            tempurls.push_back(tu);
+                        }
+                    }
+                    if (tempurls.size() == 1 || tempurls.size() == RAIDPARTS)
```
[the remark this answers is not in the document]
mattw-mega: Good idea, thanks, that is more informative than API_EINTERNAL, I've pushed the update.  I'll merge to `develop`, thanks again for the thorough review and testing!  cheers

PR #1373, opened 2019-02-15
Fix warning: Non-portable path to file '"MegaSDK+init.h"'; specified …

jnavarrom:
…path differs in case from file name on disk


jnavarrom:
trigger compilation

jnavarrom:
trigger compilation

PR #1385, opened 2019-03-15
Fix a crash seen on iOS, inside the outputs[] lookup in SimpleLogger:…

mattw-mega:
…:getOutput


polmr:
trigger compilation

polmr:
trigger compilation

polmr:
trigger compilation

PR #1386, opened 2019-03-21
support non-ascii pdf thumbnails in windows

polmr:
for non supported filenames use memory loader except for files > 100 MB
(use temporary file for those)


polmr:
trigger compilation

PR #1403, opened 2019-04-07
Move project towards cmake build system

kenkit:
EDIT:Tested with VS2017 WINDOWS 10
The static library builds also the testapp, I didn't have time to work on megacli and megasimplesynch but I assume it should be easy to add them. 
Also we should fix linux compatibilty switches.

EDIT:You should specify the location of 3rdparty folder when you run cmake generate command with
 ```-DTHIRDPARTY=folder_path```
or extract in this projects root directory.
EDIT
If not specified Cmake will try to download the dependancies and extract them for you.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,22 +30,40 @@ else()
     SET (7ZIP_EXECUTABLE "7z")
 endif()
 if(NOT EXISTS "${PROJECT_SOURCE_DIR}/3rdparty")
-    if(EXISTS "${PROJECT_SOURCE_DIR}/3rdparty.zip" OR FORCE_OPERATIONS)
+    if(EXISTS "${PROJECT_SOURCE_DIR}/3rdparty.zip")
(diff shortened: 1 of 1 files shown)


kenkit:
Now support auto-downloading of dependencies if dependencies path is not set or found.
Mega devs should set their own dependencies zip file in one of their repos and update url accordingly.

kenkit:
I think it should be possible to add appveyor as a windows build system.
When we fix cmake, it should be possible to add traviscl

kenkit:
My bad, I didn't see this https://github.com/meganz/sdk/blob/develop/contrib/cmake/CMakeLists.txt

mattw-mega:
Hi kenkit, yes as you see we already have a CMakeLists able to build the SDK on windows (and on Linux).   We're open to (preferably small and specific) improvements to it of course, but they should be aimed at the develop branch.  cheers

PR #1412, opened 2019-04-09
Allow streaming in IPv6 networks

javitrujillo:
From http://docs.libuv.org/en/v1.x/guide/networking.html#ipv6-stack-only

> IPv6 sockets can be used for both IPv4 and IPv6 communication

### DIFF
--- a/src/megaapi_impl.cpp
+++ b/src/megaapi_impl.cpp
@@ -23048,14 +23048,14 @@ void MegaTCPServer::run()
 
     uv_tcp_keepalive(&server, 0, 0);
 
-    struct sockaddr_in address;
+    struct sockaddr_in6 address;
     if (localOnly)
     {
-        uv_ip4_addr("127.0.0.1", port, &address);
+        uv_ip6_addr("::1", port, &address);
     }
     else
     {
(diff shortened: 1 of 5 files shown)


src/megaapi_impl.cpp
```diff
@@ -23267,7 +23267,7 @@ char *MegaTCPServer::getLink(MegaNode *node, string protocol)
     allowedHandles.insert(lastHandle);
 
     ostringstream oss;
-    oss << protocol << (useTLS ? "s" : "") << "://127.0.0.1:" << port << "/";
+    oss << protocol << (useTLS ? "s" : "") << "://[::1]:" << port << "/";
```
polmr: This causes issues in non IPv6 supportive clients: e.g: gvsfd-dav in Ubuntu 18.04. 
Browsers & other common clients (e.g vlc) seem to be good with it for local interface address (although fail to use scoped local links).  Will need some testing in different Windows versions in MEGAsync & probably some wording in MEGAcmd inner docs to alert possible issues

javitrujillo: We could add a parameter to the `getLink` function to return either an IPv4 URL or an IPv6 one, default IPv4 to keep compatibility with those clients.

PR #1414, opened 2019-04-10
Fix building with C++11 using qmake

sergiohs84:
In older OS like Ubuntu, the default C++ standard is c++0x -->
compilation fails.


polmr:
change included in https://github.com/meganz/sdk/pull/1405

PR #1418, opened 2019-04-15
Hotfix - Add to utils the function m_clock_gettime

alber2510:


src/utils.cpp
```diff
@@ -1459,6 +1459,36 @@ m_time_t m_mktime(struct tm* stm)
     return mktime(stm);
 }
 
+int m_clock_gettime(clockid_t clock_id, timespec *t)
+{
+    if (!t)
```
sergiohs84: IMHO, this is an internal method. We can assume the caller will always pass a valid pointer

sergiohs84:
@mattw-mega please, have a look at the `clock_gettime()` for Windows. We used your code from the MEGAchat repo: `InitStats::currentTime()`. Thanks

PR #1422, opened 2019-04-18
transfer resumption: make sure to `setIsRaid` on resumption if it alr…

mattw-mega:
…eady konws its URLs

also updated cmake script for USE_QT

### DIFF
--- a/examples/megacli.cpp
+++ b/examples/megacli.cpp
@@ -2116,14 +2116,15 @@ public:
 
     static string server(const string& url)
     {
-        size_t n = url.find("://");
-        if (n != string::npos)
+        const string pattern("://");
+        size_t start_index = url.find(pattern);
+        if (start_index != string::npos)
         {
-            n += 3;
-            size_t m = url.find("/", n);
-            if (m != string::npos)
+            start_index += pattern.size();
+            const size_t end_index = url.find("/", start_index);
+            if (end_index != string::npos)
             {
-                return url.substr(n, m - n);
+                return url.substr(start_index, end_index - start_index);
             }
         }
         return "";
@@ -2167,7 +2168,7 @@ public:
                             {
                                 cout << n->displaypath() << endl;
 
-                                for (auto& url : tempurls)
+                                for (const auto& url : tempurls)
                                 {
                                     stack->servers.insert(server(url));
                                 }
--- a/include/mega/raid.h
+++ b/include/mega/raid.h
@@ -56,7 +56,7 @@ namespace mega {
         void setIsRaid(const std::vector<std::string>& tempUrls, m_off_t resumepos, m_off_t readtopos, m_off_t filesize, m_off_t maxDownloadRequestSize);
 
         // indicate if the file is raid or not.  Most variation due to raid/non-raid is captured in this class
-        bool isRaid();
+        bool isRaid() const;
 
         // in case URLs expire, use this to update them and keep downloading without wasting any data
         void updateUrlsAndResetPos(const std::vector<std::string>& tempUrls);
@@ -101,7 +101,7 @@ namespace mega {
         bool detectSlowestRaidConnection(unsigned thisConnection, unsigned& slowestConnection);
 
         // returns how far we are through the file on average, including uncombined data
-        m_off_t progress();
+        m_off_t progress() const;
 
         RaidBufferManager();
         ~RaidBufferManager();
--- a/src/megaclient.cpp
+++ b/src/megaclient.cpp
@@ -3240,10 +3240,6 @@ bool MegaClient::dispatch(direction_t d)
                               << " Partial: " << p << " Size: " << nexttransfer->size
                               << " ultoken: " << (nexttransfer->ultoken != NULL);
 
-                    if (!nexttransfer->tempurls.empty())
-                    {
-                        ts->transferbuf.setIsRaid(nexttransfer, nexttransfer->tempurls, nexttransfer->pos, ts->maxRequestSize);
-                    }
                 }
                 else
                 {
@@ -3298,6 +3294,7 @@ bool MegaClient::dispatch(direction_t d)
                 // dispatch request for temporary source/target URL
                 if (nexttransfer->tempurls.size())
                 {
+                    ts->transferbuf.setIsRaid(nexttransfer, nexttransfer->tempurls, nexttransfer->pos, ts->maxRequestSize);
                     app->transfer_prepare(nexttransfer);
                 }
                 else
--- a/src/megaclient.cpp
+++ b/src/megaclient.cpp
@@ -13003,7 +13003,7 @@ node_vector MegaClient::getRecentNodes(unsigned maxcount, m_time_t since, bool i
 
     // 2. Order them chronologically and restrict them to a maximum of `maxcount`
     node_vector v2;
-    unsigned maxItems = std::max(maxcount, unsigned(v.size()));
+    unsigned maxItems = std::max<unsigned>(maxcount, unsigned(v.size()));
     v2.reserve(maxItems);
     while (v2.size() < maxItems && !v.empty())
     {
--- a/src/megaclient.cpp
+++ b/src/megaclient.cpp
@@ -24,6 +24,10 @@
 #include <cctype>
 #include <algorithm>
 
+#undef min // avoid issues with std::min and std::max
+#undef min
+#undef max
+
 namespace mega {
 
 // FIXME: generate cr element for file imports
@@ -13002,8 +13006,9 @@ node_vector MegaClient::getRecentNodes(unsigned maxcount, m_time_t since, bool i
     std::make_heap(v.begin(), v.end(), nodes_ctime_less);
 
     // 2. Order them chronologically and restrict them to a maximum of `maxcount`
+
     node_vector v2;
-    unsigned maxItems = std::max<unsigned>(maxcount, unsigned(v.size()));
+    unsigned maxItems = std::max(maxcount, unsigned(v.size()));
     v2.reserve(maxItems);
     while (v2.size() < maxItems && !v.empty())
     {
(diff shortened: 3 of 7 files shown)


src/raid.cpp
```diff
@@ -767,6 +771,27 @@ bool RaidBufferManager::detectSlowestRaidConnection(unsigned thisConnection, uns
     return false;
 }
 
+
+m_off_t RaidBufferManager::progress()
+{
+    assert(isRaid());
+    m_off_t reportPos = 0;
+
+    for (unsigned j = RAIDPARTS; j--; )
```
cb-mega: Still think a reverse loop should only be used if there's a reason for it. Otherwise might confuse the reader...

src/transferslot.cpp
```diff
@@ -433,6 +433,20 @@ void TransferSlot::doio(MegaClient* client)
                 continue;
             }
 
+            if (reqs[i]->status == REQ_FAILURE && reqs[i]->httpstatus == 200 && transfer->type == GET && transferbuf.isRaid())
+            {
+                // check if we got some data and the failure occured partway through the part chunk.  If so, best not to waste it, convert to success case with less data
+                HttpReqDL *downloadRequest = (HttpReqDL *)reqs[i];
```
cb-mega: `static_cast`?

src/transferslot.cpp
```diff
@@ -433,6 +433,20 @@ void TransferSlot::doio(MegaClient* client)
...
+                LOG_debug << "Connection " << i << " received " << downloadRequest->bufpos << " before failing, processing data.";
+                if (downloadRequest->contentlength == downloadRequest->size && downloadRequest->bufpos >= RAIDSECTOR)
+                {
+                    downloadRequest->bufpos -= downloadRequest->bufpos % RAIDSECTOR;  // always on a raidline boundary
+                    downloadRequest->size = unsigned(downloadRequest->bufpos);
```
cb-mega: `static_cast`?

src/transferslot.cpp
cb-mega: And maybe just get a reference instead of a pointer?

src/transferslot.cpp
```diff
@@ -433,6 +433,20 @@ void TransferSlot::doio(MegaClient* client)
                 continue;
             }
 
+            if (reqs[i]->status == REQ_FAILURE && reqs[i]->httpstatus == 200 && transfer->type == GET && transferbuf.isRaid())
```
cb-mega: Can you leave a comment as to why `200` is to be checked for here?

src/transferslot.cpp
```diff
@@ -1083,7 +1097,21 @@ void TransferSlot::doio(MegaClient* client)
         }
     }
 
-    p += transfer->progresscompleted;
+    if (transfer->type == GET && transferbuf.isRaid())
```
cb-mega: Since we don't put this stuff into functions can we leave comments as to what the individual blocks of code do?

src/transferslot.cpp
```diff
@@ -1083,7 +1097,21 @@ void TransferSlot::doio(MegaClient* client)
...
+    {
+        p = transferbuf.progress();
```
cb-mega: Any way we can rename `p` to `currentProgress` or something?

examples/megacli.cpp
```diff
@@ -2113,6 +2114,21 @@ class FileFindCommand : public Command
         }
     }
 
+    static string server(const string& url)
+    {
+        size_t n = url.find("://");
```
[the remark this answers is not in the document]
mattw-mega: done (except const, as it does change)

examples/megacli.cpp
```diff
@@ -2113,6 +2114,21 @@ class FileFindCommand : public Command
...
+        if (n != string::npos)
+        {
+            n += 3;
```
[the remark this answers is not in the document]
mattw-mega: all set

examples/megacli.cpp
```diff
@@ -2113,6 +2114,21 @@ class FileFindCommand : public Command
...
+            size_t m = url.find("/", n);
```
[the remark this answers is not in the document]
mattw-mega: done

examples/megacli.cpp
```diff
@@ -2150,6 +2166,11 @@ class FileFindCommand : public Command
                             if (Node* n = client->nodebyhandle(h))
                             {
                                 cout << n->displaypath() << endl;
+
+                                for (auto& url : tempurls)
```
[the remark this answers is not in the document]
mattw-mega: yes, updated

examples/megacli.cpp
```diff
@@ -2171,6 +2192,10 @@ class FileFindCommand : public Command
         else if (!stack->filesLeft)
         {
             cout << "<find complete>" << endl;
+            for (auto s : stack->servers)
```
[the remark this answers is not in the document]
mattw-mega: oops - missed this one earlier somehow, and I've already pushed the update.   Can we leave it for next time?

include/mega/raid.h
```diff
@@ -147,6 +150,9 @@ namespace mega {
         // the point we are at in the output file.  asyncoutputbuffers contain data from this point.
         m_off_t outputfilepos;
 
+        // the point we started at in the output file.
+        m_off_t startfilepos;
```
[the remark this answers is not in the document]
mattw-mega: It wasn't used until after setIsRaid (which must be the first thing done).  I've set it to 0 for completeness though

include/mega/raid.h
```diff
@@ -100,6 +100,9 @@ namespace mega {
         // indicate that this connection has responded with headers, and see if we now know which is the slowest connection, and make that the unused one
         bool detectSlowestRaidConnection(unsigned thisConnection, unsigned& slowestConnection);
 
+        // returns how far we are through the file on average, including uncombined data
+        m_off_t progress();
```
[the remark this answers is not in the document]
mattw-mega: done

src/megaclient.cpp
```diff
@@ -13001,7 +13003,7 @@ node_vector MegaClient::getRecentNodes(unsigned maxcount, m_time_t since, bool i
 
     // 2. Order them chronologically and restrict them to a maximum of `maxcount`
     node_vector v2;
-    unsigned maxItems = std::max(maxcount, unsigned(v.size()));
+    unsigned maxItems = std::max<unsigned>(maxcount, unsigned(v.size()));
```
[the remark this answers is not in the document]
mattw-mega: Apparently not when compiling in QT, with the Visual Studio 2015 compiler :-(

src/raid.cpp
```diff
@@ -690,12 +691,15 @@ bool RaidBufferManager::tryRaidHttpGetErrorRecovery(unsigned errorConnectionNum)
     g_faultyServers.add(tempurls[errorConnectionNum]);
 
     unsigned errorSum = 0;
+    unsigned highestErrors = 0;
     for (unsigned i = RAIDPARTS; i--; )
     {
         errorSum += raidHttpGetErrorCount[i];
+        highestErrors = std::max<unsigned>(highestErrors, raidHttpGetErrorCount[i]);
```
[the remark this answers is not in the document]
mattw-mega: same as above

src/raid.cpp
```diff
@@ -690,12 +691,15 @@ bool RaidBufferManager::tryRaidHttpGetErrorRecovery(unsigned errorConnectionNum)
...
     }
 
-    if (errorSum < 3)
+    // Allow for one nonfunctional channel and one glitchy channel.  We can still make progress swapping back and forth
+    if ((errorSum - highestErrors) < 5)
```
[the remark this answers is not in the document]
mattw-mega: yes, updated

src/raid.cpp
```diff
@@ -767,6 +771,27 @@ bool RaidBufferManager::detectSlowestRaidConnection(unsigned thisConnection, uns
...
 
+
+m_off_t RaidBufferManager::progress()
+{
+    assert(isRaid());
+    m_off_t reportPos = 0;
+
+    for (unsigned j = RAIDPARTS; j--; )
+    {
+        for (auto& p : raidinputparts[j])
```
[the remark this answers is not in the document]
mattw-mega: I've switched it to FilePiece*, but can't make it const as the isNull() function is not const - changing that is probably getting too far away from the purpose of this PR.

src/raid.cpp
mattw-mega: I think anyone familiar with our code base knows about this one :-)  It is a very nice, very concise method once you're used to it

src/transferslot.cpp
mattw-mega: updated with static_cast.  I had followed the existing usage, but have swapped all of them now, total of 8 in the file.  I've left them as pointers, I think it's more natural to keep that the same on both sides.

src/transferslot.cpp
mattw-mega: I think this is an old update to suppress warnings.  I prefer the shorter syntax for these, or changing the variable types to match but I don't think we can change the type of either of those.   Since static_cast was deliberately designed to be verbose and ugly to discourage its use, I'd rather use the shorter expression here?

src/transferslot.cpp
mattw-mega: done

src/transferslot.cpp
mattw-mega: comment added :-)

src/transferslot.cpp
mattw-mega: That variable has been there such a long time and I it's usage is quite compilcated and spread through this large function.  I think it would need a thorough analysis to ensure the name matched its use properly.   In this case I'm just changing its value only for the isRaid() case, before it's existing use in reporting progress.

src/raid.cpp
sergiohs84: I tend to agree with Christian here, but let's leave it as it is :)

PR #1431, opened 2019-04-29
Enable putting commands in separate batches for server execution.

mattw-mega:
Reworked the Request queue which used to have two batches and send them alternately.  Commands can be marked to be put in their own batch, or for the bactch to be sent without SID (for the misc flags command).
Some of the Request queue logic is formatted with an eye to processing the response items gradually, for a future feature (that this code is adapted from).


src/commands.cpp
```diff
@@ -3377,6 +3384,13 @@ void CommandGetUserData::procresult()
             }
             break;
 
+        case MAKENAMEID4('s', 'm', 's', 'v'):
+            if (!client->json.storeobject(&smsv))
```
sergiohs84: Why are we parsing this parameter?? it's used nowhere. Does `smsv` value represent the phone-number configured by the user for SMS verification? I don't know. However, the API command `smsv` is used to verify a received code, so it's strange to me to see it in the `CommandGetUserData`

sergiohs84: Just read the `smsv` will be the phone-number you have verified. Since this parameter is not used and unrelated to this pull-request, it would be better to skip it. Same applies to the `smsve` inside the `flags` array.

src/commands.cpp
```diff
@@ -3363,6 +3367,9 @@ void CommandGetUserData::procresult()
                     case MAKENAMEID5('a', 'p', 'l', 'v', 'p'):   // apple VOIP push enabled
                         aplvp = bool(client->json.getint());
                         break;
+                    case MAKENAMEID5('s', 'm', 's', 'v', 'e'):   // 2 = Opt-in and unblock SMS allowed 1 = Only unblock SMS allowed 0 = No SMS allowed
```
sergiohs84: Unrelated to this PR. IMHO it should not be here.

src/commands.cpp
```diff
@@ -3416,6 +3430,60 @@ void CommandGetUserData::procresult()
     }
 }
 
+CommandGetMiscFlags::CommandGetMiscFlags(MegaClient *client)
```
sergiohs84: This command is unrelated to this PR. Also, FYI, it's missing the new `smsve` flag, which probably is included in the response.

src/request.cpp
```diff
@@ -25,17 +25,166 @@
 #include "mega/megaclient.h"
 
 namespace mega {
+
+//ProcResultQueue::entry::~entry()
```
sergiohs84: We have a powerful versioning tool, no need to keep obsolete code here as comments. Please, clean it up.

include/mega/request.h
```diff
@@ -23,50 +23,65 @@
...
+    deque<Request> nextreqs;
 
     static const int MAX_COMMANDS = 10000;
 
 public:
     RequestDispatcher();
 
-    void nextRequest();
-
+    // Queue a command to be send to MEGA. Some commands must go in their own batch (in case other commands fail the whole batch), in which case set solo=true
```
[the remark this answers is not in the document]
mattw-mega: good spotting, thanks - that is a very out of date comment, that was originally a parameter on add() but didn't work well.  Updated

src/commands.cpp
```diff
@@ -1583,6 +1584,7 @@ void CommandPrelogin::procresult()
 CommandLogin::CommandLogin(MegaClient* client, const char* email, const byte *emailhash, int emailhashsize, const byte *sessionkey, int csessionversion, const char *pin)
 {
     cmd("us");
+    batchSeparately = true;  // in case the account is blocked (we need to get a sid so we can issue whyamiblocked)
```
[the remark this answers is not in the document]
mattw-mega: Thanks for the scenario, which I hadn't tested.  I've tried it out now, and the ephemeral session resumption and `whyamiblocked()` succeed as it is, I'll send you the log to confirm.  cheers

src/commands.cpp
mattw-mega: thanks for spotting that; I've removed those in this PR.  cheers

src/commands.cpp
mattw-mega: yes, I had removed the smsve related code in this one already.   I think we should keep the command in the PR though, as it is one of the ones that needs to be batched independently, which is what this PR is about, and it's not specific to SMS (except that we had never noticed that we needed it until that project).   I think it's just as easy to review in this PR, and we won't call it until the SMS changes are merged so it won't break anything.   I hope that sounds ok - thanks!

src/request.cpp
mattw-mega: quite right, I thought I had already done that actually - maybe I had only removed it from the header.

src/commands.cpp
mattw-mega: updated 👍

src/commands.cpp
sergiohs84: Alright, let's keep this new command in the PR. But in general terms, we should avoid introducing unrelated features. It makes code's management easier (in terms of debugging, tracking issues, to guess the root of them, etc) ;)

PR #1432, opened 2019-04-29
Fix progress calculation for cloudraid files

mattw-mega:
Include any progress prior to pause/resume or serialize/deserialize


sergiohs84:
This fix is considered a hotfix. It's replaced by https://github.com/meganz/sdk/pull/1433

PR #1434, opened 2019-04-29
Business accounts

jgandres:
Add support for business accounts in SDK

### DIFF
--- a/include/mega/megaclient.h
+++ b/include/mega/megaclient.h
@@ -1438,12 +1438,17 @@ public:
     // the SDK is trying to log out
     int loggingout;
 
-    // Account type (pro level)
-    int proLevel;
-
-    //Business flags
+    // true if the account is a business account
     bool business;
+
+    // true if the account is a master business account, false if it's a sub-user account
     bool businessMaster;
+
+    /** business account status, the possible values are:
+     *      - MegaApi::BUSINESS_STATUS_EXPIRED = -1
+     *      - MegaApi::BUSINESS_STATUS_ACTIVE = 1
+     *      - MegaApi::BUSINESS_STATUS_GRACE_PERIOD = 2
+     */
     int businessStatus;
 
     MegaClient(MegaApp*, Waiter*, HttpIO*, FileSystemAccess*, DbAccess*, GfxProc*, const char*, const char*);
--- a/include/mega/megaclient.h
+++ b/include/mega/megaclient.h
@@ -1444,11 +1444,7 @@ public:
     // true if the account is a master business account, false if it's a sub-user account
     bool businessMaster;
 
-    /** business account status, the possible values are:
-     *      - MegaApi::BUSINESS_STATUS_EXPIRED = -1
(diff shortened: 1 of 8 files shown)


include/mega/types.h
```diff
@@ -531,6 +532,23 @@ struct recentaction
...
+{
+    ACCOUNT_TYPE_FREE = 0,
+    ACCOUNT_TYPE_PROI = 1,
+    ACCOUNT_TYPE_PROII = 2,
+    ACCOUNT_TYPE_PROIII = 3,
+    ACCOUNT_TYPE_LITE = 4,
+    ACCOUNT_TYPE_BUSINESS = 100
+} account_t;
+
+typedef enum
```
sergiohs84: Move the enumeration to the public interface (`MegaApi` or any suitable `MegaXXX` object).

include/mega/types.h
```diff
@@ -531,6 +532,23 @@ struct recentaction
...
 
+typedef enum
+{
+    ACCOUNT_TYPE_FREE = 0,
+    ACCOUNT_TYPE_PROI = 1,
+    ACCOUNT_TYPE_PROII = 2,
+    ACCOUNT_TYPE_PROIII = 3,
+    ACCOUNT_TYPE_LITE = 4,
+    ACCOUNT_TYPE_BUSINESS = 100
+} account_t;
```
sergiohs84: Remove this enumeration, it's not needed and the apps should use the one from `MegaAccountDetails`

include/mega/megaclient.h
```diff
@@ -1438,6 +1438,14 @@ class MEGA_API MegaClient
     // the SDK is trying to log out
     int loggingout;
 
+    // Account type (pro level)
+    int proLevel;
```
sergiohs84: Get rid of `proLevel`, since it should be retrieved in the old way, through `uq` command (`getuseraccountdetails()`?)

src/commands.cpp
```diff
@@ -3377,6 +3380,35 @@ void CommandGetUserData::procresult()
...
+                            endobject = true;
+                            break;
+                        default:
+                            if (!client->json.storeobject())
+                            {
+                                return client->app->userdata_result(NULL, NULL, NULL, jid, API_EINTERNAL);
+                            }
+                    }
+                }
+                client->json.leaveobject();
```
sergiohs84: I miss some integrity checks for the received data.

src/commands.cpp
```diff
@@ -3394,6 +3426,14 @@ void CommandGetUserData::procresult()
             client->nsr_enabled = nsre;
             client->aplvp_enabled = aplvp;
             client->k = k;
+
+            if (b)
```
sergiohs84: I think not possible now, but what if a business account becomes a non-business in runtime?
We need to review this code.

src/commands.cpp
sergiohs84: done

src/commands.cpp
sergiohs84: done!

sergiohs84:
trigger compilation

PR #1439, opened 2019-05-03
RequestDispatcher: fix `logout` response calling `clear()` on active …

mattw-mega:
…structure

Since RequestDispatcher::clear() is called from one of the items managed by the dispatcher, and clear() wipes the entire data structure,
delay that wiping until the function returns (and avoid calling any further command responses in the meantime).
Once we can be sure that none of that data structure is in use anywhere, we perform the clear().
(also fixed a merge issue with the CMakeLists.txt)

### DIFF
--- a/contrib/cmake/CMakeLists.txt
+++ b/contrib/cmake/CMakeLists.txt
@@ -14,14 +14,9 @@
 
 cmake_minimum_required(VERSION 3.7.2)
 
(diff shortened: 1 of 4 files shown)


sergiohs84:
This PR should target `release/v3.5.1` in case we want to rebuild new binaries for native apps, since it fixes a hotfix already merged into the release branch.

sergiohs84:
Unfortunately, it seems this branch was not forked from `release` branch, but from `develop`. Need to rebase, @mattw-mega

PR #1451, opened 2019-05-10
improvements for building on windows, from release 3.5.2

mattw-mega:
pdfium debug build
cmake reference to pdfium
qt.cpp had problems streaming QT strings to log
constexpr in raid.cpp wasn't accepted by older compilers

### DIFF
--- a/src/gfx/qt.cpp
+++ b/src/gfx/qt.cpp
@@ -395,7 +395,7 @@ GfxProcQT::GfxProcQT()
     dir.setFilter(QDir::Files);
     foreach(QString dirFile, dir.entryList())
     {
-        LOG_warn << "Removing unexpected temporary file found from previous executions: " << dirFile.toStdString();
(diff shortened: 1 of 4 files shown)


src/gfx/qt.cpp
```diff
@@ -395,7 +395,7 @@ GfxProcQT::GfxProcQT()
     dir.setFilter(QDir::Files);
     foreach(QString dirFile, dir.entryList())
     {
-        LOG_warn << "Removing unexpected temporary file found from previous executions: " << dirFile;
+        LOG_warn << "Removing unexpected temporary file found from previous executions: " << dirFile.toStdString();
```
polmr: I think we'd rather use `.toUtf8().constData()` : toStdString return encodes in UTF-8 in qt 5, but as utf8 in qt4

mattw-mega: updated with that change, thanks!

mattw-mega:
trigger compilation

PR #1453, opened 2019-05-13
Use system-wide monotonic clock that includes time spent in suspension

polmr:


src/utils.cpp
```diff
@@ -1481,9 +1481,13 @@ int m_clock_getmonotonictime(timespec *t)
     t->tv_sec = tb.time;
     t->tv_nsec = tb.millitm = tb.millitm * 1000000;
     return 0;
+#else
+#ifdef CLOCK_BOOTTIME
```
sergiohs84: Maybe better to add it as another `#elif CLOCK_BOOTTIME`? So we don't open an extra `ifdef` block

PR #1458, opened 2019-05-14
Feature/address book

cb-mega:
This implements two new commands:
CommandGetRegisteredContacts
CommandGetCountryCallingCodes
and gives access to those through the api layer.


include/mega/command.h
```diff
@@ -28,6 +28,11 @@
 #include "http.h"
 
 namespace mega {
+
+struct JSON;
+struct MegaApp;
+class MegaStringMap;
```
mattw-mega: MegaStringMap is part of the public interface, and so can't be used in the core code as some clients are built without the public api (ie. megaapi_impl.cpp is not included)  It's a one-way dependency (and I made that mistake initially too :-) )

include/mega/command.h
```diff
@@ -86,7 +91,9 @@ class MEGA_API Command
     const char* getstring() const;
 
     Command();
-    virtual ~Command() { }
+    virtual ~Command() = default;
+
+    MEGA_DISABLE_COPY_MOVE(Command)
```
mattw-mega: Having looked at this a bit closer, there are no owning pointers etc in this class that would cause problems should one of these be copied.   Should we really disallow something that's not problematic?  There could be legitimate uses for it, and in fact there could be third party clients relying on that somehow.

include/megaapi.h
```diff
@@ -2125,13 +2132,93 @@ class MegaStringList
...
+    virtual ~MegaStringTable();
+
+    static MegaStringTable *createInstance();
+
+    virtual MegaStringTable* copy() const;
+
+    /**
+     * @brief Appends a new string list to the end of the table
+     *
+     * The parameter passed is copied.
```
mattw-mega: I think most clients would construct the string list specifically for this call, and then would have to delete it after passing it to this function, meaning we are forcing a second copy to be created.   Wouldn't it be more efficient to take ownership of the parameter, and if the client wants to keep and use the original, they can easily just call copy() on the list.

include/megaapi.h
```diff
@@ -2125,13 +2132,93 @@ class MegaStringList
...
+    virtual const MegaStringList* get(const char* key) const;
+
+    /**
+     * @brief Sets a value in the map for the given key.
+     *
+     * If the key already exists, the value will be overwritten by the
+     * new value.
+     *
+     * The map does not take ownership of the parameters passed, it makes
+     * a local copy.
```
mattw-mega: Same here - wouldn't it be more efficient and flexible to take ownership of the string list object?

include/megaapi.h
```diff
@@ -13976,7 +14093,25 @@ class MegaApi
...
+         *
+         * @param contacts The map of contacts to get registered contacts from
+         * @param listener MegaRequestListener to track this request
+         */
+        void getRegisteredContacts(const MegaStringMap* contacts, MegaRequestListener *listener = NULL);
 
+        /**
+         * @brief Requests the currently available country calling codes
+         *
+         * Check the unit tests for an example response.
```
mattw-mega: I think we should at least name the request type and which getters on the request are valid on success?   Most consumers of the API would expect to find all the info about using it here.

include/megaapi.h
```diff
@@ -13976,7 +14093,25 @@ class MegaApi
...
         void checkSMSVerificationCode(const char* verificationCode, MegaRequestListener *listener = NULL);
 
+        /**
+         * @brief Requests the contacts that are registered at MEGA (currently verified through SMS)
+         *
+         * The request will return any of the provided contacts that are registered at MEGA, i.e.,
+         * are verified through SMS (currently). Check the unit tests for an example response.
+         *
+         * @param contacts The map of contacts to get registered contacts from
+         * @param listener MegaRequestListener to track this request
```
mattw-mega: same here - we should name the request type and valid members

include/megaapi_impl.h
```diff
@@ -1102,6 +1108,8 @@ class MegaRequestPrivate : public MegaRequest
         MegaTextChatList *chatList;
 #endif
         MegaStringMap *stringMap;
+        MegaStringListMap *stringListMap;
+        MegaStringTable *stringTable;
```
mattw-mega: maybe `=nullptr;` now we have c++11, to make it easy to be sure these are appropriately set in the constructor?

src/commands.cpp
```diff
@@ -19,6 +19,7 @@
  * program.
  */
 
+#include "megaapi.h"
```
mattw-mega: as megaapi.h is the public interface, we can't have a dependency on it in the core code

src/commands.cpp
```diff
@@ -7007,5 +7008,186 @@ void CommandSMSVerificationCheck::procresult()
...
+    }
+    app.getregisteredcontacts_result(API_OK, &registered_contacts);
+}
+
+void CommandGetRegisteredContacts::procresult()
+{
+    processResult(*client->app, client->json);
+}
+
+CommandGetRegisteredContacts::CommandGetRegisteredContacts(MegaClient* client, const MegaStringMap& contacts)
```
mattw-mega: The pattern in this .cpp file is that the constructor is first, then the procresult() after - I've been pulled up on that one :-)

include/megaapi_impl.h
mattw-mega: actually, as these are owning pointers and we now have c++11, we could use unique_ptr<> (which would make the constructor, setters, and destructor code simpler)

src/megaapi_impl.cpp
```diff
@@ -3572,68 +3604,130 @@ MegaStringMapPrivate::MegaStringMapPrivate(const MegaStringMapPrivate *megaStrin
     delete keys;
 }
 
-MegaStringListPrivate::MegaStringListPrivate()
+
+MegaStringListPrivate::MegaStringListPrivate(char **newlist, int size)
+{
+    for (int i = 0; i < size; i++)
+    {
+        append(newlist[i]);
```
mattw-mega: This function used to take ownership of the parameters but now copies them - are all the callers adjusted, and can we be sure this isn't introducing a memory leak?

src/megaapi_impl.cpp
```diff
@@ -3572,68 +3604,130 @@ MegaStringMapPrivate::MegaStringMapPrivate(const MegaStringMapPrivate *megaStrin
...
-        return;
+        return m_list[i].c_str();
     }
+    return nullptr;
+}
 
-    list = new const char*[s];
-    for (int i = 0; i < s; i++)
-        list[i] = MegaApi::strdup(stringList->get(i));
+int MegaStringListPrivate::size() const
```
mattw-mega: I'm finding it very hard to review the changes to MegaStringListPrivate, as the diffs are all mixed up with the code for the  new classes, and also I think it's introducing an inefficiency where it used to just take ownership of some pointers, and now copies the content instead, affecting any call sites as well as the class itself.    I think we should leave this one alone - an "if it ain't broke" case

src/megaapi_impl.cpp
```diff
@@ -19955,6 +20117,24 @@ void MegaApiImpl::sendPendingRequests()
...
+        case MegaRequest::TYPE_GET_REGISTERED_CONTACTS:
+        {
+            const auto contacts = request->getMegaStringMap();
+            if (contacts)
+            {
+                client->reqs.add(new CommandGetRegisteredContacts{client, *contacts});
+            }
+            else
+            {
+                assert(false && "contacts must be valid");
```
mattw-mega: we should also set                 e = API_EARGS;  so that non-debug builds behave as expected (listeners always complete, and with an appropriate error code)

include/mega/command.h
cb-mega: Ok, will resolve

include/mega/command.h
cb-mega: I think we should disallow because copying/moving `levels` doesn't work by default. So either we implement explicit semantics or disallow. What do you think?

include/megaapi.h
cb-mega: That feels like a premature optimization to me. It's safer to take ownership and makes more sense to me as it's the default behaviour for most containers in the wild.

include/megaapi.h
cb-mega: But then shouldn't we be consistent with other data structures in this module as well. Shouldn't they take ownerships of all elements? Like StringList should then also take ownership etc

include/megaapi.h
cb-mega: I think we should describe what `contacts` actually holds (same with the result). But I don't feel like we should describe the query as that's hidden and the client shouldn't need to know.

include/megaapi.h
cb-mega: Will add more info

include/megaapi_impl.h
cb-mega: Since there are many other pointers I want to keep it consistent and initialize in the constructor bodies and keep raw pointers. However, I did forget to delete in the destructor. Adding that...

src/commands.cpp
cb-mega: Will fix

src/commands.cpp
cb-mega: That is weird. Shouldn't the order in the cpp match the header??

src/megaapi_impl.cpp
cb-mega: I actually thought this was a bug that it wouldn't copy. I will check...

src/megaapi_impl.cpp
cb-mega: Yes, you're right. I thought there was a bug but it probably is an issue in my test. Will change back...

src/megaapi_impl.cpp
cb-mega: Will do

include/megaapi.h
cb-mega: After discussion IRL, will make the two new data structures take ownership of the string list passed.

include/megaapi.h
cb-mega: After discussion IRL, will make the two new data structures take ownership of the string list passed.

include/mega/command.h
cb-mega: After IRL realized that `levels` is fine to copy/move. Agreed IRL to default all special member functions.

include/megaapi.h
mattw-mega: It's common practice to handle the responses in a global listener by switching on the request type (instead of passing a listener to that specific request).  You'll see all (or at least most) of the other commands contain a comment like "The associated request type with this request is MegaRequest::XYZ" followed by a description of the relevant getters.

src/commands.cpp
mattw-mega: Well, it's the pattern everyone is used to, I guess

cb-mega: Will fix

src/megaapi_impl.cpp
cb-mega: Reverted the changes to this class

include/megaapi.h
```diff
@@ -2125,13 +2125,93 @@ class MegaStringList
...
+
+    /**
+     * @brief Returns the string list at position i
+     *
+     * The table retains the ownership of the returned string list. It will be only valid until
+     * the table is deleted.
+     *
+     * The returned pointer is null if i is out of range.
+     *
+     * @return Number of string lists in the table
```
mattw-mega: return parameter description could be improved?  thanks

PR #1461, opened 2019-05-17
Fix json parser to process `ub` actionpacket

jgandres:
`ub` actionpacket has no payload so it breaks json parser. In order to
fix this issue we need to set pos pointer one position back to be able
to close json object properly.

### DIFF
--- a/src/megaclient.cpp
+++ b/src/megaclient.cpp
@@ -4177,6 +4177,15 @@ bool MegaClient::procsc()
                                 // last acknowledged
                                 sc_la();
                                 break;
+
+                            case MAKENAMEID2('u', 'b'):
+                                // business account update
+                                // `ub` AP has no payload, so we need to set pos
+                                // one character back to avoid to consume `}` character
+                                // and close object properly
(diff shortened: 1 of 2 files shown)


src/megaclient.cpp
```diff
@@ -4177,6 +4177,15 @@ bool MegaClient::procsc()
...
                                 sc_la();
                                 break;
+
+                            case MAKENAMEID2('u', 'b'):
+                                // business account update
+                                // `ub` AP has no payload, so we need to set pos
+                                // one character back to avoid to consume `}` character
+                                // and close object properly
+                                sc_ub();
+                                jsonsc.pos--;
```
mattw-mega: It looks like we would have called `getnameid()` on 'ub' which doesn't consume the '}', how are we iterating past that one?  It's pretty unusual to adjust the `pos` directly, I think it shouldn't be necessary? thanks

jgandres:
Closed and replaced by https://github.com/meganz/sdk/pull/1463

PR #1463, opened 2019-05-20
#12214. Business accounts maintenance

jgandres:
Parse `ub` actionpacket and notify apps the change with an event.


src/megaclient.cpp
```diff
@@ -6152,6 +6157,37 @@ void MegaClient::sc_la()
...
+    for (;;)
+    {
+        switch (jsonsc.getnameid())
+        {
+            case 's':
+                status = (int)jsonsc.getint();
+                break;
+
+            case EOO:
+                if (status == 0)
```
sergiohs84: In this case, we should skip the notification, IMO. The `0` value is not even expected by the apps.

sergiohs84: Actually, if the master account goes back to a non-business plan, the status could be `0` (not yet implemented, but API is working on it). So perhaps allow (and add to doc) that status and use a different invalid value here. Also, remember to update the `MegaClient::businessStatus` accordingly.

src/megaclient.cpp
```diff
@@ -6152,6 +6157,43 @@ void MegaClient::sc_la()
...
+            case 's':
+                statusRecv = true;
+                status = (int)jsonsc.getint();
+                break;
+
+            case EOO:
+                if (!statusRecv)
+                {
+                    std::string err = "Missing status in `ub` action packet";
+                    LOG_warn << err;
```
sergiohs84: This is clearly an error, not a simple warning. It would cause issues and un-expected behavior, so better to state it as Error in the log message.
```suggestion
                    LOG_err << err;
```

src/megaclient.cpp
```diff
@@ -6152,6 +6157,43 @@ void MegaClient::sc_la()
...
+    {
+        switch (jsonsc.getnameid())
+        {
+            case 's':
+                statusRecv = true;
+                status = (int)jsonsc.getint();
+                break;
+
+            case EOO:
+                if (!statusRecv)
```
```suggestion
                if (!statusRecv || (status < -1 || status > 2))
```

src/megaclient.cpp
```diff
@@ -6152,6 +6157,43 @@ void MegaClient::sc_la()
...
+                {
+                    std::string err = "Missing status in `ub` action packet";
+                    LOG_warn << err;
+                    sendevent(99449, err.c_str(), 0);
+                    return;
+                }
+
+                businessStatus = status;
```
sergiohs84: What about the other state's flags? Please, make them consistent.
```suggestion
                businessStatus = status;
                business = bool(status);
                if (!status)
                {
                    businessMaster = false;
                }
```

src/megaclient.cpp
```diff
@@ -6152,6 +6157,43 @@ void MegaClient::sc_la()
...
+        {
+            case 's':
+                statusRecv = true;
+                status = (int)jsonsc.getint();
+                break;
+
+            case EOO:
+                if (!statusRecv)
+                {
+                    std::string err = "Missing status in `ub` action packet";
```
```suggestion
                    std::string err = "Missing or invalid status in `ub` action packet";
```

sergiohs84:
Please, @mattw-mega , check the code for failed `g`/`p` commands returning `API_EBUSINESSPASTDUE` paying attention to the failed transfer. If it's correct, please, approve (do not merge, I'll do). Thanks

sergiohs84:
> The code looks fine, however it will result in many `ug` requests sent to the server, as we request it on every command failure reply when the system is in that state. Checking with the API team, they say the API could keep sending that error indefinitely, and eventually stop when the business pays their account, at which time a `ub` actionpacket would be sent. So in fact, since `ub` has been changed to deliver that data, it should be sufficient to just listen to that since that one already updates the `business, businessMaster, businessState` flags. I don't think we need to send the `ug` in response to that error anymore ?

Thanks for the review, @mattw-mega . I agree with you, if API will send `ub` for updates, we don't really need to trigger the `ug` upon failure. The update should be received by the client sooner or later.

Reg prevent syncs/transfers... yes, I think MEGAsync should disable syncs. And all clients should prevent any transfer in the UI. We will comunicate that to mobile apps ;) Please, take care of MEGAsync

src/megaapi_impl.cpp
```diff
@@ -13921,6 +13928,14 @@ void MegaApiImpl::whyamiblocked_result(int code)
         {
             reason = "Your account has been suspended due to multiple breaches of Mega's Terms of Service. Please check your email inbox.";
         }
+        else if (code == 400)
+        {
+            reason = "Your account has been disabled by your administrator. You may contact your business account administrators for further details.";
```
```suggestion
            reason = "Your account has been disabled by your administrator. You may contact your business account administrator for further details.";
```

src/megaapi_impl.cpp
```diff
@@ -13921,6 +13928,14 @@ void MegaApiImpl::whyamiblocked_result(int code)
...
+        }
+        else if (code == 401)
+        {
+            reason = "Your account has been removed by your administrator. You may contact your business account administrators for further details.";
```
```suggestion
            reason = "Your account has been removed by your administrator. You may contact your business account administrator for further details.";
```

PR #1469, opened 2019-05-24
WIP: Prevent log messages too long

sergiohs84:
The `cs Received:` are already truncated in the middle of the JSON when posted as log message (for efficiency, to not overload the logger that may do I/O). This commit aims to achieve the same for the `cs Sending:`.

We have detected crashes upon trying to import a huge number of nodes from a folder link, resulting in a huge _putnodes_ that produces a too large log --> crash due to OOM in iOS.

### DIFF
--- a/src/posix/net.cpp
+++ b/src/posix/net.cpp
@@ -1251,7 +1251,14 @@ void CurlHttpIO::send_request(CurlHttpContext* httpctx)
     }
     else
     {
-        LOG_debug << httpctx->req->logname << "Sending: " << *req->out;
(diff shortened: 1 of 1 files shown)


sergiohs84:
> Looks like a good idea - just a couple of technicalities - in.size() is being used in a couple of places, shouldn't it be out->size()? Also no need to call c_str() when streaming, the string can be streamed directly. thanks!

You're right, Matt. I've copy&pasted from the other location and forgot to adapt some pointers `in` --> `out`. I've amended the previous commit. Thanks

PR #1470, opened 2019-05-27
Thread safety when multiple threads call getMimeType when it hasn't b…

mattw-mega:
…een called before.


tests/tests.cpp
```diff
@@ -56,6 +61,27 @@ size_t checksize(size_t& n, size_t added)
     return n;
 }
 
+TEST(MegaApi, getMimeType)
+{
+
+    vector<unique_ptr<thread>> threads;
+
+    atomic<int> successCount = 0;
```
cb-mega: Doesn't build with GCC, just do `successCount{0}`

tests/tests.cpp
```diff
@@ -56,6 +61,27 @@ size_t checksize(size_t& n, size_t added)
...
+
+    for (int i = 0; i < 20; ++i)
+    {
+        threads.emplace_back(unique_ptr<thread>(new thread([&successCount]() {
```
cb-mega: I find that proving that something's thread-safe is really hard and takes too much effort. Even if this test passes it doesn't mean the function is thread-safe. I would simply test the function in the single-threaded case and ensure all branches are covered. I think these cases are missing:
- input is a nullptr
- the input begins with '.'

src/megaapi.cpp
```diff
@@ -3937,576 +3937,574 @@ char *MegaApi::getMimeType(const char *extension)
...
 
     string key = extension;
-    std::transform(key.begin(), key.end(), key.begin(), ::tolower);
-    map<string, string>::iterator it = mimeMap->find(key);
-    if (it == mimeMap->end())
+    std::transform(key.begin(), key.end(), key.begin(), [](char c) { return static_cast<char>(::tolower(c)); });
+    map<string, string>::iterator it = mimeMap.find(key);
+    if (it == mimeMap.end())
     {
         return NULL;
```
cb-mega: return `nullptr`?

src/megaapi.cpp
```diff
@@ -3937,576 +3937,574 @@ char *MegaApi::getMimeType(const char *extension)
...
-        (*mimeMap)["xsl"]="text/xml";
-        (*mimeMap)["xslt"]="text/xml";
-        (*mimeMap)["xsn"]="application/octet-stream";
-        (*mimeMap)["xss"]="application/xml";
-        (*mimeMap)["xtp"]="application/octet-stream";
-        (*mimeMap)["xwd"]="image/x-xwindowdump";
-        (*mimeMap)["z"]="application/x-compress";
-        (*mimeMap)["zip"]="application/x-zip-compressed";
-    }
+    static map<string, string> mimeMap{
```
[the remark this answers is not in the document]
cb-mega: How do you feel about turning this into `map<const char*, const char*>` with a custom comparator? That way we don't construct any string objects. All strings come from local storage.

src/megaapi.cpp
```diff
@@ -3937,576 +3937,574 @@ char *MegaApi::getMimeType(const char *extension)
...
+        {"xsd", "text/xml"},
+        {"xsf", "text/xml"},
+        {"xsl", "text/xml"},
+        {"xslt", "text/xml"},
+        {"xsn", "application/octet-stream"},
+        {"xss", "application/xml"},
+        {"xtp", "application/octet-stream"},
+        {"xwd", "image/x-xwindowdump"},
+        {"z", "application/x-compress"},
+        {"zip", "application/x-zip-compressed"}
```
cb-mega: I would normally add a comma here to make changing the map a bit easier in the future...

tests/tests.cpp
```diff
@@ -56,6 +61,27 @@ size_t checksize(size_t& n, size_t added)
...
+            if (::mega::MegaApi::getMimeType("nosuch") == nullptr) ++successCount;
+            if (::mega::MegaApi::getMimeType("323") == string("text/h323")) ++successCount;
+            if (::mega::MegaApi::getMimeType("zip") == string("application/x-zip-compressed")) ++successCount;
+        })));
+    }
+
+    for (auto& a : threads) a->join();
+
+    ASSERT_EQ(successCount, 60);
```
cb-mega: Normally, the expected value goes first

tests/tests.cpp
```diff
@@ -56,6 +61,27 @@ size_t checksize(size_t& n, size_t added)
...
+    for (int i = 0; i < 20; ++i)
+    {
+        threads.emplace_back(unique_ptr<thread>(new thread([&successCount]() {
+            if (::mega::MegaApi::getMimeType("nosuch") == nullptr) ++successCount;
+            if (::mega::MegaApi::getMimeType("323") == string("text/h323")) ++successCount;
+            if (::mega::MegaApi::getMimeType("zip") == string("application/x-zip-compressed")) ++successCount;
+        })));
+    }
+
+    for (auto& a : threads) a->join();
```
cb-mega: To prevent bugs, I'd prefer if we always used bodies no matter how long the statement

cb-mega: Also, `a` deserves a better name ;)

src/megaapi.cpp
```diff
@@ -3937,576 +3937,574 @@ char *MegaApi::getMimeType(const char *extension)
...
-        (*mimeMap)["xslt"]="text/xml";
-        (*mimeMap)["xsn"]="application/octet-stream";
-        (*mimeMap)["xss"]="application/xml";
-        (*mimeMap)["xtp"]="application/octet-stream";
-        (*mimeMap)["xwd"]="image/x-xwindowdump";
-        (*mimeMap)["z"]="application/x-compress";
-        (*mimeMap)["zip"]="application/x-zip-compressed";
-    }
+    static map<string, string> mimeMap{
+        // from c++11, this sort of static local initialization is one-time and thread safe
```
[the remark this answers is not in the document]
mattw-mega: cool, this comment is meant to be specific to the usage here, so our peers who see it will know it's ok, and where to look it up if needed (like me :-) )

src/utils.cpp
```diff
@@ -1499,7 +1499,7 @@ m_time_t m_mktime_UTC(const struct tm *src)
     t = mktime(&dst);
     TIME_ZONE_INFORMATION TimeZoneInfo;
     GetTimeZoneInformation(&TimeZoneInfo);
-    t += timezone.Bias * 60 - dst.tm_isdst * 3600;
+    t += TimeZoneInfo.Bias * 60 - dst.tm_isdst * 3600;
```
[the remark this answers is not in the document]
mattw-mega: It didn't - I think it was a new function added from a different platform

tests/tests.cpp
```diff
@@ -56,6 +61,27 @@ size_t checksize(size_t& n, size_t added)
     return n;
 }
 
+TEST(MegaApi, getMimeType)
+{
+
+    vector<unique_ptr<thread>> threads;
```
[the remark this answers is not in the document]
mattw-mega: true!  nice spotting, thanks

tests/tests.cpp
mattw-mega: done, thanks

tests/tests.cpp
mattw-mega: I've added those two now, thanks.   I think it's important to test the threaded case, since the whole change's purpose is to fix the thread safety, otherwise we may as well not make the change at all.

src/megaapi.cpp
mattw-mega: That part worked fine, so I left it alone - but since it's come up in review I've switched it now, and made the returns much more compact also.

src/megaapi.cpp
mattw-mega: cool, maybe next time - do all the compilers support that?

tests/tests.cpp
mattw-mega: I think that is done for == comparisons when there's a risk of the programmer accidentally using = ?   I find naming the main tested thing first is closer to natural language.

tests/tests.cpp
mattw-mega: that's a lot of extra lines for very little code though, and it's only a unit test  - I could keep it compact with for_each and a lambda, but I think this formulation is much easier to read?  thanks

Renamed to 't' for thread, though its scope is only a single line so we don't need to remember its purpose for long :-)

src/megaapi.cpp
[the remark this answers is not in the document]
mattw-mega: sounds good - for a future update :-)  I think we spent enough time on this one, and it was strings before with no complaints.  thanks

PR #1474, opened 2019-05-31
CUv2 - Update iOS binding according to SDK changes

simonwng:
This pr is to update iOS binding project according to SDK changes

### DIFF
--- a/bindings/ios/MEGABackgroundMediaUpload.h
+++ b/bindings/ios/MEGABackgroundMediaUpload.h
@@ -28,7 +28,7 @@ NS_ASSUME_NONNULL_BEGIN
 @interface MEGABackgroundMediaUpload : NSObject
 
 /**
- * @brief Initial step to upload a photo/video via iOS low-power background upload feature
+ * @brief Initial step to upload a photo/video via iOS low-power background upload feature.
  *
  * Creates an object which can be used to encrypt a media file, and upload it outside of the SDK,
  * eg. in order to take advantage of a particular platform's low power background upload functionality.
@@ -55,7 +55,7 @@ NS_ASSUME_NONNULL_BEGIN
 - (BOOL)analyseMediaInfoForFileAtPath:(NSString *)inputFilepath;
 
 /**
- * @brief Encrypt the file or a portion of it
+ * @brief Encrypt the file or a portion of it.
  *
  * Call this function once with the file to be uploaded. It uses mediainfo to extract information that will
  * help the webclient show or play the file in various browsers. The information is stored in this object
@@ -71,7 +71,7 @@ NS_ASSUME_NONNULL_BEGIN
  * so that RAM usage is not excessive.
  *
  * @param inputFilePath The file to encrypt a portion of (and the one that is ultimately being uploaded).
- * @param start The index of the first byte of the file to encrypt
+ * @param start The index of the first byte of the file to encrypt.
  * @param length The number of bytes of the file to encrypt. The function will round this value up by up to 1MB to fit the
  *        MEGA internal chunking algorithm. The number of bytes actually encrypted and stored in the new file is the updated number.
  *        You can supply -1 as input to request the remainder file (from start) be encrypted.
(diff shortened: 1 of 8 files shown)


bindings/ios/MEGASdk.mm
```diff
@@ -1485,22 +1485,13 @@ - (void)setUploadLimitWithBpsLimit:(NSInteger)bpsLimit {
...
+    self.megaApi->backgroundMediaUploadRequestUploadURL(filesize, mediaUpload.getCPtr, [self createDelegateMEGARequestListener:delegate singleListener:YES]);
 }
 
-- (BOOL)completeBackgroundMediaUpload:(MEGABackgroundMediaUpload *)mediaUpload fileName:(NSString *)fileName parentNode:(MEGANode *)parentNode fingerprint:(NSString *)fingerprint originalFingerprint:(NSString *)originalFingerprint binaryUploadToken:(NSData *)token delegate:(id<MEGARequestDelegate>)delegate {
-    std::string binaryToken = std::string((const char *)token.bytes, token.length);
-    return self.megaApi->backgroundMediaUploadComplete(mediaUpload.getCPtr, fileName.UTF8String, parentNode.getCPtr, fingerprint.UTF8String, originalFingerprint.UTF8String, &binaryToken, [self createDelegateMEGARequestListener:delegate singleListener:YES]);
+- (void)completeBackgroundMediaUpload:(MEGABackgroundMediaUpload *)mediaUpload fileName:(NSString *)fileName parentNode:(MEGANode *)parentNode fingerprint:(NSString *)fingerprint originalFingerprint:(NSString *)originalFingerprint binaryUploadToken:(NSData *)token delegate:(id<MEGARequestDelegate>)delegate {
```
jnavarrom: Can we pass the thumbnails & preview to the `backgroundMediaUploadComplete`?
This way, the `putnode` will be done with the file attributes, and avoid at the app layer setting up the thumbnail and the preview after the `putnode`, right?

simonwng: Good question, since we have the discussion in Slack, let's move the discussion to Slack.

I have changed code according to new SDK changes. Now, we don't need to specify thumbnail & preview handles in put node.

bindings/ios/MEGABackgroundMediaUpload.h
```diff
@@ -91,6 +91,20 @@ NS_ASSUME_NONNULL_BEGIN
...
+ * The node created via completeBackgroundMediaUpload:fileName:parentNode:fingerprint:originalFingerprint:binaryUploadToken:delegate: in MEGASdk
+ * will gain these coordinates as part of the
+ * node creation. If the unshareable flag is set, the coodinates are encrypted in a way that even if the
+ * node is later shared, the GPS coordinates cannot be decrypted by a different account.
+ *
+ * @param latitude The GPS latitude
+ * @param latitude The GPS longitude
+ * @param unshareable Set this true to prevent the coordinates being readable by other accounts.
+ */
+- (void)setCoordinatesWithLatitude:(NSNumber *)latitude longitude:(NSNumber *)longitude isUnshareable:(BOOL)unshareable;
```
jnavarrom: Maybe it is better use double directly as parameters. What do you think?
(I know there is a lot of places where we use NSNumbers instead primitive types)

jnavarrom: This way, in the app layer code, we don't need create a `NSNumber` from a double, that finally will be convert to a double when pass it to the SDK

simonwng: In `MEGASdk` we need to use `NSNumber` to add ability to remove coordinates from nodes as described in MegaApi.
>To remove the existing coordinates, set both the latitude and longitude to the value MegaNode::INVALID_COORDINATE.

I think it would be good to keep consistent? Especially if you want to remove coordinates later on.

jnavarrom: You meant keep consistency between: `setNodeCoordinates` and `setCoordinatesWithLatitude` ?

I would use `double` in both cases as parameters.
If we would want remove the coordinates we can use the `MegaNode::INVALID_COORDINATE` as parameter.

simonwng: Yes, I was to keep consistency with `setNodeCoordinates` and `setCoordinatesWithLatitude`. But here is a different case where we don't even have a node. So, agree with you. I have changed to `double`.

PR #1481, opened 2019-06-06
#12280 Improve retrieval of info about folder links

jgandres:
The information retrieval of a folder link that contains a lot of files/folders, it's very slow process with a fetchnodes.
We need to add support to `pli` command in order to retrieve the information faster.

### DIFF
--- a/include/megaapi.h
+++ b/include/megaapi.h
@@ -2595,7 +2595,7 @@ class MegaRequest
             TYPE_ADD_BACKUP, TYPE_REMOVE_BACKUP, TYPE_TIMER, TYPE_ABORT_CURRENT_BACKUP,
             TYPE_GET_PSA, TYPE_FETCH_TIMEZONE, TYPE_USERALERT_ACKNOWLEDGE,
             TYPE_CHAT_LINK_HANDLE, TYPE_CHAT_LINK_URL, TYPE_SET_PRIVATE_MODE, TYPE_AUTOJOIN_PUBLIC_CHAT,
-            TYPE_CATCHUP,
+            TYPE_CATCHUP, TYPE_PUBLIC_LINK_INFORMATION,
             TOTAL_OF_REQUEST_TYPES
         };
 
@@ -14304,6 +14304,32 @@ class MegaApi
          * @param listener MegaRequestListener to track this request
          */
         void catchup(MegaRequestListener *listener = NULL);
+
+        /**
+         * @brief Retrieve basic information about a folder link
+         *
+         * This function retrieves basic information from a folder link, like the number of files / folders
+         * and the name of the folder. For folder links containing a lot of files/folders,
+         * this function is more efficient than a fetchnodes.
+         *
+         * Valid data in the MegaRequest object received on all callbacks:
+         * - MegaRequest::getNodeHandle - Returns the public handle of the folder link
+         *
+         * Valid data in the MegaRequest object received in onRequestFinish when the error code
+         * is MegaError::API_OK:
+         * - MegaRequest::getNodeHandle() - Returns the public handle of the folder
+         * - MegaRequest::getMegaFolderInfo() - Returns information about the contents of the folder
+         * - MegaRequest::getParentHandle() - Returns the handle of the owner of the folder
+         * - MegaRequest::getText() - Returns the name of the folder
+         *
+         * On the onRequestFinish error, the error code associated to the MegaError can be:
+         * - MegaError::API_EARGS  - If the public handle not corresponds to a valid folder link
+         * - MegaError::API_ENOENT - If the public handle is not valid.
+         *
+         * @param ph MegaHandle that represents the public handle of the folder
+         * @param listener MegaRequestListener to track this request
+         */
+        void getPublicLinkInformation(MegaHandle ph, MegaRequestListener *listener = NULL);
 
 private:
         MegaApiImpl *pImpl;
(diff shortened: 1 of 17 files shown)


src/megaapi_impl.cpp
```diff
@@ -11578,6 +11587,65 @@ void MegaApiImpl::chatlinkjoin_result(error e)
...
+        handle h;
+        size_t t = 0;
+        string nodekey;
+        string attrs;
+        byte* buf = NULL;
+        const char* k = NULL;
+        unsigned int keylength = FOLDERNODEKEYLENGTH;
+        byte binkey[FILENODEKEYLENGTH];
+        SymmCipher cipher;
+        SymmCipher* sc = &client->key;
```
sergiohs84: This is the master key of the logged in account. It cannot be used to decrypt the received key. Instead, you need to use the key part of the folder link. Note you're calling `folderaccess(link)`, which sets the `client->key` to the folder link, but it will break any de/ciphering requiring the account's masterkey.
A possible workaround is to modify `folderaccess()` to receive a `SymmCipher` as parameter and store the folder key in there, rather than in `Client::key`.

src/megaapi_impl.cpp
```diff
@@ -11578,6 +11587,65 @@ void MegaApiImpl::chatlinkjoin_result(error e)
...
+
+        if (key->size())
```
sergiohs84: An empty key should be handled at the `procresult()` of the command

src/megaapi_impl.cpp
```diff
@@ -20051,6 +20119,23 @@ void MegaApiImpl::sendPendingRequests()
...
+        case MegaRequest::TYPE_PUBLIC_LINK_INFORMATION:
+        {
+            const char *link = request->getLink();
+            if (!link)
+            {
+                e = API_EARGS;
+                break;
+            }
+
+            e = client->folderaccess(link);
```
sergiohs84: This is a `NO NO NO`, @jgandres . If you check the code in that function, it will set the master key to the key of the folder link, and also the rootnode to the public handle, plus the authentication token to the public handle for API requests.

src/megaapi_impl.cpp
```diff
@@ -11578,6 +11587,65 @@ void MegaApiImpl::chatlinkjoin_result(error e)
...
+        else
+        {
+            LOG_warn << "The folder link information doesn't contains the decryption key";
+            return fireOnRequestFinish(request, API_EINTERNAL);
+        }
+
+        MegaFolderInfoPrivate *folderInfo = new MegaFolderInfoPrivate(numFiles, numFolders - 1, numVersions, currentSize, versionsSize);
+        request->setNodeHandle(ph);
+        request->setParentHandle(owner);
+        request->setText((const char*)buf);
```
sergiohs84: Documentation says the `MegaRequest::getText()` will return the name of the folder, but here you don't get rid of the rest of attributes nor the prefix of each.

include/megaapi_impl.h
```diff
@@ -2734,6 +2735,7 @@ class MegaApiImpl : public MegaApp
         virtual void chatlinkurl_result(handle, int, string*, string*, int, m_time_t, error);
         virtual void chatlinkclose_result(error);
         virtual void chatlinkjoin_result(error);
+        virtual void folderlinkinfo_result(error, handle , handle, string*, string* , m_off_t, uint32_t , uint32_t , m_off_t , uint32_t);
```
sergiohs84: This method is not specific of chat, despite it's first usage will be for chats. It shouldn't be inside the `ENABLE_CHAT` ifdefs

src/megaapi_impl.cpp
```diff
@@ -11578,6 +11587,65 @@ void MegaApiImpl::chatlinkjoin_result(error e)
...
+        {
+            // Discard user handle for received key
+            while ((t = key->find_first_of(':', t)) != string::npos)
+            {
+               k = key->c_str() + (++t);
+               break;
+            }
+
+            // Decrypt the received key with our folder key
+            if (client->decryptkey(k, binkey, keylength, sc, 0, ph))
```
sergiohs84: If you pass the public handle of the node, this function may trigger the rewrite-key procedure, which you do not want to happen!! Pass `UNDEF` to avoid it.

src/megaapi_impl.cpp
```diff
@@ -11578,6 +11587,65 @@ void MegaApiImpl::chatlinkjoin_result(error e)
...
+    if (e == API_OK)
+    {
+        handle h;
+        size_t t = 0;
+        string nodekey;
+        string attrs;
+        byte* buf = NULL;
+        const char* k = NULL;
+        unsigned int keylength = FOLDERNODEKEYLENGTH;
+        byte binkey[FILENODEKEYLENGTH];
```
sergiohs84: Isn't it a folder key? Then, the length of this array is wrong

include/megaapi.h
```diff
@@ -14305,6 +14305,32 @@ class MegaApi
...
+         * Valid data in the MegaRequest object received in onRequestFinish when the error code
+         * is MegaError::API_OK:
+         * - MegaRequest::getMegaFolderInfo() - Returns information about the contents of the folder
+         * - MegaRequest::getNodeHandle() - Returns the public handle of the folder
+         * - MegaRequest::getParentHandle() - Returns the handle of the owner of the folder
+         * - MegaRequest::getText() - Returns the name of the folder
+         *
+         * On the onRequestFinish error, the error code associated to the MegaError can be:
+         * - MegaError::API_EARGS  - If the link is not a valid folder link
+         * - MegaError::API_EINCOMPLETE - If the public link does not contain the key
```
sergiohs84: Update possible error codes and their reason.

sergiohs84:
trigger compilation

PR #1482, opened 2019-06-06
Allow to cancel account while logged in any account

sergiohs84:
The confirmation of email's changes is not allowed to be done by a different account than the one than requested the change. But the account's cancellation is allowed.


sergiohs84:
Rejected. The issue has been solved by the API team to return the correct userhandle for all type of links. Taking into account these links expire in 12 hours, there's no point to add this patch to the SDK.

PR #1483, opened 2019-06-13
address warnings

mattw-mega:
lossy 32/64 bit or int/char conversions
signed/unsigned comparisons
use of variables that the compiler thinks may not be initialized
variables of the same name eclipsing others

### DIFF
--- a/src/megaclient.cpp
+++ b/src/megaclient.cpp
@@ -668,7 +668,7 @@ void MegaClient::getprivatekey(const char *code)
     reqs.add(new CommandGetPrivateKey(this, code));
 }
 
-void MegaClient::confirmrecoverylink(const char *code, const char *email, const char *password, const byte *masterkey, int accountversion)
+void MegaClient::confirmrecoverylink(const char *code, const char *email, const char *password, const byte *masterkeyptr, int accountversion)
 {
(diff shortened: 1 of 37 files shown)


src/megaclient.cpp
```diff
@@ -4887,7 +4887,8 @@ bool MegaClient::sc_shares()
                 }
 
                 // ignore unrelated share packets (should never be triggered)
-                if (!ISUNDEF(oh) && !(outbound = (oh == me)) && (uh != me))
+                outbound = (oh == me);
+                if (!ISUNDEF(oh) && !(outbound) && (uh != me))
```
sergiohs84: Brackets around `outbound` not needed ;P

src/megaclient.cpp
```diff
@@ -5020,7 +5021,7 @@ void MegaClient::sc_paymentreminder()
         case EOO:
             if (statecurrent)
             {
-                useralerts.add(new UserAlert::PaymentReminder(expiryts, useralerts.nextId()));
+                useralerts.add(new UserAlert::PaymentReminder(expiryts ? expiryts : m_time(), useralerts.nextId()));
```
sergiohs84: Good catch!

PR #1485, opened 2019-06-13
escape/unescape ".." & "."

polmr:


polmr:
Before merging we might want to attend to this use case:
`/local/path` <- synched with -> `/remote/path`

polmr:
> Before merging we might want to attend to this use case:
> `/local/path` <- synched with -> `/remote/path`
> file `XXX` renamed to `.` in /remote/path
> gets renamed back to `XXX` in MEGA.
> 
> Still, this seems to be happening regardless of the changes of this PR

This seems to be because `..` is excluded by default. The same reproduces is `validfilename` is renamed as `excludedfilename` (given `excludedfilename` is in the exclusion list)

PR #1492, opened 2019-06-16
Factor calculation of progress from chunkmap

mattw-mega:
Two nearly identical pieces of code, for figuring out our current upload/download progress
from the chunkmap records - just one extra line in one - can easily be a function.

### DIFF
--- a/src/transferslot.cpp
+++ b/src/transferslot.cpp
@@ -358,7 +358,7 @@ void chunkmac_map::calcprogress(m_off_t size, m_off_t& chunkpos, m_off_t& progre
     chunkpos = 0;
     progresscompleted = 0;
 
-    for (chunkmac_map::iterator it = begin(); it != end(); it++)
+    for (chunkmac_map::iterator it = begin(); it != end(); ++it)
     {
         m_off_t chunkceil = ChunkedHash::chunkceil(it->first, size);
 
@@ -377,7 +377,7 @@ void chunkmac_map::calcprogress(m_off_t size, m_off_t& chunkpos, m_off_t& progre
             progresscompleted += it->second.offset;
             if (lastblockprogress)
             {
-                *lastblockprogress = it->second.offset;
+                *lastblockprogress += it->second.offset;
             }
         }
     }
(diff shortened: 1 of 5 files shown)


src/transferslot.cpp
```diff
@@ -353,6 +353,36 @@ bool chunkmac_map::unserialize(const char*& ptr, const char* end)
...
+        {
+            m_off_t chunksize = chunkceil - ChunkedHash::chunkfloor(it->first);
+            progresscompleted += chunksize;
+        }
+        else
+        {
+            progresscompleted += it->second.offset;
+            if (lastblockprogress)
+            {
+                *lastblockprogress = it->second.offset;
```
cb-mega: I think this should be `+=`

src/transferslot.cpp
```diff
@@ -353,6 +353,36 @@ bool chunkmac_map::unserialize(const char*& ptr, const char* end)
     return true;
 }
 
+void chunkmac_map::calcprogress(m_off_t size, m_off_t& chunkpos, m_off_t& progresscompleted, m_off_t* lastblockprogress)
+{
+    chunkpos = 0;
```
cb-mega: I don't see where these two variables are set to 0 in transfer.cpp? Did the original code do that?

src/transferslot.cpp
```diff
@@ -353,6 +353,36 @@ bool chunkmac_map::unserialize(const char*& ptr, const char* end)
...
+    progresscompleted = 0;
+
+    for (chunkmac_map::iterator it = begin(); it != end(); it++)
```
cb-mega: I'd change this to `++it`

src/transferslot.cpp
mattw-mega: good point, updated - so it's exactly the same

src/transferslot.cpp
mattw-mega: In transfer.cpp, those are passed from the `t = new Transfer` object, and have been initialised to 0 in its constructor.  good to check though, thanks!

src/transferslot.cpp
mattw-mega: done :-)

src/transferslot.cpp
cb-mega: Isn't that a weird side-effect of the function to set these variables to 0 if the map is empty for instance?

cb-mega: Maybe those should be set to 0 by the caller?

PR #1493, opened 2019-06-17
update windows megafs.h changed function signature for override

mattw-mega:
getextension() had changed a parameter from int to size_t

### DIFF
--- a/src/win32/fs.cpp
+++ b/src/win32/fs.cpp
@@ -1081,7 +1081,7 @@ bool WinFileSystemAccess::getextension(string* filename, char* extension, size_t
         - (filename->size() & 1));   // if the string has had an extra null char added for surety, get back on wchar_t boundary.
 
     char c;
-    int i, j;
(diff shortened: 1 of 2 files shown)


src/win32/fs.cpp
```diff
@@ -1081,7 +1081,7 @@ bool WinFileSystemAccess::getextension(string* filename, char* extension, size_t
         - (filename->size() & 1));   // if the string has had an extra null char added for surety, get back on wchar_t boundary.
 
     char c;
-    int i, j;
+    size_t i, j;
```
cb-mega: I'd move these declarations to the for-loops (since you're changing this function anyway...)

mattw-mega: I thought that too, but then saw that `j` is used after the `j` loop, so it has to be declared early anyway, so I'm not sure there would be much gain - leaving it as it keeps it internally consistent with all vairables pre-declared.

PR #1494, opened 2019-06-17
#12412 SMS verification

cb-mega:
This implements:
- contact verification via sms
- retrieve registered contacts given a set of client contacts
- retrieve country calling codes


bindings/ios/MEGASdk.h
```diff
@@ -161,6 +161,19 @@ typedef NS_ENUM(NSUInteger, StorageState) {
...
+typedef NS_ENUM(NSInteger, SMSState) {
+    SMSStateNotAllowed = 0,
+    SMSStateOnlyUnblock = 1,
+    SMSStateOptInAndUnblock = 2,
+};
+
+typedef NS_ENUM(NSInteger, AccountSuspensionType) {
+    AccountSuspensionTypeNone = 0, // The account is not blocked
+    AccountSuspensionTypeNonCopyright = 200, // suspension for any type of suspension, but copyright suspension
+    AccountSuspensionTypeCopyright = 300, // suspension only for multiple copyright violations
```
sergiohs84: Very likely, the PR https://github.com/meganz/sdk/pull/1463 will get merged sooner than this one. Perhaps you want to add the new reasons to the enumeration at that time: https://github.com/meganz/sdk/pull/1463/files#diff-c3de3d92d9d9472cc3d715cb978d2940R13851

cb-mega: Not sure I understand. How about just resolve the conflicts when they occur?

include/megaapi.h
```diff
@@ -14716,10 +14862,79 @@ class MegaApi
...
+         */
+        void checkSMSVerificationCode(const char* verificationCode, MegaRequestListener *listener = NULL);
+
+        /**
+         * @brief Requests the contacts that are registered at MEGA (currently verified through SMS)
+         *
+         * The request will return any of the provided contacts that are registered at MEGA, i.e.,
+         * are verified through SMS (currently).
+         *
+         * contacts is a MegaStringMap from 'user detail' to the user's name. For instance:
```
sergiohs84: Document as the standard:
```
         * The associated request type with this request is MegaRequest::TYPE_GET_REGISTERED_CONTACTS
         * Valid data in the MegaRequest object received on callbacks:
         * - MegaRequest::getMegaStringMap - Returns...
```
I'll do this one as an example for reference

include/megaapi.h
```diff
@@ -14716,10 +14862,79 @@ class MegaApi
...
+         *
+         * @param contacts The map of contacts to get registered contacts from
+         * @param listener MegaRequestListener to track this request
+         */
+        void getRegisteredContacts(const MegaStringMap* contacts, MegaRequestListener *listener = NULL);
+
+        /**
+         * @brief Requests the currently available country calling codes
+         *
+         * The response value is stored as a MegaStringListMap mapping from two-letter country code
```
sergiohs84: Adjust the documentation to the standard. Basically, members of the requests that are available at all callbacks, members available when the request finishes with API_OK and possible error codes with a hint for the reasons.

include/megaapi.h
```diff
@@ -7002,6 +7114,27 @@ class MegaApi
          */
         bool appleVoipPushEnabled();
 
+        /**
+         * @brief Check if the opt-in or account ublocking SMS is allowed
+         *
+         * The result indicated whether the sendSMSVerificationCode() function can be used.
```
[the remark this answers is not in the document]
sergiohs84: Fixed

include/megaapi.h
```diff
@@ -14716,10 +14862,79 @@ class MegaApi
          *
          * When onRequestFinish is called with MegaError::API_OK, the SDK is guaranteed to be
          * up to date (as for the time this function is called).
+         * @brief Send a verification code txt to the supplied phone number
```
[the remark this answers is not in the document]
sergiohs84: Fixed

include/megaapi.h
sergiohs84: Done! Please, see the example and follow it for every new public interface.

src/commands.cpp
```diff
@@ -3579,6 +3584,13 @@ void CommandGetUserData::procresult()
             }
             break;
 
+        case MAKENAMEID4('s', 'm', 's', 'v'):
+            if (!client->json.storeobject(&smsv))
+            {
+                return client->app->userdata_result(NULL, NULL, NULL, API_EINTERNAL);
```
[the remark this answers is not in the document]
sergiohs84: Fixed

src/megaclient.cpp
```diff
@@ -3565,6 +3566,8 @@ void MegaClient::locallogout()
     ssrs_enabled = false;
     nsr_enabled = false;
     aplvp_enabled = false;
+    smsve_state = -1;
+    sms_verifiedphone.clear();
```
sergiohs84: Always remember to clear any new variable related to the state of the account upon logout. In this case, the verified phone was kept ;)

bindings/qt/sdk.pri
```diff
@@ -58,9 +58,8 @@ CONFIG(USE_MEGAAPI) {
   }
 }
 
-CONFIG += c++11
 !win32 {
-    QMAKE_CXXFLAGS += -std=c++11
+    QMAKE_CXXFLAGS += -std=c++11 -Wextra -Wconversion -Wno-unused-parameter
```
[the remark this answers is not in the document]
cb-mega: Yes and we still are. On a different line I simply removed `CONFIG += c++11` which ensures we're not using gnu extensions.
For a future change: We shouldn't actually define the language standard here at all because it affects downstream projects. Like if a project using the SDK decided to use C++17 it simply can't because we're overriding it here with C++11.

include/megaapi.h
```diff
@@ -2133,13 +2133,102 @@ class MegaStringList
...
 
     /**
      * @brief Returns the number of strings in the list
      * @return Number of strings in the list
      */
-    virtual int size();
+    virtual int size() const;
+};
+
+class MegaStringListMap
```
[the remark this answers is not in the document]
cb-mega: Will add

include/megaapi.h
```diff
@@ -2150,13 +2150,102 @@ class MegaStringList
...
+    virtual void set(const char* key, const MegaStringList* value);
+
+    /**
+     * @brief Returns the number of (string, string list) pairs in the map
+     * @return Number of pairs in the map
+     */
+    virtual int size() const;
+};
+
+class MegaStringTable
```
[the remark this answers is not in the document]
cb-mega: Will add

include/megaapi.h
cb-mega: will do

src/commands.cpp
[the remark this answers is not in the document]
cb-mega: Ok thanks, but I feel it may be better for the original author to make changes. That way it'd be easier to follow and understand as to what's going on.

src/commands.cpp
```diff
@@ -7074,6 +7092,266 @@ void CommandSetLastAcknowledged::procresult()
...
+                        {
+                            LOG_err << "Failed to parse 'get registered contacts' response";
+                            app.getregisteredcontacts_result(API_EINTERNAL, nullptr);
+                            return;
+                        }
+                    }
+                }
+            }
+            json.leaveobject();
+            if (entry_user_detail.empty() || id.empty() || user_detail.empty())
```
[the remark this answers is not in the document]
cb-mega: I think failing to read one of the items should fail the whole json parsing cuz it means there's something inherently wrong and the data cannot be trusted.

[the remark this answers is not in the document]
mattw-mega: The command request and responses are in batches, so the reply from the server is often one big JSON array of individual responses, eg  [{"a":"b"},-9,[1,2,3]].  So we must leave the json pointer pointing to the next response in the overall array, even if this current one can't be parsed (often just due to a newly added field, or a value turns out to be an array instead of a string etc.  That way the next command response can be processed correctly, minimizing the possibility of a new feature breaking existing functionality.

include/mega/megaclient.h
```diff
@@ -216,6 +216,12 @@ class MEGA_API MegaClient
     // Account has VOIP push enabled (only for Apple)
     bool aplvp_enabled;
 
+    // 2 = Opt-in and unblock SMS allowed 1 = Only unblock SMS allowed 0 = No SMS allowed  -1 = flag was not received
+    int smsve_state;
```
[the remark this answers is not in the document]
mattw-mega: added :-)

include/megaapi.h
```diff
@@ -14716,10 +14862,79 @@ class MegaApi
...
+         * Make sure to test the result of smsAllowedState() before calling this function.
          *
          * @param listener MegaRequestListener to track this request
          */
-        void catchup(MegaRequestListener *listener = NULL);
+        void sendSMSVerificationCode(const char* phoneNumber, MegaRequestListener *listener = NULL, bool reverifying_whitelisted = false);
+ 
+        /**
+         * @brief Check a verification code that the user should have received via txt
+         *
```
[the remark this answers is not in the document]
mattw-mega: updated

include/megaapi.h
```diff
@@ -14716,10 +14862,79 @@ class MegaApi
...
+         * This function validates that the user received the txt sent by sendSMSVerificationCode().
+         * If the user supplied the right code, the result will be API_OK.
+         *
```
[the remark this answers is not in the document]
mattw-mega: The error code comes direct from the JSON reply, so these are server controlled, and the list of possible error may change or expand in future - if we list the current ones, it may lead 3rd party developers to think that the list will not change in future?

include/megaapi.h
```diff
@@ -14716,10 +14862,79 @@ class MegaApi
...
+         * @param listener MegaRequestListener to track this request
```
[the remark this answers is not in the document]
mattw-mega: done :-)

src/megaclient.cpp
mattw-mega: good point - perhaps we should arrange for that to happen automatically as it's an easy one to miss (for me, anyway).   Maybe a function that is called to initialize those values on both startup and on logout/reset (but also leaving aside any that really should be kept).   Or perhaps assign a default-constructed version of the MegaClient (but preserve anything that should not be overridden)

src/commands.cpp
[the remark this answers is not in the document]
cb-mega: Just had a chat with Matt about the json parsing. I will change the parsing to ensure we get to the end of the json in both success and error case. That way a failure of the current json doesn't fail the whole batch.

bindings/ios/MEGASdk.h
simonwng: Hi Sergio, the enum is created in iOS binding project to provide support to iOS code. This is to avoid magic numbers in iOS code. Yes, we need to add the new suspension types (400, 401) to the list later on. Thanks for your reminder.

include/megaapi.h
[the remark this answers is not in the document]
mattw-mega: As it turns out, these are documented in the wiki - I've added them to the header documentation.  cheers

src/megaclient.cpp
sergiohs84: I think it's fine as it is right now. The function you suggest is actually `init()`, which cleans anything related to the account's state and re-initializes the `MegaClient`. It is called during initialization and also as part for the `locallogout()` :)

bindings/qt/sdk.pri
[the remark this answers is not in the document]
sergiohs84: I disagree. The SDK uses an specific version of the C standard. As today, it is C++11, so the main `.pri` file should specify it. Everything in the SDK must be able to compile using C++11. If a sub-project like the sync-tests want to use a different version, it should only apply to that project and you can unset the `c++11` and set the `c++17`
```
CONFIG -= c++11
QMAKE_CXXFLAGS-=-std=c++11
CONFIG += c++17
QMAKE_CXXFLAGS+=-std=c++17
```

include/megaapi.h
```diff
@@ -14716,10 +14862,79 @@ class MegaApi
...
+         * The code is txted to the user.  Once the user has it, they can type it into the app
+         * and the call checkSMSVerificationCode() can be used to validate the user did receive the txt,
+         * so that really is their phone number.
+         *
+         * The frequency with which this call can be used is very limited (once or twice per
+         * 24 hour period, at the time of writing this), so it's important to get the 
+         * number right on the first try.  The result will be API_ETEMPUNAVAIL if it has
+         * been tried too frequently.
+         *
+         * Make sure to test the result of smsAllowedState() before calling this function.
```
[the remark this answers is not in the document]
sergiohs84: Not sure if `reverifying_whitelisted` should be available at the public interface. It is for internal/debugging purposes only, so perhaps we should enclose it in:
```
#ifndef NDEBUG
```
That way, official apps built in release mode will not have the chance to delete their existing verifications.

bindings/qt/sdk.pri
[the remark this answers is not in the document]
cb-mega: That seems a little cumbersome, no? IMO the build target should define the standard. Of course, the SDK will only compile with at least C++11 but also with anything newer.

include/megaapi.h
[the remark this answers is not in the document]
mattw-mega: Yes I agree, I would have preferred to keep it out of the interface and would like to remove it in future.   However the iOS and Android apps needed to be able to test with it, and I don't think they only build/install debug.   It only has an effect for test accounts that have a special flag set on the server side, so it won't cause any issues.

bindings/qt/sdk.pri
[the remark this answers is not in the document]
sergiohs84: Note that the SDK itself is a target as well for the apps, who use the `sdk.pri` for that purpose, so according to your idea it should define the standard too :)
A suggestion from @polmr is to define a flag to allow SDK's target apps to decide. Imagine the following in the `sdk.pri`:
```
!CONFIG(DO_NOT_SET_CPP_STANDARD) 
{
    CONFIG+=C++11
}
```
and in the SDK's target apps, their `.pro`s may define their desired standard before including the `sdk.pri` by:
```
CONFIG+=C++17
CONFIG+=DO_NOT_SET_CPP_STANDARD
```
That way, we have the best of both worlds. How does it sound?

Asi dejamos evolucionar a ambar partes por separado: podemos pasar a usar c++14 en el sdk.pri sin tener que tocar configuracion de las apps, y tenemos un standar usado por defecto en caso de que no este establecido por la aplicacion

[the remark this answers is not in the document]
cb-mega: Yeah that seems reasonable. The only problem with `CONFIG+=C++11` is that QMake, for whatever reason, adds gnu extensions which it really shouldn't. 
Sorry I wasn't clear what I meant by "build target". I meant an executable target should define the standard. However, configuring it the way you suggested should work too :)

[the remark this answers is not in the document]
mattw-mega: in CMAKE you can iterate on the items in the CXX_FLAGS variables, and check if any standard was set, and add -std=c++11 if none was set by the enclosing application.  Maybe the same approach is possible in qmake?   Just a thought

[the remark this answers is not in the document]
sergiohs84: I don't know if `CXX_FLAGS` can be checked, but if you feel the `DO_NOT_SET_CPP_STANDARD` makes the trick, go ahead, please, and push the change :)

sergiohs84:
Please, @cb-mega , take a look to my comments. I've addressed all of them, but for you to check I'm doing the right thing, breaking nothing, and also to be aware of some minor adjustments, such as create methods for new features at `MegaClient` level, rather than queue new commands from outside :)
I'll keep working tomorrow on the review.

bindings/ios/MEGARequest.mm
```diff
@@ -187,7 +187,7 @@ - (MEGAFolderInfo *)megaFolderInfo {
 }
 
 - (NSDictionary<NSString *, MEGAStringList *> *)megaStringListDictionary {
-    MegaStringListMap *map = self.megaRequest->getMegaStringListMap();
+    MegaStringListMap *map = self.megaRequest->getMegaStringListMap()->copy();
```
sergiohs84: Not sure if you really need to make a copy here. The `map` is not deleted here, not sure if that would represent a memory-leak?

sergiohs84: I guess what you want to do is to keep valid `MegaStringList` objects, which should be done by `copy()` on the items of the map, not copying the full map and keeping it alive forever.

cb-mega: Yes, after looking at this more closely the `keyList` should be deleted and key and value should be copied into `dict`. Then a call to `copy` isn't necessary. I will talk to Simon about this.

simonwng: Thanks for point out this issue. I have updated the iOS binding code according to Christian's suggestions.

bindings/ios/MEGASdk.mm
```diff
@@ -2160,6 +2160,34 @@ - (void)getPublicLinkInformationWithFolderLink:(NSString *)folderLink {
...
+}
+
+- (void)getRegisteredContacts:(NSArray<NSDictionary *> *)contacts delegate:(id<MEGARequestDelegate>)delegate {
+    MegaStringMap *stringMapContacts = MegaStringMap::createInstance();
+    for (NSDictionary *contact in contacts) {
+        NSString *key = contact.allKeys[0];
+        NSString *value = contact.allValues[0];
+        stringMapContacts->set(key.UTF8String, value.UTF8String);
+    }
+    self.megaApi->getRegisteredContacts(stringMapContacts, [self createDelegateMEGARequestListener:delegate singleListener:YES]);
```
sergiohs84: Another memory leak: the `MegaStringMap` is not deleted, and the `getRegisteredContacts()` doesn't take the ownership, but makes a copy. I'll fix it.

bindings/ios/MEGASdk.h
```diff
@@ -165,6 +165,21 @@ typedef NS_ENUM(NSUInteger, StorageState) {
...
+    AccountSuspensionTypeBusinessDiabled = 400, // the subuser of a business account has been disabled
```
sergiohs84: Typo:
```suggestion
    AccountSuspensionTypeBusinessDisabled = 400, // the subuser of a business account has been disabled
```

bindings/ios/MEGASdk.mm
```diff
@@ -2160,6 +2160,34 @@ - (void)getPublicLinkInformationWithFolderLink:(NSString *)folderLink {
...
+#pragma mark - SMS
+
+- (SMSState)smsAllowedState {
+    return (SMSState)self.megaApi->smsAllowedState();
+}
+
+- (void)getRegisteredContacts:(NSArray<NSDictionary *> *)contacts delegate:(id<MEGARequestDelegate>)delegate {
+    MegaStringMap *stringMapContacts = MegaStringMap::createInstance();
+    for (NSDictionary *contact in contacts) {
+        NSString *key = contact.allKeys[0];
```
simonwng: @carolzato, please use `contact.allKeys.firstObject` to make it safer

bindings/ios/MEGASdk.mm
```diff
@@ -2160,6 +2160,34 @@ - (void)getPublicLinkInformationWithFolderLink:(NSString *)folderLink {
...
+        NSString *value = contact.allValues[0];
```
simonwng: @carolzato, same here, `firstObject`

bindings/ios/MEGASdk.mm
simonwng: Thanks Sergio for point it out. Hi @carolzato, please delete the temporary object `stringMapContacts` created here to fix the memory leak.

bindings/ios/MEGASdk.h
```diff
@@ -7038,6 +7053,96 @@ typedef NS_ENUM(NSInteger, BusinessStatus) {
...
+ *
+ * @param verificationCode A string supplied by the user, that they should have received via txt.
+ * @param delegate A MEGARequestDelegate callback to track this request
+ */
+- (void)checkSMSVerificationCode:(NSString *)verificationCode delegate:(id<MEGARequestDelegate>)delegate;
+
+/*
+ * @brief Requests the user contacts registered in MEGA and verificated through SMS.
+ *
+ * The associated request type with this request is MegaRequest::TYPE_GET_REGISTERED_CONTACTS
```
simonwng: @jcmartinac, for the document of the method `getRegisteredContacts`, please convert the SDK classes and enums to our iOS binding project counterparts. For example, please use `MEGARequest`, `MEGAErrorTypeApiOk`, and `MEGARequestDelegate`.

bindings/ios/MEGARequest.h
```diff
@@ -484,6 +485,20 @@ typedef NS_ENUM (NSInteger, MEGANodeAccessLevel) {
...
+@property (readonly, nonatomic) NSDictionary<NSString *, MEGAStringList*> *megaStringListDictionary;
+
+/**
+ * @brief Gets the string table response from a request.
+ *
+ * This value is valid for these requests:
+ * - [MEGASdk getRegisteredContacts:] - Obtains the user contacts registered in MEGA and verificated through SMS.
+ *
+ */
+@property (readonly, nonatomic) NSArray<MEGAStringList *> *megaStringTableArray;
```
simonwng: @jcmartinac , I think maybe `megaStringListArray` is a better name? the term `table` can be quite confusing in naming. It doesn't imply an Array.

bindings/ios/MEGARequest.mm
```diff
@@ -193,4 +208,14 @@ - (NSInteger)numDetails {
     return self.megaRequest ? self.megaRequest->getNumDetails() : 0;
 }
 
+- (NSArray<MEGAStringList *> *)megaStringTableArray {
+    MegaStringTable *table = self.megaRequest ? self.megaRequest->getMegaStringTable() : nil;
```
simonwng: @jcmartinac, we don't need to check NULL here, as both `self.megaRequest` and `self.megaRequest->getMegaStringTable()` should not be NULL when the related request comes back successfully. So, `MegaStringTable *table = self.megaRequest->getMegaStringTable()` is enough here.

bindings/ios/MEGARequest.mm
```diff
@@ -193,4 +208,14 @@ - (NSInteger)numDetails {
...
+    NSMutableArray<MEGAStringList *> *array = NSMutableArray.alloc.init;
```
simonwng: @jcmartinac , since we know the size here, please use `arrayWithCapacity:` method.

bindings/ios/MEGARequest.mm
```diff
@@ -193,4 +208,14 @@ - (NSInteger)numDetails {
...
+    for (int i = 0; i < table->size(); i++) {
+        [array addObject:[[MEGAStringList alloc] initWithMegaStringList:(MegaStringList *)table->get(i) cMemoryOwn:YES]];
```
simonwng: @jcmartinac, memory leaks here. Please copy the returned `MegaStringList` because we need to take the ownership here. `table->get(i)->copy()`

```suggestion
        [array addObject:[[MEGAStringList alloc] initWithMegaStringList:(MegaStringList *)table->get(i)->copy() cMemoryOwn:YES]];
```

simonwng:
To make @cb-mega's life easier, to merge this PR sooner, and to let @cb-mega focus on his awesome C++ code, I have created a separate branch for iOS Binding project of SMS Verification - https://github.com/meganz/sdk/tree/feature/sms_verification_iOS_binding. This branch is only for the iOS binding code on top of the SDK sms verification branch.

Later on, after we finished the iOS development, we can create a separate PR to merge the iOS binding branch to a target SDK branch.

This way, we don't need to mix the iOS code review and SDK code review together, which makes the code reviews on both side harder.

@jnavarrom @sergiohs84 , what do you guys think? If you guys agree, we can setup a rule to always create a separate iOS binding branch on top of SDK feature branch going forward.

@jcmartinac, please also start using the new SDK branch https://github.com/meganz/sdk/tree/feature/sms_verification_iOS_binding in your iOS project. I would also suggest you create another SDK branch for your invitation flow. At the end of the development, you can submit 2 PRs, one for iOS, the other for iOS binding. What do you think?

bindings/ios/MEGASdk.mm
jcmartinac: Done

bindings/ios/MEGARequest.mm
jcmartinac: Thanks Simon. I have been talking with Navarro and Santana about the naming and the return object to decide returning native classes instead of SDK classes, so I have refactored this method. This change is not needed.

bindings/ios/MEGARequest.h
jcmartinac: I use `table` following the naming of the SDK property. It think this is hard to name it in iOS but after talking with Navarro and Santana (and refactoring the code), `megaStringListArray` doesn't make sense because we are not going to use `MEGAStringList` in the return object. For the moment it is called `stringTableArray` without _mega_ prefix. But I think we will discuss this in the near future.

bindings/ios/MEGARequest.mm
simonwng: Thanks Carlos. Yes, I also think `NSArray<NSString *> *` would be better than `MEGAStringList` for our native code. I can not see your changes. Have you pushed your local changes?

bindings/ios/MEGARequest.h
simonwng: Thanks Carlos. Yes, the `table` in SDK means a list of list or array of array. If you use `megaStringTableArray`, I would think this is a 3 layers list. 

I am OK to remove `mega` prefix. Maybe name it to `stringTable` if you want to follow its SDK naming.

bindings/ios/MEGARequest.mm
jcmartinac: Simon, I have just pushed the commit. I wanted to check some things with Sergio before push it.

bindings/ios/MEGARequest.h
```diff
@@ -484,6 +485,20 @@ typedef NS_ENUM (NSInteger, MEGANodeAccessLevel) {
...
+@property (readonly, nonatomic) NSDictionary<NSString *, MEGAStringList*> *megaStringListDictionary;
+
+/**
+ * @brief Gets the string table response from a request mapped into a collection of NSArray of NSStrings.
+ *
+ * This value is valid for these requests:
+ * - [MEGASdk getRegisteredContacts:] - Obtains the user contacts registered in MEGA and verificated through SMS.
+ *
+ */
+@property (readonly, nonatomic) NSArray<NSArray <NSString *> *> *stringTableArray;
```
simonwng: spacing, should be `NSArray<NSString *>`

bindings/ios/MEGARequest.mm
```diff
@@ -193,4 +208,19 @@ - (NSInteger)numDetails {
     return self.megaRequest ? self.megaRequest->getNumDetails() : 0;
 }
 
+- (NSArray<NSArray <NSString *> *> *)stringTableArray {
+    MegaStringTable *table = self.megaRequest->getMegaStringTable();
+    NSMutableArray<NSArray <NSString *> *> *stringTableArray = [NSMutableArray.alloc initWithCapacity:table->size()];
```
simonwng: spacing here as well

PR #1495, opened 2019-06-17
Send the request reason with `uq`

mattw-mega:
for requests for storage stats only - to determine how to best reduce server load.

### DIFF
--- a/include/megaapi.h
+++ b/include/megaapi.h
@@ -8600,7 +8600,7 @@ class MegaApi
          * @param pro If true, pro level of account is requested
          * @param listener MegaRequestListener to track this request
          */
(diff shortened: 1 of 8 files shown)


include/megaapi.h
```diff
@@ -8600,7 +8600,7 @@ class MegaApi
          * @param pro If true, pro level of account is requested
          * @param listener MegaRequestListener to track this request
          */
-        void getSpecificAccountDetails(bool storage, bool transfer, bool pro, MegaRequestListener *listener = NULL);
+        void getSpecificAccountDetails(bool storage, bool transfer, bool pro, MegaRequestListener *listener, int source);
```
sergiohs84: Let's keep the `listener` as the very last parameter of the function, as it is in all the asynchronous methods.
Also, `source` needs to be documented.

PR #1498, opened 2019-06-17
#12285 Ability to cancel registration's signup link

sergiohs84:


src/megaapi_impl.cpp
```diff
@@ -6614,6 +6622,38 @@ bool MegaApiImpl::isScheduleNotifiable()
...
+    {
+        std::map<int,MegaBackupController*>::iterator it = backupsMap.begin();
+        delete it->second;
+        backupsMap.erase(it);
+    }
+
+    while (!requestMap.empty())
+    {
+        std::map<int,MegaRequestPrivate*>::iterator it = requestMap.begin();
+        if(it->second)
```
mattw-mega: if it->second is null, it will loop forever?

src/megaapi_impl.cpp
```diff
@@ -6614,6 +6622,38 @@ bool MegaApiImpl::isScheduleNotifiable()
...
+        {
+            fireOnRequestFinish(it->second, MegaError(preverror ? preverror : MegaError::API_EACCESS));
+        }
+    }
+
+    while (!transferMap.empty())
+    {
+        std::map<int, MegaTransferPrivate *>::reverse_iterator it = transferMap.rbegin();
+        if (it->second)
```
mattw-mega: if it->second is null, it will loop forever?

src/megaapi_impl.cpp
sergiohs84: That's true at first sight, but no item in these maps will point to a `nullptr`, AFAIK. However, that makes me think about the reason for that checkup. I'll adjust the loop to avoid the situation. Thanks

src/megaapi_impl.cpp
sergiohs84: Same as above

PR #1511, opened 2019-06-26
Fix/prevent shell dll blocking windows explorer

mattw-mega:


include/megaapi_impl.h
```diff
@@ -2532,7 +2508,9 @@ class MegaApiImpl : public MegaApp
         vector<string> excludedPaths;
         long long syncLowerSizeLimit;
         long long syncUpperSizeLimit;
-        MegaMutex sdkMutex;
+        std::recursive_timed_mutex sdkMutex;
+        typedef std::unique_lock<std::recursive_timed_mutex> SdkMutexGuard;
```
cb-mega: I'd use `using` here

src/raid.cpp
```diff
@@ -39,12 +39,7 @@ struct FaultyServers
     // This class may be shared amongst many megaclients, so thread safety is needed
     typedef map<string, m_time_t> Map;
     Map recentFails;
-    MUTEX_CLASS m;
-
-    FaultyServers()
-    {
-        m.init(false);
-    }
+    std::mutex m;
```
cb-mega: Can we rename `m` to something sensible?

include/megaapi_impl.h
```diff
@@ -2532,7 +2508,9 @@ class MegaApiImpl : public MegaApp
...
+        bool syncPathStateLockTimeout = false;
```
cb-mega: atomic

include/megaapi_impl.h
```diff
@@ -2532,7 +2508,9 @@ class MegaApiImpl : public MegaApp
         vector<string> excludedPaths;
         long long syncLowerSizeLimit;
         long long syncUpperSizeLimit;
-        MegaMutex sdkMutex;
+        std::recursive_timed_mutex sdkMutex;
+        using SdkMutexGuard = std::unique_lock<std::recursive_timed_mutex>;   // (equivalent to typedef)
+        std::atomic<bool> syncPathStateLockTimeout{ false };
```
polmr: we might run in to some compilation issues with this: I recall some compilator having issues with initializations in the class (though it's c++11 compliant). We might want to initialize this via atomic_init  in ::init

sergiohs84: Yes, WP does not support C++11 and it was required to adjust those initializations for the latest version. However, it's not expected to maintain that app any more, since UWP is the standard now. So AFAIK it should be OK

mattw-mega: thanks - let's try it like this for now then, and if it becomes apparent that we need to init manually in the constructor then we can make that change later.   Setting a default initialization in the header is more maintainable if we can do that, eg. a class with multiple constructors can easily end up with only one constructor updated when a new member is added.

PR #1530, opened 2019-07-02
WIP: Do not use `cout` in GTests, but logger

sergiohs84:


sergiohs84:
@mattw-mega will surround with `#ifdef`s those tests that should run exclusively in debug mode, but will first wait for a big update and rearrangement of the test code pending.

mattw-mega:
It could be a while before that rearrangement makes it to `develop`, so I've updated it here.  cheers

PR #1531, opened 2019-07-02
Task #12221 Add the creation-timestamp to public-links

jgandres:


src/node.cpp
```diff
@@ -477,9 +489,11 @@ bool Node::serialize(string* d)
         d->append(fileattrstring.c_str(), ll);
     }
 
+    bool hasLinkCreationTs = true;
     char isExported = plink ? 1 : 0;
     d->append((char*)&isExported, 1);
-    d->append("\0\0\0\0\0\0", 7);
+    d->append((char*)&hasLinkCreationTs, 1);
```
sergiohs84: I think it would be better to declare `hasLinkCreationTs` as a `char`, rather than a `boolean`. Later you will read it as char, so...
Also, please, declare variables next to their usage.
I'll fix it

sergiohs84:
> Apart from this, we need to discuss whether a reload of the cache is worthy to get the right creation timestamp from API.

Apps should hide the timestamp if it's not available locally at client side. Do not perform a refresh of local cache for this minor feature.

PR #1544, opened 2019-07-10
#1234 Contact aliases

jgandres:
Users will be allowed to set nicknames/aliases for their contacts or chat participants.


src/megaapi_impl.cpp
```diff
@@ -14216,73 +14278,155 @@ void MegaApiImpl::getua_result(byte* data, unsigned len, attr_t type)
...
 
     if (tlv)
     {
-        // TLV data usually includes byte arrays with zeros in the middle, so values
-        // must be converted into Base64 strings to avoid problems
-        MegaStringMap *stringMap = new MegaStringMapPrivate(tlv->getMap(), true);
-        request->setMegaStringMap(stringMap);
-
-        switch (request->getParamType())
+        if (request->getType() == MegaRequest::TYPE_SET_ATTR_USER)
```
sergiohs84: This `if` should finish in a `return`, since you don't want to finish the request at this stage, but when the `putua()` finishes. That way you can also avoid moving the existing block into an `else` (and generating an inconvenient diff).

src/megaapi_impl.cpp
```diff
@@ -14216,73 +14278,155 @@ void MegaApiImpl::getua_result(byte* data, unsigned len, attr_t type)
...
+            // Iterate though the string map <uh_B64, alias>
+            for (int i=0; i < keys->size(); i++)
             {
-                const char *num = stringMap->get("num");
+                key = keys->get(i);
+                alias = newAliasesMap->get(key);
 
-                if (request->getNumDetails() == 0)  // used to check if rich-links are enabled
+                // If the user alias doesn't exists or has been changed, update the TLV container
+                if (alias)
```
sergiohs84: Is it possible to get an `alias` set to `null`? :thinking:

src/megaapi_impl.cpp
```diff
@@ -14216,73 +14278,155 @@ void MegaApiImpl::getua_result(byte* data, unsigned len, attr_t type)
...
                 {
-                    if (num)
+                    rcvAlias = tlv->find(key) ?tlv->get(key).c_str() :NULL;
```
sergiohs84: Spacing, please :pray:

src/megaapi_impl.cpp
```diff
@@ -14216,73 +14278,155 @@ void MegaApiImpl::getua_result(byte* data, unsigned len, attr_t type)
...
-        // must be converted into Base64 strings to avoid problems
-        MegaStringMap *stringMap = new MegaStringMapPrivate(tlv->getMap(), true);
-        request->setMegaStringMap(stringMap);
-
-        switch (request->getParamType())
+        if (request->getType() == MegaRequest::TYPE_SET_ATTR_USER)
         {
-            // prepare request params to know if a warning should show or not
-            case MegaApi::USER_ATTR_RICH_PREVIEWS:
+            bool modified = false;
```
sergiohs84: Why do we need to check if the app's request would modify or not the attribute? It seems to over-complicate stuff, IMHO.

src/megaapi_impl.cpp
```diff
@@ -14216,73 +14278,155 @@ void MegaApiImpl::getua_result(byte* data, unsigned len, attr_t type)
...
+                modified = true;
+            }
+
+            delete keys;
+
+            if (modified)
+            {
+                // serialize and encrypt the TLV container
+                string *container = tlv->tlvRecordsToContainer(client->rng, &client->key);
+                attr_t type = attr_t(request->getParamType());
```
sergiohs84: We should be using the `type` from the input params of this callback, when possible like in this case :)

src/megaapi_impl.cpp
```diff
@@ -14216,73 +14278,155 @@ void MegaApiImpl::getua_result(byte* data, unsigned len, attr_t type)
...
-                    request->setNodeHandle(nodehandle);
+                    // If a uh was set in the request we have to find it in the aliases map and return it
+                    const char *uh = request->getText();
+                    if (uh)
+                    {
+                        const char *buf, *key;
+                        bool found = false;
+
+                        MegaStringList *keys = stringMap->getKeys();
+                        for (int i=0; i < keys->size(); i++)
```
sergiohs84: Why do we iterate here, instead of simply get the value for the known key in the map?
```request->setName(stringMap->get(uh));```

File: src/megaapi_impl.cpp [hunk 7]
```diff
@@ -18427,30 +18571,39 @@ void MegaApiImpl::sendPendingRequests()
...
-                    key = keys->get(i);
-                    buf = stringMap->get(key);
-
-                    size_t len = strlen(buf)/4*3+3;
-                    value.resize(len);
-                    value.resize(Base64::atob(buf, (byte *)value.data(), int(len)));
-
-                    tlv.set(key, value);
+                    // always get updated value before update it
+                    const char *uh = getMyUserHandle();
```
sergiohs84: We have now the `MegaClient::uid` that doesn't need to convert `me` into B64 and delete later on. Better to use that one :)

src/megaapi_impl.cpp
```diff
@@ -18427,30 +18571,39 @@ void MegaApiImpl::sendPendingRequests()
...
+                    client->getua(uh, type);
+                    delete [] uh;
                 }
-                delete keys;
+                else
```
sergiohs84: In order to avoid moving the block below, you could `break;` at the previous block for the alias --> nicer `diff`

include/megaapi.h
```diff
@@ -6459,7 +6460,8 @@ class MegaApi
             USER_ATTR_GEOLOCATION = 22,          // private - byte array
             USER_ATTR_CAMERA_UPLOADS_FOLDER = 23,// private - byte array
             USER_ATTR_MY_CHAT_FILES_FOLDER = 24, // private - byte array
-            USER_ATTR_PUSH_SETTINGS = 25         // private - char array
+            USER_ATTR_PUSH_SETTINGS = 25,        // private - char array
+            USER_ATTR_ALIAS = 27,                // private - byte array
```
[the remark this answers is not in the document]
jgandres: It exists `ATTR_UNSHAREABLE_KEY =  26`

include/megaapi.h
```diff
@@ -9747,6 +9753,64 @@ class MegaApi
...
+         * @brief Gets the alias for an user
+         *
+         * The associated request type with this request is MegaRequest::TYPE_GET_ATTR_USER
+         * Valid data in the MegaRequest object received on callbacks:
+         * - MegaRequest::getParamType - Returns the attribute type MegaApi::USER_ATTR_ALIAS
+         * - MegaRequest::getText - user handle encoded in B64
+         *
+         * Valid data in the MegaRequest object received in onRequestFinish when the error code
+         * is MegaError::API_OK:
+         * - MegaRequest::getMegaStringMap - MegaStringMap with user handles and alias both encoded in B64
```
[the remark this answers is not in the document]
jgandres: Yes, we can save some bytes by no returning the attribute. The reason to return API_EACCESS is to difference if the attribute doesn't exists API_ENOENT  or doesn't contains the alias for the target user API_EACCESS.

src/megaapi_impl.cpp
jgandres: If we haven't modified the attribute we don't need to send a putua

src/megaapi_impl.cpp
jgandres: MegaStringMapPrivate::get returns NULL if don't find the key

include/mega/types.h
```diff
@@ -438,7 +438,8 @@ typedef enum {
     ATTR_CAMERA_UPLOADS_FOLDER = 23,        // private - byte array - non-versioned
     ATTR_MY_CHAT_FILES_FOLDER = 24,         // private - byte array - non-versioned
     ATTR_PUSH_SETTINGS = 25,                // private - non-encripted - char array in B64 - non-versioned
-    ATTR_UNSHAREABLE_KEY = 26               // private - char array
+    ATTR_UNSHAREABLE_KEY = 26,              // private - char array
+    ATTR_ALIAS = 27,                        // private - byte array - non-versioned
```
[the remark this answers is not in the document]
jgandres: The attribute stores multiple aliases but it's name is `!>alias` so I'm not sure if we should change it.

include/megaapi.h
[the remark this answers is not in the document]
sergiohs84: You are right! And it's good to not list it there, since the app cannot and should not access to that internal attribute.

src/megaapi_impl.cpp
sergiohs84: Agree! Let's keep that optimization, the chances for race-conditions are very low and their consequences are not critical.

src/megaapi_impl.cpp
```diff
@@ -14276,14 +14379,33 @@ void MegaApiImpl::getua_result(TLVstore *tlv, attr_t)
...
                 break;
             }
+
+            case MegaApi::USER_ATTR_ALIAS:
+            {
+                // If a uh was set in the request we have to find it in the aliases map and return it
+                const char *uh = request->getText();
+                if (uh)
+                {
+                    request->setMegaStringMap(NULL);
```
sergiohs84: I think we can keep the original object in there, it could be useful, despite we don't expose it in the documentation, but internally. I'll change it ;)

src/user.cpp
```diff
@@ -709,6 +717,7 @@ int User::needversioning(attr_t at)
         case ATTR_GEOLOCATION:
         case ATTR_MY_CHAT_FILES_FOLDER:
         case ATTR_PUSH_SETTINGS:
+        case ATTR_ALIAS:
```
sergiohs84: Since this attribute should be updated always based on the latest version, we should consider versioning. Otherwise, we can overwrite aliases set by a different client if we haven't received yet the corresponding actionpacket that invalidates our version and triggers the corresponding fetch.
Remember the `!` modifier in the name means the API doesn't store the history of values, but we can avoid race conditions thanks to sending to API our version, and API will reject the update if we are not up to date.
I'll change it to versioning.

src/megaapi_impl.cpp
```diff
@@ -9825,6 +9830,30 @@ void MegaApiImpl::setMyChatFilesFolder(MegaHandle nodehandle, MegaRequestListene
...
+    request->setNodeHandle(uh);
+    request->setText(Base64Str<MegaClient::USERHANDLE>(uh));
+    requestQueue.push(request);
+    waiter->notify();
+}
+
+void MegaApiImpl::setUserAlias(MegaHandle uh, const char *alias, MegaRequestListener *listener)
+{
+    MegaRequestPrivate *request = new MegaRequestPrivate(MegaRequest::TYPE_SET_ATTR_USER, listener);
+    MegaStringMap *stringMap = new MegaStringMapPrivate();
```
mattw-mega: In cases like this where the object will be deleted again a few lines later, it can just be declared on the stack: `MegaStringMapPrivate stringMap;`, and then passed to functions with `&stringMap`, and then there's less heap activity, and no need to manually delete it.    Alternatively the setMegaStringMap could just take ownership, and then there would be no need to make a copy and delete the original, which seems like a lot more work than necessary.  Maybe add a flag on setMegaStringMap as to whether it should assume ownership, or copy?    thanks

sergiohs84: Thank you, Matt. The `stringMap` can be created in the heap, indeed. I agree. I'll adjust it.
For the setter of the `MegaRequest`, I'd prefer to keep it as it is for historical reasons, where they always do a copy. I'm aware of the extra work, but it helps Java bindings not to mess up with the passed objects. Basically, the SDK never takes ownership of the passed object :)

PR #1546, opened 2019-07-11
Fix/new test structure

cb-mega:
This restructures our tests in:
- unit tests
- integration tests
- tools
- python tests

This is done for QMake, CMake, and autoconf.
Once this code makes it to develop we need to adjust a few Jenkins jobs.


mattw-mega:
It's quite a big rearrangement - just trying to see where the main SDK tests went, they seem to be in SdkTest_test.cpp now, but it's lost all its history?

cb-mega:
It's a rename: `tests/sdk_test.cpp → tests/integration/SdkTest_test.cpp` so history should be preserved. Note that I followed the standard naming scheme where the test file contains the name of the test sute.

polmr:
trigger compilation

cb-mega:
@polmr It won't run successfully on jenkins currently because the test executables have moved. E.g. Unit tests are now at `tests/test_unit` when built with autotools. Once this change makes it to develop we'll have to adjust jenkins.

mattw-mega:
Having found the right setting to 'follow renames', the change is pretty straightforward - I'll approve it.

polmr:
trigger compilation

polmr:
trigger compilation

polmr:
trigger compilation

polmr:
trigger compilation

PR #1547, opened 2019-07-12
#12268 Improve localnode handling when fsid changes

cb-mega:


src/sync.cpp
```diff
@@ -1276,5 +1312,154 @@ bool Sync::movetolocaldebris(string* localpath)
...
+            if (isscannable(localpath))
+            {
+                auto localfa = std::unique_ptr<FileAccess>{client->fsaccess->newfileaccess()};
+                if ((success = localfa->fopen(&localpath, true, false)))
+                {
+                    if (localfa->type == FILENODE)
+                    {
+                        success = assignfsid(*localfa, localpath, nodes);
+                    }
+                    else if (localfa->type == FOLDERNODE)
```
mattw-mega: shouldn't we assign fsids to localnodes representing folders also, so we can detect folder moves?  Though, it might be easier to do so by walking the localnode tree rather than walking the filesystem

src/sync.cpp
```diff
@@ -411,16 +414,54 @@ LocalNode* Sync::localnodebypath(LocalNode* l, string* localpath, LocalNode** pa
...
+    auto rootpath = localroot.localname;
+    if ((success = fa->fopen(&rootpath, true, false)))
+    {
+        if ((success = fa->type == FOLDERNODE))
+        {
+            assignfsids(success, rootpath, std::move(fa), nodes);
+        }
+        else
+        {
+            LOG_err << "rootpath not a folder";
```
mattw-mega: I think syncs can also be to the whole cloud folder, or an incoming node, not just foldernode?

src/sync.cpp
```diff
@@ -1276,5 +1312,154 @@ bool Sync::movetolocaldebris(string* localpath)
...
+        const auto node = static_cast<LocalNode*>(*nodeIt);
+        std::string nodePath;
+        node->getlocalpath(&nodePath);
+        if (nodePath == path)
+        {
+            localNode = node;
+            break;
+        }
+    }
+    localNode->setfsid(fa.fsid);
```
mattw-mega: should we assign if the path doesn't match?  I think that may cause issues

src/megaclient.cpp
```diff
@@ -11373,6 +11373,18 @@ error MegaClient::addsync(string* rootpath, const char* debris, string* localdeb
             Sync* sync = new Sync(this, rootpath, debris, localdebris, remotenode, fsfp, inshare, tag, appData);
             sync->isnetwork = isnetwork;
 
+            if (!sync->fsstableids)
+            {
+                // Ensures that unmatched nodes (local nodes that don't have a fingerprint that's
+                // the same as a file on disk) have invalid IDs.
+                sync->invalidatefsids();
+
+                if (!sync->assignfsids())
```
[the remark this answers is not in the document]
cb-mega: The former is logged in sync.cpp. I will add logging for successful scanning as well.

src/sync.cpp
cb-mega: We're only recursing through local folders here so the root must be a foldernode. It's equivalent to the general scanning we do when adding a sync (see Sync::scan). Does that make sense?

src/sync.cpp
cb-mega: We cannot assign IDs to folders because we don't know which ID to assign. Folders don't have a full fingerprint so we cannot construct any heuristic to map fingerprints to IDs like we do with files. Does that make sense?

src/sync.cpp
cb-mega: Yes, we need to assign. The matching path is just a heuristic because for the most likely case. However, it would also be correct to assign the ID to any random node in the range because we simply don't know what happened on the file system while the sync was inactive. Thoughts?

src/win32/fs.cpp
```diff
@@ -1259,6 +1259,25 @@ fsfp_t WinDirNotify::fsfingerprint()
 #endif
 }
 
+bool WinDirNotify::fsstableids() const
```
cb-mega: @mattw-mega Would you mind testing this on windows? Set up a sync to a FAT drive and ensure this function returns false? Thanks!

src/sync.cpp
mattw-mega: I think if you set up a sync to your whole cloud drive, the root cloud node type would be ROOTNODE.

include/mega/filefingerprint.h
```diff
@@ -19,37 +19,33 @@
...
+    bool isvalid{};
 
-    FileFingerprint& operator=(FileFingerprint&);
+    bool genfingerprint(FileAccess* fa, bool ignoremtime = false);
+    bool genfingerprint(InputStreamAccess* is, m_time_t cmtime, bool ignoremtime = false);
+    void serializefingerprint(string* d) const;
+    int unserializefingerprint(string* d);
 
-    FileFingerprint();
+    FileFingerprint() = default;
```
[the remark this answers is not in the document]
mattw-mega: any update on this one?  thanks

include/mega/filesystem.h
```diff
@@ -112,7 +112,7 @@ struct MEGA_API FileAccess
     bool fread(string *, unsigned, unsigned, m_off_t);
 
     // absolute position read to byte buffer
-    bool frawread(byte *, unsigned, m_off_t);
+    virtual bool frawread(byte *, unsigned, m_off_t);
```
mattw-mega: I don't think this should become virtual, that's affecting/slowing production code - the tests that want to use this can override the four customization functions that are already virtual per the design: sysread/sysstat/sysopen/sysclose - just like a customization for a particular platform, but that platform is virtual. thanks!

include/mega/filesystem.h
```diff
@@ -248,7 +252,7 @@ struct MEGA_API FileSystemAccess : public EventTrigger
 
     // convert MEGA-formatted filename (UTF-8) to local filesystem name; escape
     // forbidden characters using urlencode
-    void local2name(string*) const;
+    virtual void local2name(string*) const;
```
mattw-mega: similarly, this one shouldn't become virtual, it calls the virtual implementation path2local/local2path, the customization should be on those two.

include/mega/node.h
```diff
@@ -337,7 +338,7 @@ struct MEGA_API LocalNode : public File
 
     void setnotseen(int);
 
-    void setfsid(handle);
+    void setfsid(handle newfsid, handlelocalnode_map& fsidnodes);
```
mattw-mega: I think we should not make this change - the old one is more efficient, and clearer.   We do always operate on the map from the client, and this signature implies we might operate on some other map.   Also it adds a lot of extra diff lines, for no functional change.

include/mega/node.h
```diff
@@ -110,7 +110,7 @@ struct Fingerprints
 // filesystem node
 struct MEGA_API Node : public NodeCore, FileFingerprint
 {
-    MegaClient* client;
+    MegaClient* client = nullptr;
```
mattw-mega: I don't think it can ever be null in practice though?   setting it null here is a bit misleading?

include/mega/sync.h
```diff
@@ -26,31 +26,44 @@
...
+                         const string& localdebris, const string& localseparator, bool followsymlinks);
+
 class MEGA_API Sync
 {
 public:
-    void *appData;
+    void* appData = nullptr;
 
-    MegaClient* client;
+    MegaClient* client = nullptr;
```
mattw-mega: client can never actually be null?

include/mega/utils.h
```diff
@@ -408,6 +408,59 @@ struct CacheableReader
...
+/////// forEach over a std::tuple, unrolled at compile time
+
+template<typename Tuple, typename Functor>
+void forEach(Tuple&& tup, Functor&& functor)
+{
+    constexpr auto size = std::tuple_size<typename std::decay<Tuple>::type>::value;
+    using IndexType = typename IndexRange<0, size>::type;
+    forEachIndex(IndexType{}, std::forward<Tuple>(tup), std::forward<Functor>(functor));
+}
+
```
mattw-mega: As this is only used from the tests, can we put it in the test headers rather than as part of the main SDK please?

include/mega/win32/megafs.h
```diff
@@ -139,7 +140,6 @@ class MEGA_API WinFileAccess : public FileAccess
     bool fopen(string*, bool, bool, bool);
     void updatelocalname(string*) override;
     bool fread(string *, unsigned, unsigned, m_off_t);
-    bool frawread(byte *, unsigned, m_off_t);
```
mattw-mega: these classes should always override the four virtual implementation functions?

src/filefingerprint.cpp
```diff
@@ -107,16 +104,6 @@ FileFingerprint *FileFingerprint::unserialize(string *d)
...
-FileFingerprint& FileFingerprint::operator=(FileFingerprint& rhs)
-{
-    isvalid = rhs.isvalid;
-    size = rhs.size;
-    mtime = rhs.mtime;
-    memcpy(crc, rhs.crc, sizeof crc);
-
-    return *this;
-}
-
```
mattw-mega: it's a lot easier to review, and to see what's happened in the history, if we don't move these functions around

src/filefingerprint.cpp
```diff
@@ -373,6 +360,31 @@ int FileFingerprint::unserializefingerprint(string* d)
...
+    for (const auto val : crc)
+    {
+        hashCombine(value, val);
+    }
+    return value;
+}
+
+FileFingerprint& FileFingerprint::operator=(const FileFingerprint& other)
+{
+    if (this != &other)
```
mattw-mega: I don't think we need to add this self-assignment check, since the self-assignment case will work ok without it, and the test is just extra work that has to be done every other time (I doubt we have any self assignment of these anyway)

src/node.cpp
```diff
@@ -1496,7 +1496,7 @@ void LocalNode::setnotseen(int newnotseen)
 }
 
 // set fsid - assume that an existing assignment of the same fsid is no longer current and revoke
-void LocalNode::setfsid(handle newfsid)
+void LocalNode::setfsid(handle newfsid, handlelocalnode_map& fsidnodes)
```
mattw-mega: clearer if we access sync->client->fsidnode directly, rather than passing it in by parameter - no need for this change

src/node.cpp
```diff
@@ -1538,6 +1538,11 @@ LocalNode::~LocalNode()
         return;
     }
 
+    if (!sync->client)
```
mattw-mega: client can never be null?

src/node.cpp
```diff
@@ -1628,7 +1633,7 @@ LocalNode::~LocalNode()
     }
 }
 
-void LocalNode::getlocalpath(string* path, bool sdisable) const
+void LocalNode::getlocalpath(string* path, bool sdisable, const std::string* localseparator) const
```
mattw-mega: we are always using the client's fsaccess's localseparator, no need for this change?

src/sync.cpp
```diff
@@ -36,6 +38,425 @@ const int Sync::FILE_UPDATE_DELAY_DS = 30;
...
+
+private:
+    size_t mHash = 0;
+    bool mIsValid = false;
+};
+
+bool operator==(const Fingerprint& lhs, const Fingerprint& rhs)
+{
+    return lhs.getHash() == rhs.getHash();
+}
```
mattw-mega: I think we can reuse FileFingerprint instead of this extra class - this one tries to boil a fingerprint down to a size_t but that will lose precision - FileFingerprint is already not perfectly guaranteed not to collide, but we should maintain that level of accuracy, I think - one size_t is far, far less accurate.    For a folder we can put the hash of the filenames in the 4 crc uint32s, and maybe set a high bit in the size field, to ensure they don't collide with the fingerprint of any actual file.

src/sync.cpp
```diff
@@ -135,13 +559,19 @@ Sync::~Sync()
     if (localroot.node)
     {
         TreeProcDelSyncGet tdsg;
-        client->proctree(localroot.node, &tdsg);
+        if (client)
```
mattw-mega: client can never be null?

src/sync.cpp
```diff
@@ -135,13 +559,19 @@ Sync::~Sync()
...
+        {
+            client->proctree(localroot.node, &tdsg);
+        }
     }
 
     delete statecachetable;
 
-    client->syncs.erase(sync_it);
-    client->syncactivity = true;
+    if (client)
```
mattw-mega: ditto

include/mega/filesystem.h
cb-mega: Ok. That should be doable. Will fix

include/mega/filesystem.h
cb-mega: Indeed. Will fix.

include/mega/node.h
cb-mega: This initialization is not used when the constructor initializer list is used. If not, then this is the default as opposed to some random memory address. I think this makes code well defined and predictable. 

Also, I have unit tests coming up in another branch that require the client to be null. We need this so that we can circumvent certain code that requires a valid client. Note that the client can currently not be mocked out. Alternatively, we could introduce a ClientInterface but that's a lot of changes we might not want.

include/mega/node.h
cb-mega: This change is required because we cannot mock out the client. I doubt it's any less efficient with modern compilers.

include/mega/sync.h
cb-mega: Has to be null initialized. Similar arguments as for `Node`.

include/mega/utils.h
cb-mega: This should be part of production code and I am soon planning on using it to iterate over a tuple.

include/mega/win32/megafs.h
cb-mega: frawread is not overriden if you look at the cpp of this class. frawread is not even virtual. It was an oversight to paste it here.

src/filefingerprint.cpp
cb-mega: I think the order of functions in the header should match that in the cpp. We should clean this up whenever we change code in this area. This will make reading this module easier and this review only needs to done once.

src/filefingerprint.cpp
cb-mega: I will add an assert instead

src/node.cpp
cb-mega: See my above above for the header.

src/node.cpp
cb-mega: Can be when unit testing.

src/node.cpp
cb-mega: We need this for unit testing. Similar to `setfsid`

src/sync.cpp
cb-mega: Will change to as discussed IRL.

src/sync.cpp
cb-mega: See comments above re unit testing.

src/sync.cpp
cb-mega: See comments above re unit testing.

src/sync.cpp
cb-mega: We're dealing with local nodes here.

src/filefingerprint.cpp
cb-mega: I moved it back where it was for now.

include/mega/utils.h
cb-mega: I just added another change that uses `forEach` in `MegaApiImpl::getregisteredcontacts_result`. I ran the tests hitting this code and they're still passing. 

Sorry for this not being related to the actual fix anymore but I thought it would still be useful to have.

tests/unit/Commands_test.cpp
```diff
@@ -1,7 +1,4 @@
 /**
- * @file tests/commands_test.cpp
```
cb-mega: I removed these two lines (here and elsewhere) as they lead to copy/paste errors and become stale when the file is renamed.

contrib/QtCreator/MEGAtests/MEGAtest_unit/MEGAtest_unit.pro
```diff
@@ -16,15 +17,30 @@ CONFIG += USE_MEDIAINFO
 CONFIG += USE_LIBRAW
 CONFIG += USE_FFMPEG
 CONFIG -= qt
+CONFIG += object_parallel_to_source
```
cb-mega: We need this so that the QMake build doesn't break. By default, QMake puts all object files into a flat directory which means that if you have utils.cpp under src and utils.cpp under test/unit your build breaks because the object files overwrite each other.

examples/megacli.cpp
```diff
@@ -506,7 +511,7 @@ AppFileGet::AppFileGet(Node* n, handle ch, byte* cfilekey, m_off_t csize, m_time
 
         if (!cfingerprint->size() || !unserializefingerprint(cfingerprint))
         {
-            memcpy(crc, filekey, sizeof crc);
+            memcpy(crc.data(), filekey, sizeof crc);
```
cb-mega: FileFingerprint's `crc` is now a `std::array` which makes working with it a lot easier.

include/mega/sync.h
```diff
@@ -26,31 +26,44 @@
 #include "megaclient.h"
 
 namespace mega {
+
```
cb-mega: The following four functions are free, pure functions to make the functionality they implement easier to unit test.

include/mega/sync.h
```diff
@@ -82,55 +95,63 @@ class MEGA_API Sync
...
     // values related to possible files being updated
-    m_off_t updatedfilesize;
-    m_time_t updatedfilets;
-    m_time_t updatedfileinitialts;
-    
+    m_off_t updatedfilesize = 0;
+    m_time_t updatedfilets = 0;
+    m_time_t updatedfileinitialts = 0;
+
+    Sync() = default;
```
cb-mega: This constructor is currently required for unit testing.

include/mega/node.h
sergiohs84: I like class members (specially pointers) to be always initialized. IMHO, I don't mind if it's set in there, but it's also true that the only user-defined `Node` ctor accepts and sets the `client`. If you have a branch where the `Node` is not created with:
```Node::Node(MegaClient* cclient, node_vector* dp, handle h, handle ph,
           nodetype_t t, m_off_t s, handle u, const char* fa, m_time_t ts)```
then this initialization makes sense in that branch, not here.

File: src/megaapi_impl.cpp [hunk 27]
```diff
@@ -15033,9 +15033,10 @@ void MegaApiImpl::getregisteredcontacts_result(error e, vector<tuple<string, str
                 for (const auto& row : *data)
                 {
                     vector<char*> list;
-                    list.emplace_back(MegaApi::strdup(std::get<0>(row).c_str()));
-                    list.emplace_back(MegaApi::strdup(std::get<1>(row).c_str()));
-                    list.emplace_back(MegaApi::strdup(std::get<2>(row).c_str()));
+                    forEach(row, [&list](const string& value)
```
sergiohs84: Before committing this kind of changes, it would be nice if they were discussed with the team. Personally, for a 3 elements list, I find the old syntax much more readable than the new one.

PR #1551, opened 2019-07-12
#12043. Chat reactions

sergiohs84:
Pending to add test-vectors from Webclient


tests/unit/Crypto_test.cpp
```diff
@@ -112,6 +112,48 @@ TEST(Crypto, xxTea)
             ASSERT_TRUE(data2[i] == uint32_t(-(int)i));
         }
     }
+
```
sergiohs84: Since all the crypto related to reactions is tested by the MEGAchat's test suite, I think we should remove the ones below (which, on top of that, were written during an initial stage, more like an experiment).

sergiohs84:
trigger compilation

PR #1552, opened 2019-07-12
#9445. Fingerprint verification

sergiohs84:
Support for user attributes related to the authentication of keys (signing, RSA and chat). Keys for all users/contacts are fetched, along with their corresponding signatures (if signed key) in order to verify the authenticity of the keys: RSA, Ed25519 and Cu25519. It's done at startup, when there's no cached session yet, as well as a new contact (not previously ex-contact) is added.
Additionally, the _Seen_ level of trust for Ed25519 can be upgraded to _Fingerprint comparison_ if the user check them.

*Risk area/s:*
- Creation of new accounts.
- Authentication of keys upon new session.
- Verification & reset of credentials should work like in Webclient.
- New contact's relationships.


src/megaclient.cpp
```diff
@@ -3983,12 +3980,15 @@ bool MegaClient::procsc()
...
-                        if (sid.size())
+                        if (!loggedinfolderlink())
                         {
+                            // historic user alerts are not supported for public folders
                             // now that we have loaded cached state, and caught up actionpackets since that state
                             // (or just fetched everything if there was no cache), our next sc request can be for useralerts
                             useralerts.begincatchup = true;
+
+                            // now that we know all contacts, fetch public keys and check authentication
+                            fetchContactsKeys();
```
mattw-mega: I'm not sure about fetching the keys of contacts at this point - we have just decided we are `statecurrent`, so the app may expect everything including keys to be present at this point.   Also, do we need to fetch all keys of all contacts on every startup?  Hopefully these can be stored in cache, and we can just fetch those that need to be updated?   thanks

src/megaclient.cpp
```diff
@@ -5325,12 +5323,24 @@ void MegaClient::sc_userattr()
...
+                                        break;
+                                    }
+                                    case ATTR_AUTHRING:     // fall-through
+                                    case ATTR_AUTHCU255:    // fall-through
+                                    case ATTR_AUTHRSA:
+                                    {
+                                        mAuthRings.erase(type);
+                                        getua(u, type, 0);
+                                        break;
+                                    }
```
mattw-mega: It would be great if we could get the proper update via actionpacket, rather than re-requesting - I presume that has been looked into?  thanks

src/megaclient.cpp
sergiohs84: Regarding this:
> do we need to fetch all keys of all contacts on every startup? Hopefully these can be stored in cache, and we can just fetch those that need to be updated? 
It is exactly the current behavior. The `getua()`/`getpubk()` looks for the attribute in the cache. If not found, it fetches from server.

As for fetching the keys at this point on every startup, I also doubt if it's really worthy. Apart from the first update for native users with missing authrings, where checking the authrings on resume session is desired, there could be some edge cases where a new contact or a key/signature has not been checked and the app is closed. In that case, keys won't be authenticated until a new session is created.
Taking into account the computation time must be very low for the authentication, I would prefer to keep it there on every startup (as Webclient does, according to my testing). But if you think it's too heavy, we can move it a bit upwards in the code and do authentication only upon new session.

src/megaclient.cpp
sergiohs84: Yes, it would be great. But it seems it would result in much larger actionpackets in the API's cache, so the API team rejected to include the value, but only the most recent version (you may not receive one AP for each update, but only one AP for the latest change).

src/megaclient.cpp
mattw-mega: No, that's fine, thanks - I had not realised the getua() delivered from cache if it's present, rather than fetching from the servers every time :-)  cheers

src/megaclient.cpp
mattw-mega: ok, thanks for the info :-)

PR #1555, opened 2019-07-16
megacli: add a function to help check the raw attributes of a node

mattw-mega:
Also adjusted preprocessor guards for exec_log and similar functions to match their definitions (should fix mingw32 build)

### DIFF
--- a/examples/megacli.cpp
+++ b/examples/megacli.cpp
@@ -2502,7 +2502,7 @@ void exec_querytransferquota(autocomplete::ACState& s);
 
 void exec_showattributes(autocomplete::ACState& s)
 {
-    if (Node* n = nodeFromRemotePath(s.words[1].s))
+    if (const Node* n = nodeFromRemotePath(s.words[1].s))
     {
         for (auto pair : n->attrs.map)
         {
@@ -2516,9 +2516,7 @@ void exec_showattributes(autocomplete::ACState& s)
                     << " crc " << std::hex << f.crc[0] << " " << f.crc[1] << " " << f.crc[2] << " " << f.crc[3] << std::dec << ")"
 
                     << " (node fingerprint: size " << n->size << " mtime " << n->mtime
-                    << " crc " << std::hex << n->crc[0] << " " << n->crc[1] << " " << n->crc[2] << " " << n->crc[3] << std::dec << ")"
-
-                    << endl;
+                    << " crc " << std::hex << n->crc[0] << " " << n->crc[1] << " " << n->crc[2] << " " << n->crc[3] << std::dec << ")" << endl;
             }
             else
             {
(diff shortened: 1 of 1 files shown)


examples/megacli.cpp
```diff
@@ -2500,6 +2500,35 @@ void exec_treecompare(autocomplete::ACState& s);
 void exec_querytransferquota(autocomplete::ACState& s);
 
 
+void exec_showattributes(autocomplete::ACState& s)
+{
+    if (Node* n = nodeFromRemotePath(s.words[1].s))
```
cb-mega: Might be more expressive to name it `node` instead of `n`? Also should be declared const: `const Node* const node`

examples/megacli.cpp
```diff
@@ -2500,6 +2500,35 @@ void exec_treecompare(autocomplete::ACState& s);
...
+    {
+        for (auto pair : n->attrs.map)
+        {
+            char namebuf[10]{};
+            AttrMap::nameid2string(pair.first, namebuf);
+            if (pair.first == 'c')
```
cb-mega: Can we add a comment as to what `c` means?

examples/megacli.cpp
```diff
@@ -2500,6 +2500,35 @@ void exec_treecompare(autocomplete::ACState& s);
...
+            {
+                FileFingerprint f;
+                f.unserializefingerprint(&pair.second);
+                cout << namebuf << ": " << pair.second << " (fingerprint: size " << f.size << " mtime " << f.mtime
+                    << " crc " << std::hex << f.crc[0] << " " << f.crc[1] << " " << f.crc[2] << " " << f.crc[3] << std::dec << ")"
```
cb-mega: Can you output crc in a loop so we don't need to know how many elements it holds?

examples/megacli.cpp
mattw-mega: Technically `Node const * const`  would be more correct - however making pointers const is unusual and distracting - I've updated it to just `const Node*` :-)

examples/megacli.cpp
mattw-mega: The comment would just say that 'c' is the index of the FileFingerprint which is made plain on the very next line, in code.

examples/megacli.cpp
mattw-mega: I think it's a lot more compact this way - with the rules around curly braces the function would be at least 8 lines longer if those were loops.   The crc array is fundamental and won't be changing.  cheers

cb-mega: I like that you can see the future Matt :)

examples/megacli.cpp
cb-mega: In the future the code below  might move elsewhere and line 2511 might be isolated and it's not clear what it means.

PR #1556, opened 2019-07-16
Address some warnings

jnavarrom:
Override
Result of comparison of constant xxx with expression of type 'mega::error' (aka 'mega::ErrorCodes') is always false

### DIFF
--- a/include/megaapi_impl.h
+++ b/include/megaapi_impl.h
@@ -2854,7 +2854,7 @@ protected:
         void http_result(error, int, byte *, int) override;
 
         // notify about a business account status change
-        virtual void notify_business_status(BizStatus status) override;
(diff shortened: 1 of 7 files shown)


jnavarrom:
trigger compilation

include/megaapi_impl.h
```diff
@@ -2854,7 +2854,7 @@ class MegaApiImpl : public MegaApp
         void http_result(error, int, byte *, int) override;
 
         // notify about a business account status change
-        virtual void notify_business_status(BizStatus status);
+        virtual void notify_business_status(BizStatus status) override;
```
sergiohs84: We should get rid of the `virtual` here :)

polmr:
trigger compilation

PR #1559, opened 2019-07-23
WIP: Notify expiration of PRO susbcription

polmr:
send pro uq after fetch nodes to ensure a valid expiration time (for pro
accounts)

### DIFF
--- a/src/commands.cpp
+++ b/src/commands.cpp
@@ -4011,9 +4011,14 @@ void CommandGetUserQuota::procresult()
                 break;
 
             case MAKENAMEID6('s', 'u', 'n', 't', 'i', 'l'):
+            {
                 // expiry of last active Pro plan (may be different from current one)
                 details->pro_until = client->json.getint();
-                break;
+                m_time_t now = m_time(nullptr);
+                dstime diff = (details->pro_until - now) * 10;
+                client->btsuntilexpiration.backoff(diff);
+            }
+            break;
 
             case MAKENAMEID7('b', 'a', 'l', 'a', 'n', 'c', 'e'):
                 // account balances
--- a/src/megaclient.cpp
+++ b/src/megaclient.cpp
@@ -1036,6 +1036,7 @@ void MegaClient::init()
     btsc.reset();
     btpfa.reset();
     btbadhost.reset();
+    btsuntilexpiration.backoff(NEVER);
 
     abortlockrequest();
 
@@ -1194,7 +1195,6 @@ MegaClient::MegaClient(MegaApp* a, Waiter* w, HttpIO* h, FileSystemAccess* f, Db
     h->setmaxdownloadspeed(0);
     h->setmaxuploadspeed(0);
 
-    btsuntilexpiration.backoff(NEVER);
 }
 
 MegaClient::~MegaClient()
--- a/src/megaclient.cpp
+++ b/src/megaclient.cpp
@@ -1052,7 +1052,7 @@ void MegaClient::init()
 }
 
 MegaClient::MegaClient(MegaApp* a, Waiter* w, HttpIO* h, FileSystemAccess* f, DbAccess* d, GfxProc* g, const char* k, const char* u)
-    : useralerts(*this), btugexpiration(rng), btcs(rng), btbadhost(rng), btworkinglock(rng), btsc(rng), btpfa(rng)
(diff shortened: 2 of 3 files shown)


src/megaclient.cpp
```diff
@@ -10716,6 +10725,9 @@ void MegaClient::fetchnodes(bool nocache)
         if (!loggedinfolderlink())
         {
             reqs.add(new CommandGetUA(this, uid.c_str(), ATTR_DISABLE_VERSIONS, NULL, 0));
+
+            static AccountDetails ad_useless;
+            reqs.add(new CommandGetUserQuota(this, &ad_useless, false, false, true, 99));
```
sergiohs84: We should eventually document the `source`s in use. Here the `99` seems to be a magic number, I guess it doesn't collide with the sources used by MEGAsync?

src/megaclient.cpp
```diff
@@ -1193,6 +1194,7 @@ MegaClient::MegaClient(MegaApp* a, Waiter* w, HttpIO* h, FileSystemAccess* f, Db
     h->setuseragent(&useragent);
     h->setmaxdownloadspeed(0);
     h->setmaxuploadspeed(0);
+
```
sergiohs84: Empty line

src/megaclient.cpp
mattw-mega: I think 99 is fine for now, we should remove that parameter from the function anyway, later after the uq issues are solved.   However, this GetUserQuota is doubling up on the ones sent by the apps such as MEGAsync anyway, perhaps the app should ask the SDK to start the timer?  Or the SDK could just listen in on the replies to requests the app is sending anyway?

src/commands.cpp
```diff
@@ -4011,9 +4011,14 @@ void CommandGetUserQuota::procresult()
                 break;
 
             case MAKENAMEID6('s', 'u', 'n', 't', 'i', 'l'):
+            {
                 // expiry of last active Pro plan (may be different from current one)
                 details->pro_until = client->json.getint();
-                break;
+                m_time_t now = m_time(nullptr);
+                dstime diff = (details->pro_until - now) * 10;
+                client->btsuntilexpiration.backoff(diff);
```
mattw-mega: Nice use of the backoff timers, however the client side clock could be ahead of the clocks on the servers by a few second, and so the timer could go off early (from the server side perspective).  I think we should allow for that, perhaps by retrying a few times with increasing backoff times, or coordinating with the app layer somehow?  Once the account is detected to have lapsed back to a free account, we can cancel the backoffs.   Or is this field changed/absent for a lapsed-back-to-free account?  thanks

src/megaclient.cpp
sergiohs84: You are correct, the parameter should eventually be removed. Forget about adding documentation about it ;)

As for the double calls, from SDK and app, the call from the app based on a timer would be removed in pro of a more low-level call from the SDK, and the SDK would notify all the apps as needed.

src/commands.cpp
sergiohs84: If I understand the backoff correctly, I expect it to not be triggered unless the `uq` has returned an `suntil`. Once it is armed, it should result in only one `uq` and no more (as long as the account has lapsed back to a free account and remains as free).

On the other hand, I agree about the clock's skew issue. Instead of over-complicating this case with retries and alike, perhaps we can simply add a few additional seconds to the backoff? Perhaps not ideal, but simpler and effective, IMHO.

mattw-mega: Yes, that will avoid the issue initially.  Also I was concerned that we could get in a situation where the timer expires, we send uq, get the `suntil`, set the timer again which expires soon or immediately, and now we're in a loop burning the user's bandwidth, and DDOSing ourselves - so long as we protect against that somehow, I think it's good.  thanks

sergiohs84: A possible workaround could be similar to the `cachedug` flag for `ug` requests. If the flag is set, no request is sent again when the timer expires (unless the app explicitly requests it). When the backoff is armed, we reset the flag and send the request (which I think it will have the common exponential backoffs of any request). Upon response, we set the flag. Additionally, if the `suntil` is in the past, simply forget about it to not request once and again.
@polmr would you mind to apply the changes above if you agree with the suggestion? Thanks

PR #1560, opened 2019-07-29
megacli: option for du to count storage for all nodes

mattw-mega:
or to break it down per root (cloud root, rubbish, shares, etc)

### DIFF
--- a/examples/megacli.cpp
+++ b/examples/megacli.cpp
@@ -2571,7 +2571,7 @@ autocomplete::ACN autocompleteSyntax()
     p->Add(exec_rm, sequence(text("rm"), remoteFSPath(client, &cwd)));
     p->Add(exec_mv, sequence(text("mv"), remoteFSPath(client, &cwd, "src"), remoteFSPath(client, &cwd, "dst")));
     p->Add(exec_cp, sequence(text("cp"), remoteFSPath(client, &cwd, "src"), either(remoteFSPath(client, &cwd, "dst"), param("dstemail"))));
-    p->Add(exec_du, sequence(text("du"), remoteFSPath(client, &cwd)));
+    p->Add(exec_du, sequence(text("du"), either(sequence(flag("-all"), opt(flag("-breakdown"))), remoteFSPath(client, &cwd))));
 #ifdef ENABLE_SYNC
     p->Add(exec_sync, sequence(text("sync"), opt(sequence(localFSPath(), either(remoteFSPath(client, &cwd, "dst"), param("cancelslot"))))));
 #endif
@@ -3316,6 +3316,40 @@ void exec_cp(autocomplete::ACState& s)
 
 void exec_du(autocomplete::ACState& s)
 {
+    if (s.extractflag("-all"))
+    {
+        bool breakdown = s.extractflag("-breakdown");
+        struct counters { m_off_t filecount = 0, filebytes = 0, fileversioncount = 0, fileversionbytes = 0, foldercount = 0; };
+        map<Node*, counters> values;
+        for (auto& n : client->nodes)
+        {
+            auto& c = values[breakdown ? n.second->firstancestor() : nullptr];
+            switch (n.second->type)
+            {
+                case FILENODE:
+                {
+                    bool isversion = n.second->parent && n.second->parent->type == FILENODE;
+                    (isversion ? c.fileversioncount : c.filecount) += 1;
+                    (isversion ? c.fileversionbytes : c.filebytes) += n.second->size;
+                    break;
+                }
+                case FOLDERNODE:
+                {
+                    c.foldercount += 1;
+                    break;
+                }
+            }
+        }
+        for (auto& v : values)
+        {
+            if (breakdown)  cout << v.first->nodehandle << endl;
+            cout << "files: " << v.second.filecount << " file bytes: " << v.second.filebytes << endl;
+            cout << "versions: " << v.second.fileversioncount << " version bytes: " << v.second.fileversionbytes << endl;
+            cout << "folders: " << v.second.foldercount << endl;
+        }
+        return;
+    }
+
     Node *n;
     TreeProcDU du;
 
--- a/examples/megacli.cpp
+++ b/examples/megacli.cpp
@@ -3343,7 +3343,7 @@ void exec_du(autocomplete::ACState& s)
         }
         for (auto& v : values)
         {
-            if (breakdown)  cout << v.first->nodehandle << endl;
+            if (breakdown)  cout << Base64Str<MegaClient::NODEHANDLE>(v.first->nodehandle) << endl;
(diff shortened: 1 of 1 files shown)


examples/megacli.cpp
```diff
@@ -3316,6 +3316,40 @@ void exec_cp(autocomplete::ACState& s)
...
+        struct counters { m_off_t filecount = 0, filebytes = 0, fileversioncount = 0, fileversionbytes = 0, foldercount = 0; };
+        map<Node*, counters> values;
+        for (auto& n : client->nodes)
+        {
+            auto& c = values[breakdown ? n.second->firstancestor() : nullptr];
+            switch (n.second->type)
+            {
+                case FILENODE:
+                {
+                    bool isversion = n.second->parent && n.second->parent->type == FILENODE;
```
cb-mega: isversion should be `const`

examples/megacli.cpp
```diff
@@ -3316,6 +3316,40 @@ void exec_cp(autocomplete::ACState& s)
 
 void exec_du(autocomplete::ACState& s)
 {
+    if (s.extractflag("-all"))
+    {
+        bool breakdown = s.extractflag("-breakdown");
```
cb-mega: breakdown should be const

examples/megacli.cpp
```diff
@@ -3316,6 +3316,40 @@ void exec_cp(autocomplete::ACState& s)
...
+        struct counters { m_off_t filecount = 0, filebytes = 0, fileversioncount = 0, fileversionbytes = 0, foldercount = 0; };
```
cb-mega: structs are pascal case

examples/megacli.cpp
```diff
@@ -3316,6 +3316,40 @@ void exec_cp(autocomplete::ACState& s)
...
+        map<Node*, counters> values;
+        for (auto& n : client->nodes)
```
cb-mega: nodePair instead of n? can it be const?

examples/megacli.cpp
```diff
@@ -3316,6 +3316,40 @@ void exec_cp(autocomplete::ACState& s)
...
+        {
+            auto& c = values[breakdown ? n.second->firstancestor() : nullptr];
```
cb-mega: counters instead of c?

examples/megacli.cpp
```diff
@@ -3316,6 +3316,40 @@ void exec_cp(autocomplete::ACState& s)
...
+                    break;
+                }
+                case FOLDERNODE:
+                {
+                    c.foldercount += 1;
+                    break;
+                }
+            }
+        }
+        for (auto& v : values)
```
cb-mega: value instead of v?

examples/megacli.cpp
```diff
@@ -3316,6 +3316,40 @@ void exec_cp(autocomplete::ACState& s)
...
 {
+    if (s.extractflag("-all"))
+    {
+        bool breakdown = s.extractflag("-breakdown");
+        struct counters { m_off_t filecount = 0, filebytes = 0, fileversioncount = 0, fileversionbytes = 0, foldercount = 0; };
+        map<Node*, counters> values;
+        for (auto& n : client->nodes)
+        {
+            auto& c = values[breakdown ? n.second->firstancestor() : nullptr];
+            switch (n.second->type)
```
cb-mega: We get a warning here because not all enum values are handled. You need to add a default case that has `assert(false)` or something like that.

examples/megacli.cpp
```diff
@@ -2571,7 +2571,7 @@ autocomplete::ACN autocompleteSyntax()
     p->Add(exec_rm, sequence(text("rm"), remoteFSPath(client, &cwd)));
     p->Add(exec_mv, sequence(text("mv"), remoteFSPath(client, &cwd, "src"), remoteFSPath(client, &cwd, "dst")));
     p->Add(exec_cp, sequence(text("cp"), remoteFSPath(client, &cwd, "src"), either(remoteFSPath(client, &cwd, "dst"), param("dstemail"))));
-    p->Add(exec_du, sequence(text("du"), remoteFSPath(client, &cwd)));
+    p->Add(exec_du, sequence(text("du"), either(sequence(flag("-all"), opt(flag("-breakdown"))), remoteFSPath(client, &cwd))));
```
cb-mega: Does that mean a remote path cannot be called `-all`?

examples/megacli.cpp
mattw-mega: it's a local and its scope is only the next two lines, and it's obviously not changed.  I don't think we gain anything by specifying that?

examples/megacli.cpp
mattw-mega: I don't think this one needs it either.    `const` is important for member functions, parameters etc, but not for local variables.

examples/megacli.cpp
mattw-mega: surely for a local struct, with hardly any scope, in a non customer facing app, it doesn't matter?

examples/megacli.cpp
mattw-mega: just following the existing style ;-)

examples/megacli.cpp
mattw-mega: likewise

examples/megacli.cpp
mattw-mega: same

examples/megacli.cpp
mattw-mega: will update

examples/megacli.cpp
mattw-mega: correct: `-all` is done by iterating the map of all nodes, rather than recursing subtrees.

examples/megacli.cpp
cb-mega: completely disagree and so do the cpp core guidelines ;)
const should be the default in C++ but unfortunately it's not and we should add it everywhere we can to make intent clear and have the compiler help you.

examples/megacli.cpp
cb-mega: All classes in the SDK are pascal case no? That's also what we agreed on when we discussed naming on slack.

examples/megacli.cpp
cb-mega: Single letter names? This practise was explicitly criticized by Pablo and I agree. We should have good, explanatory names.

examples/megacli.cpp
sergiohs84: But perhaps using `isVersion` :P

examples/megacli.cpp
sergiohs84: I agree with Christian here. We should get used to the code style conventions we agreed, even if the code belongs to an example, a test or the main core of the library :)

sergiohs84:
Is this PR ready for a final review and mere, @mattw-mega ?

PR #1561, opened 2019-07-30
#12765. Exemption of inshares in storage used

mattw-mega:


src/node.cpp
```diff
@@ -941,6 +972,20 @@ bool Node::setparent(Node* p)
...
+    bool gotnc = false;
+
+    Node *originalancestor = firstancestor();
+    handle oah = originalancestor->nodehandle;
+    if (oah == client->rootnodes[0] || oah == client->rootnodes[1] || oah == client->rootnodes[2] || originalancestor->inshare)
+    {
+        nc = subnodeCounts();
+        gotnc = true;
+
+        // nodes moving from cloud drive to rubbish for example.  Nodes should not move between inshares and owned nodes
```
sergiohs84: Should they be copy&delete instead of move? (I don't know by heart)

mattw-mega: I checked with Jon, and in fact they can move between inshares, when the inshares are from the same user.  However the changes are sent as a delete of that node handle, followed by a node tree that contains that same node handle.   The delete processing just sets a `changed.deleted` flag on the node for that handle, that is unset when processing the node tree, followed by setparent for each received node.    So the code will behave correctly, and I'll fix the comment.  thanks

src/commands.cpp
```diff
@@ -3925,6 +3925,19 @@ void CommandGetUserQuota::procresult()
                         ns->version_bytes = client->json.getint();
                         ns->version_files = client->json.getint();
 
+#ifdef _DEBUG
```
sergiohs84: Please, check if you really want to keep this block in place for a while. I have no problem with that, despite we should eventually remove it.

sergiohs84: Then we should replace the `_DEBUG` by a more standard macro, such as `NDEBUG` (defined by C standard) or simple `DEBUG` (defined by our projects). That way, not only Windows developers will run the code, but any debug build (ie. SDK/MEGAchat devs using Linux, Android and iOS devs)

PR #1563, opened 2019-07-31
Honor the requested params in `uq`

sergiohs84:
Honor the requested params (storage|transfer|pro) to the `uq` command,
rather than rely on the returned information, since depending on the
account (age & status) the API returns different sets of data.


sergiohs84:
This hotfix is finally not needed for the release, since the API has already fixed it at server-side. However, it's good to make our code more robust, so it will be merged into develop.

PR #1569, opened 2019-08-08
Fix/double delete logging out during folder upload

mattw-mega:


include/megaapi_impl.h
```diff
@@ -172,11 +172,20 @@ class MegaSizeProcessor : public MegaTreeProcessor
         long long getTotalBytes();
 };
 
-class MegaFolderUploadController : public MegaRequestListener, public MegaTransferListener
+class MegaRecursiveTransferOpertaion
```
polmr: This interface could be renamed to a more generic MegaRecursiveOperation (btw, there's a typo now: `Opertaion`)

tests/integration/SdkTest_test.cpp
```diff
@@ -4065,3 +4069,57 @@ TEST_F(SdkTest, DISABLED_SdkGetRegisteredContacts)
...
+    if (fs::exists(p))
+    {
+        fs::remove_all(p);
+    }
+    fs::create_directories(p);
+    ASSERT_TRUE(buildLocalFolders(p.u8string().c_str(), "newkid", 3, 2, 10));
+
+    // start uploading
+    megaApi[0]->startUpload(p.u8string().c_str(), megaApi[0]->getRootNode());
+    WaitMillisec(500);
```
cb-mega: As discussed IRL, we should use values collected from listeners as opposed to sleeping.

src/megaapi_impl.cpp
```diff
@@ -2739,6 +2739,12 @@ void MegaTransferPrivate::setListener(MegaTransferListener *listener)
     this->listener = listener;
 }
 
+void MegaTransferPrivate::startRecursiveOperations(MegaRecursiveTransferOpertaion* op)
```
cb-mega: As discussed IRL, this function should take `unique_ptr` to make ownership clear

include/megaapi_impl.h
mattw-mega: thanks, renamed (with typo corrected :-) ) cheers

include/megaapi_impl.h
```diff
@@ -172,11 +172,20 @@ class MegaSizeProcessor : public MegaTreeProcessor
...
+{
+public:
+    virtual ~MegaRecursiveTransferOpertaion() {}
```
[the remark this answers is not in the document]
mattw-mega: done

tests/integration/SdkTest_test.cpp
```diff
@@ -4065,3 +4069,57 @@ TEST_F(SdkTest, DISABLED_SdkGetRegisteredContacts)
     ASSERT_EQ(js2, std::get<2>(table[1])); // ud
 }
 
+
+bool buildLocalFolders(fs::path targetfolder, const string& prefix, int n, int recurselevel, int filesperfolder)
```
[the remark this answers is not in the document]
mattw-mega: maybe in future when it's a bit more mature - I've namespaced it for now :-)  cheers

tests/integration/SdkTest_test.cpp
```diff
@@ -4065,3 +4069,57 @@ TEST_F(SdkTest, DISABLED_SdkGetRegisteredContacts)
...
+{
+    // this one used to cause a double-delete
+
+    // make new folders (and files) in the local filesystem - approx 90 
+    fs::path p = fs::current_path() / "uploadme_mega_auto_test_sdk";
+    if (fs::exists(p))
+    {
+        fs::remove_all(p);
+    }
+    fs::create_directories(p);
```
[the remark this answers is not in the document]
mattw-mega: I've added cleanup code - the problem is that if the test fails, the cleanup doesn't happen, so it needs to pre-cleanup in case the last run failed.  thanks

src/megaapi_impl.cpp
```diff
@@ -7162,17 +7172,11 @@ void MegaApiImpl::abortPendingActions(error preverror)
...
-        }
-    }
-    for (auto transfer : transfers)
+    while (!transferMap.empty())
     {
+        auto transfer = transferMap.begin()->second;
         transfer->setState(MegaTransfer::STATE_FAILED);
+        // this call can be recursive and remove multiple, eg with MegaFolderUploadController
         fireOnTransferFinish(transfer, preverror);
     }
```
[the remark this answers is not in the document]
mattw-mega: added

tests/integration/SdkTest_test.cpp
mattw-mega: done :-)

src/megaapi_impl.cpp
mattw-mega: let's discuss with the team and see if we can find a way to do it that doesn't add too much syntax :-)

sergiohs84: I also prefer a clean syntax here. Overloading the code with (syntax) decorations is not always worthy if you are careful enough and add a comment about ownership close to the declaration of the function. Nevertheless, I understand the advantage of the `unique_ptr` pointed by Christian as well.

mattw-mega: If we add our own version of the make_unique function (which was added to the c++14 standard, but is just a (Ivaridically) templated function that would work in c++11 - and let's just add the most common case), that would allow us to cover cases like this with compiler-checked/managed ownership (and zero runtime overhead) without adding too much syntax - I'll try it and see how it looks - I'll post it in the cpp channel if it seems ok.

cb-mega: `unique_ptr` is not just syntactic sugar: The compiler will be your friend and help you get ownership right. When the function takes a `unique_ptr` it is clear that the function will own the memory; no comments needed.

tests/integration/SdkTest_test.h
```diff
@@ -62,6 +64,50 @@ class MegaLoggerSDK : public MegaLogger {
     void log(const char *time, int loglevel, const char *source, const char *message);
 };
 
+struct TransferTracker : public ::mega::MegaTransferListener
+{
+    std::atomic<bool> started = false;
```
sergiohs84: It seems that Jenkins complains about this initialization. It happened to me as well in my first usage of `atomic_bool` and I solved it by using the initialization list:
```suggestion
    std::atomic<bool> started = {false};
```

PR #1570, opened 2019-08-08
Fix business accounts initialization

jgandres:
- Init mBizStatus to BIZ_STATUS_UNKNOWN
- Notify apps when we receive an invalid business status
- Nofify apps a new status change upon `ug` response after update
business mode and expiration & grace period flags

### DIFF
--- a/src/commands.cpp
+++ b/src/commands.cpp
@@ -3672,8 +3672,14 @@ void CommandGetUserData::procresult()
                 if ((s < BIZ_STATUS_EXPIRED || s > BIZ_STATUS_GRACE_PERIOD)  // status not received or invalid
                         || (m == BIZ_MODE_UNKNOWN))  // master flag not received or invalid
                 {
-                    LOG_err << "GetUserData: invalid business status / account mode";
-                    client->app->notify_business_status(BIZ_STATUS_UNKNOWN);
+                    std::string err = "GetUserData: invalid business status / account mode";
+                    LOG_err << err;
+                    client->sendevent(99450, err.c_str());
+
+                    client->mBizStatus = BIZ_STATUS_EXPIRED;
+                    client->mBizMode = BIZ_MODE_SUBUSER;
+                    client->mBizExpirationTs = client->mBizGracePeriodTs = 0;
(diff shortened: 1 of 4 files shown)


src/commands.cpp
```diff
@@ -3673,16 +3673,10 @@ void CommandGetUserData::procresult()
                         || (m == BIZ_MODE_UNKNOWN))  // master flag not received or invalid
                 {
                     LOG_err << "GetUserData: invalid business status / account mode";
+                    client->app->notify_business_status(BIZ_STATUS_UNKNOWN);
```
sergiohs84: Check with the apps teams the expected behavior for an unknown business status.

Also, if an unknown status is notified, it must update the client's state: `MegaClient::mBizStatus` and the related class members as well.

Perhaps, a possible workaround would be fallback to `BIZ_STATUS_INACTIVE` and send an event to the stats server to record the unexpected data received from API.

src/megaclient.cpp
```diff
@@ -1049,7 +1049,7 @@ void MegaClient::init()
     mNotifiedSumSize = 0;
 
     mBizMode = BIZ_MODE_UNKNOWN;
-    mBizStatus = BIZ_STATUS_INACTIVE;
+    mBizStatus = BIZ_STATUS_UNKNOWN;
```
sergiohs84: If the `BIZ_STATUS_UNKNOWN` will be exposed to apps, it needs to be added to the documentation in `MegaApi`

sergiohs84:
Also note that, in case of default initialization of `mBizStatus = BIZ_STATUS_UNKNOWN`, the `MegaApi::isBusinessAccount()` will return `true` when it's actually not known yet.

src/commands.cpp
sergiohs84: Agreed with Webclient that the fallback status in case of invalid data in the `ug` response for the `b` element is `BIZ_STATUS_EXPIRED`, since that status will keep the business level for the account, but will block most of the operations.

src/megaclient.cpp
sergiohs84: Fixed

PR #1571, opened 2019-08-12
Allow to cancel ongoing searches

sergiohs84:
Use a flag to abort ongoing searches, since for large accounts they can consume too much time and the results may become obsolete (a more accurate search is already launched or the user has moved to a different view).

### DIFF
--- a/include/megaapi.h
+++ b/include/megaapi.h
@@ -16251,11 +16251,10 @@ public:
 class MegaCancelToken
 {
 public:
-public:
 
     /**
      * @brief Creates an object which can be passed as parameter for some MegaApi methods in order to
-     * cancel the processing associated to the function. @see MegaApi::search
+     * request the cancellation of the processing associated to the function. @see MegaApi::search
      *
      * You take ownership of the returned value.
      *
--- a/include/megaapi.h
+++ b/include/megaapi.h
@@ -12957,7 +12957,8 @@ class MegaApi
          * You take the ownership of the returned value.
          *
          * This function allows to cancel the processing at any time by passing a MegaCancelToken and calling
-         * to MegaCancelToken::setCancelFlag(true).
+         * to MegaCancelToken::setCancelFlag(true). If a valid object is passed, it must be kept alive until
+         * this method returns.
          *
          * @param node The parent node of the tree to explore
          * @param searchString Search string. The search is case-insensitive
@@ -13068,7 +13069,8 @@ class MegaApi
          *  - Incoming shares from other users
          *
          * This function allows to cancel the processing at any time by passing a MegaCancelToken and calling
-         * to MegaCancelToken::setCancelFlag(true).
+         * to MegaCancelToken::setCancelFlag(true). If a valid object is passed, it must be kept alive until
+         * this method returns.
          *
          * You take the ownership of the returned value.
          *
--- a/include/megaapi.h
+++ b/include/megaapi.h
@@ -13003,6 +13003,14 @@ class MegaApi
         MegaNodeList* search(const char* searchString, int order = ORDER_NONE);
 
         /**
+         * @brief Allows to cancel an ongoing search
+         *
+         * Since searches are blocking, if the user refines the search string, the current search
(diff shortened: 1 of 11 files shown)


mattw-mega:
Looks good, though it will cancel all searches in progress - it's possible some app might have multiple windows open, viewing different folders, and the user starts a search in each.   I think ideally we would have a way to cancel a specific request, perhaps a generic cancelRequest, maybe by request tag (which would then have to be assigned on initial request, rather than once it gets to the MegaApi's thread).  thanks

sergiohs84:
> Looks good, though it will cancel all searches in progress - it's possible some app might have multiple windows open, viewing different folders, and the user starts a search in each. I think ideally we would have a way to cancel a specific request, perhaps a generic cancelRequest, maybe by request tag (which would then have to be assigned on initial request, rather than once it gets to the MegaApi's thread). thanks

Thanks for the feedback, @mattw-mega 

The SDK is not able to perform multiple searches at the same time, is it? Since the search call acquires the `sdkMutex`, no more threads can acquire it, so it will cancel the only search in progress. Correct me if I'm missing something, please. (Of course, we can implement searches in parallel in a future, but that's not currently supported AFAIK).

On the other hand, allowing to cancel any request would be a nice feature, but remember that `MegaApi::search()` is a synchronous call, not an asynchronous request :)

mattw-mega:
> On the other hand, allowing to cancel any request would be a nice feature, but remember that `MegaApi::search()` is a synchronous call, not an asynchronous request :)

Yes, good point.   I was thinking that was an async request.   However, even in the synchronous case it's important to tie the cancel to the specific search to be cancelled.    We could have an app with two file-explorer type windows (each with their own thread), viewing different folders in the same MEGA account.  And the user starts a search in each one (and maybe the search takes time to run, and the 2nd one has to wait for the first to finish of course).  And then the user decides not to wait anymore, they cancel the search in window 2, but it's actually window 1's search that gets cancelled.

Or another scenario is just a single window but with a race on the cancel - the search finishes naturally just as the user clicks cancel.   The thread that was searching has already moved on to another operation, and that one gets cancelled instead (as I see there are several different MegaApi calls that use the `processTree()` function, not just search)?   thanks

include/megaapi.h
```diff
@@ -13002,6 +13057,64 @@ class MegaApi
...
+         *
+         * - MegaApi::ORDER_ALPHABETICAL_ASC = 9
+         * Sort in alphabetical order, ascending
+         *
+         * - MegaApi::ORDER_ALPHABETICAL_DESC = 10
+         * Sort in alphabetical order, descending
+         *
+         * @return List of nodes that contain the desired string in their name
+         */
+        MegaNodeList* search(const char* searchString, MegaCancelToken *cancelToken, int order = ORDER_NONE);
```
mattw-mega: Looks great - just one thing we might want to support - if thread B calls cancel(), and thread A was just finishing search() anyway, there's a race condition as to whether the search finished naturally, or whether it noticed the flag in time, and skipped some work.   There's no way for the caller to know which of those occurred, and the caller may want to still display the results as fully valid, if the user effectively clicked a cancel button too late.    Perhaps the MegaCancelToken could also contain a flag as to whether the cancel request had any effect?     Or maybe that's too much detail for this PR - it's certainly looking good anyway.  cheers

sergiohs84: It's certainly another good improvement for the future, but I think the benefit in this case, the `search()` sync call, it's very little. Furthermore, it looks to me that the management of whether the call completed its work before or after the cancellation could require to be specific to each case (it depends on what the app has requested, it's not the same a search than a request to the API, ie.). So let's keep it simple by now.
Thanks for the feedback

include/megaapi.h
```diff
@@ -16135,6 +16248,36 @@ class MegaAchievementsDetails
     virtual long long currentTransferReferrals();
 };
 
+class MegaCancelToken
+{
+public:
+public:
```
jnavarrom:
```suggestion
```

PR #1575, opened 2019-08-19
Address precision warnings

jnavarrom:


src/commands.cpp
```diff
@@ -3723,7 +3723,7 @@ void CommandGetUserData::procresult()
                     }
                     if (auxts)
                     {
-                        dstime diff = (now - auxts) * 10;
+                        dstime diff = dstime((now - auxts) * 10);
```
sergiohs84: Static cast syntax is preferred:
```suggestion
                        dstime diff = static_cast<dstime>((now - auxts) * 10));
```

PR #1586, opened 2019-08-28
Fix de-serialization of the plink's node handle

cb-mega:
This ensures that, when `Node` is de-serialized, the node handle that `plink` carries is correctly de-serialized. Previously, the code may generate the wrong node handle as it would read 8 instead of 6 bytes.

### DIFF
--- a/src/node.cpp
+++ b/src/node.cpp
@@ -269,6 +269,7 @@ Node* Node::unserialize(MegaClient* client, string* d, node_vector* dp)
         ph = UNDEF;
     }
 
+    u = 0;
     memcpy((char*)&u, ptr, MegaClient::USERHANDLE);
     ptr += MegaClient::USERHANDLE;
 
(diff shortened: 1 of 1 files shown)


src/node.cpp
```diff
@@ -269,6 +269,7 @@ Node* Node::unserialize(MegaClient* client, string* d, node_vector* dp)
         ph = UNDEF;
     }
 
+    u = 0;
```
sergiohs84: This initialization is not needed, you are copying 8 bytes in any case.

src/node.cpp
```diff
@@ -394,7 +395,8 @@ Node* Node::unserialize(MegaClient* client, string* d, node_vector* dp)
             return NULL;
         }
 
-        handle ph = MemAccess::get<handle>(ptr);
+        handle ph = 0;
+        memcpy((char*)&ph, ptr, MegaClient::NODEHANDLE);
```
sergiohs84: Good catch, indeed!! I wonder how it happened that we never got any report about this issue, since the `ets` the handle would not match the actual public handle. Probably it's because none uses the PRO feature to set the expire timestamps for public links :( (so it's 0 and the handle gets its higher bytes set to zero by coincidence).

PR #1588, opened 2019-08-28
#13079. Generate links with new format

alber2510:


include/mega/megaclient.h
```diff
@@ -229,6 +229,9 @@ class MEGA_API MegaClient
     // pseudo-random number generator
     PrnGen rng;
 
+    // Use new format to generate Mega links
+    static bool newLinkFormat;
```
sergiohs84: Better to use a non-static member to avoid issues with multiple instances.

src/commands.cpp
```diff
@@ -3634,6 +3635,10 @@ void CommandGetUserData::procresult()
             }
             break;
 
+        case MAKENAMEID4('n', 'l', 'f', 'e'):
```
sergiohs84: It needs to be moved into the `flags` section.

src/commands.cpp
```diff
@@ -3634,6 +3635,10 @@ void CommandGetUserData::procresult()
...
+            newLinkFormat = client->json.getbool();
```
sergiohs84: Let's parse it as an integer, and do a casting to bool

src/megaapi_impl.cpp
```diff
@@ -879,19 +879,29 @@ char *MegaNodePrivate::getPublicLink(bool includeKey)
         return NULL;
     }
 
-    string strlink = "https://mega.nz/#";
-    strlink += (type ? "F" : "");
+    string strlink = "https://mega.nz/";
```
sergiohs84: It might worth to move this part into an static method in `std::string& MegaClient::getPublicLink(type, ph, key)` and use it as well in `megacli` at `exportnode_result()`

sergiohs84:
Additionally, the `MegaClient::openfilelink()` needs to be adjusted to allow both types of link, the new one and the old one. Same applies to `parsefolderlink()`.

src/megaapi_impl.cpp
sergiohs84: Also, check `megacli` can log into folder links of new format. Download file links should work fine once the `openfilelink()` is adjusted.

src/megaclient.cpp
```diff
@@ -1211,6 +1211,35 @@ MegaClient::~MegaClient()
...
+    else
+    {
+        nodeType = (type ? "#!" : "#F!");
+    }
+
+    strlink += nodeType;
+
+    char *base64ph = new char[12];
+    Base64::btoa((byte*)&(ph), MegaClient::NODEHANDLE, base64ph);
+    strlink += base64ph;
```
sergiohs84: Not a biggie, but here we should start using the new
```Base64Str<NODEHANDLE>((const byte*)&ph)```

File: src/megaclient.cpp [hunk 6]
```diff
@@ -9826,7 +9869,9 @@ error MegaClient::openfilelink(const char* link, int op)
     {
         ptr += 8;
 
-        if (*ptr == '!')
+        int a = strlen(ptr);
+        int b = sizeof key;
```
sergiohs84: Are the above variables for testing leftovers? :)

src/megaclient.cpp
```diff
@@ -1211,6 +1211,35 @@ MegaClient::~MegaClient()
...
+    strlink += base64ph;
+    strlink += (newLinkFormat ? "#" : "");
+    delete [] base64ph;
+
+    if (key)
+    {
+        strlink += (newLinkFormat ? "" : "!");
+        strlink += key;
+    }
+
```
sergiohs84: I miss here a `return strlink;`, don't you?

src/megaclient.cpp
alber2510: Fixed it

src/megaclient.cpp
alber2510: Fixed it

src/megaclient.cpp
alber2510: Fixed it

PR #1590, opened 2019-09-04
#13143. Fix issues with new filesystem partition layout of macOS 10.15

vtmateos:
Due to new configuration of system partitions that comes with Catalina (operating system is now on a separate partition from user’s data), filesystem notifications from `fsevents` come with a new prefix `/System/Volumes/Data` causing a failure during matching with `localroot.localname` of the syncs, disallowing the detection of changes at all.


include/mega/utils.h
```diff
@@ -362,6 +362,9 @@ std::string webdavnameescape(const std::string &value);
 
 void tolower_string(std::string& str);
 
+#ifdef __APPLE__
+int macOSmajorVersion();
```
sergiohs84: I think it's easier to read this way:
```suggestion
    int macOSmajorVersion();
```

src/sync.cpp
```diff
@@ -95,6 +95,36 @@ Sync::Sync(MegaClient* cclient, string* crootpath, const char* cdebris,
...
+    if (macOSmajorVersion() >= 19) //macOS catalina+
+    {
+        LOG_debug << "macOS 10.15+ filesystem detected. Checking fseventspath.";
+        string supercrootpath = "/System/Volumes/Data" + *crootpath;
+
+        int fd = open(supercrootpath.c_str(), O_RDONLY);
+        if (fd == -1)
+        {
+            LOG_debug << "Unable to open path using fseventspath.";
+            this->fsEventsPath = *crootpath;
```
sergiohs84: New class members should be named with a `m` prefix. Please, refactor `fsEventsPath` to `mFsEventsPath` or alike (and you should skip also the usage of `this->` to access the member (we avoid clashes with other variables thanks to the prefix).

include/mega/utils.h
vtmateos: Other ifdefs within same file does not apply padding, so that's the reason to keep in the same way.

src/sync.cpp
vtmateos: Will do. Thanks.

PR #1591, opened 2019-09-04
#13079. Add support for new link format when password-protected

sergiohs84:
It fixes the password protected links, which won't support the new format with the current code.


sergiohs84:
trigger compilation

sergiohs84:
Jenkins failure is unrelated to this PR, but a known bug of the `MegaFolderUploadController`.
Ready to merge

PR #1592, opened 2019-09-05
#13150. Fix init business flags

jgandres:
Change business flags initialization to avoid unnecessary notifications to apps.


jgandres:
Risk area/s:
 - Management of business flags mode and status
 - App initialization in order to obtain right values for business flags mode and status
 - Notification to apps through an event that business mode has changed

PR #1594, opened 2019-09-06
#12674. Add target to u command

jgandres:
By adding target parameter in `u` command API will response with API_EOVERQUOTA in case the destination storage is in overquota.

Risk area/s:
    - Upload the same file to multiple destinations, including inshares
    - Transfer failed due to overquota of your own account (cloud-drive)
    - Transfer failed due to overquota of the target account (inshare)


src/commands.cpp
```diff
@@ -324,6 +324,15 @@ CommandPutFile::CommandPutFile(MegaClient* client, TransferSlot* ctslot, int ms)
...
     arg("ms", ms);
+
+    for (auto &file : tslot->transfer->files)
+    {
+        if (!ISUNDEF(file->h))
+        {
+            arg("t", (byte*)&file->h, MegaClient::NODEHANDLE);
+            break;
+        }
+    }
```
mattw-mega: If I upload the same large file to multiple folders, it's just one transfer with multiple entries in the files list.   However some targets could be in the user's account, some could be in inshares.   Also multiple targets in the user's own account could take it overquota when only one would not take it overquota.   Shouldn't we take all the entries in the files list into account?   thanks

src/transfer.cpp
```diff
@@ -356,27 +356,38 @@ void Transfer::failed(error e, dstime timeleft)
...
+            {
+                    Node *node = client->nodebyhandle(slot->transfer->target);
+                    if (client->rootnodes[0] == client->getrootnode(node)->nodehandle)
+                    {
+                        bt.backoff(timeleft ? timeleft : NEVER);
+                        client->activateoverquota(timeleft);
+                    }
+                    else
+                    {
+                        client->app->transfer_failed(this, API_EOVERQUOTA, timeleft);
```
sergiohs84:
```suggestion
                        client->app->transfer_failed(this, e, timeleft);
```

src/megaapi_impl.cpp
```diff
@@ -16047,6 +16058,16 @@ void MegaApiImpl::processTransferFailed(Transfer *tr, MegaTransferPrivate *trans
     transfer->setLastError(megaError);
     transfer->setPriority(tr->priority);
     transfer->setState(MegaTransfer::STATE_RETRYING);
+
+    if (e == API_EOVERQUOTA)
+    {
+        unique_ptr<MegaNode> targetNode(getNodeByHandle(tr->target));
```
sergiohs84: In order to avoid a deleted node for storage overquota error, it would be better to store the `mForeignTarget` at the `Trasfer`, instead of the nodehandle, which requires lookups.

include/megaapi_impl.h
```diff
@@ -743,6 +745,7 @@ class MegaTransferPrivate : public MegaTransfer, public Cachable
         long long endPos;
         int retry;
         int maxRetries;
+        bool foreignOverquota;
```
sergiohs84: Rather than declaring a new member (remember, prefix them with `m`), you should be able to recycle an existing one such as `streamingTransfer` and add the corresponding documentation to the related method/s

sergiohs84:
This PR will be in standby until API adds the ability to send an array of target handles.

sergiohs84:
This PR will be in standby until API deploys the ability to send an array of target handles.

sergiohs84:
> This PR will be in standby until API adds the ability to send an array of target handles.

API change is live. We can proceed with merge. Please, @mattw-mega would you mind to review it? with the latest changes for profiling, there was a conflict in `transfer.cpp` that I'd like you to confirm the result is correct.

Thanks

src/commands.cpp
```diff
@@ -324,6 +324,33 @@ CommandPutFile::CommandPutFile(MegaClient* client, TransferSlot* ctslot, int ms)
...
+    for (auto &file : tslot->transfer->files)
+    {
+        if (!ISUNDEF(file->h))
+        {
+
+            Node *node = client->nodebyhandle(file->h);
+            if (node)
+            {
+                handle rootnode = client->getrootnode(node)->nodehandle;
+                tslot->transfer->mForeignTarget = client->rootnodes[0] != rootnode;
```
mattw-mega: mForeignTarget will be set according to the last processed `file` in the list, without taking into account that other files in the list might be 'foreign' or not.   Wouldn't it be better to set it according to any-are-foreign or all-are-foreign so that the order of the files in the list does not matter?   thanks

src/megaapi_impl.cpp
```diff
@@ -16178,6 +16178,7 @@ void MegaApiImpl::processTransferFailed(Transfer *tr, MegaTransferPrivate *trans
     transfer->setLastError(megaError);
     transfer->setPriority(tr->priority);
     transfer->setState(MegaTransfer::STATE_RETRYING);
+    transfer->setStreamingTransfer(tr->mForeignTarget);
```
mattw-mega: 'streaming transfer' refers to whether the transfer was started with a call to `MegaApi::startStreaming`.   I think we should make a new appropriately named getter rather than overloading this setter/getter with a meaning that doesn't match the name?   Also overquota errors could occur for streaming or non-streaming in which case we won't know which of those reasons this flag was set?  thanks

mattw-mega:
Thinking about that mForeignTarget flag a bit more, perhaps it should also be updated when additional files are added to the transfer while the upload is going on?    Or, it would be much simpler to calculate that flag on-the-fly when it's used in Transfer::failed (probably it should be true if none of the files are in the user's own account?), and then we don't even need to keep it as state?  thanks

src/megaapi_impl.cpp
jgandres: Hi, Matt, we try to reuse class members if it's possible in order to not make this kind of classes bigger. This error only occurs for not streaming transfers so we thought we could use this member to indicate the error.

sergiohs84: Yes, we try to reuse existing flags / getter inside `MegaRequest` and alike, since the class is multi-purpose and adding a new getter or a new member for every different usage would result on a even longer list.

In this case, I suggested the _streaming-transfer_ since it's a flag and it assumed it was not in use for uploads (only downloads can be streaming. However, a quick search of `MegaTransfer::isStreamingTransfer()`... and boom, it's used at `MegaApiImpl::getStreamingTransfers()`, so we would start returning uploads in the set of streaming transfers --> bug!

Please, @jgandres , proceed with the addition of a new flag + setter/getter and adjust the code accordingly, please (and the documentation). Thanks

jgandres: Sure, I'll adjust now

src/commands.cpp
sergiohs84: You're right, @mattw-mega . I thought it was managed on a per-`File` basis, rather than per `Transfer`. Since we want to adjust the behavior of failed transfers for `u` with -17 based on the set of target handles, we do need to know if all-are-foreign, if there are mixed foreign-plus-own handles or all were own handles at the time the `u` is sent, and do not retry transfers with foreign target if the `u` failed with -17. If there was a mix, then keep retrying the ones with targets in you own account, but permanent fail the foreign.

sergiohs84:
> Thinking about that mForeignTarget flag a bit more, perhaps it should also be updated when additional files are added to the transfer while the upload is going on? Or, it would be much simpler to calculate that flag on-the-fly when it's used in Transfer::failed (probably it should be true if none of the files are in the user's own account?), and then we don't even need to keep it as state? thanks

When additional files are added to the transfer while the upload is going on, the `u` has already returned a valid upload ticket, so none of the target accounts should be overquota at that moment. In consequence, the new transfer should only fail with overquota (1) for bandwidth --> temporary error, or (2) at the `p` stage / putnodes, and that is handled as a permanent error, IIRC. I'd need to check.

src/transfer.cpp
```diff
@@ -356,29 +356,78 @@ void Transfer::failed(error e, dstime timeleft)
...
+            else
+            {
+                bool allForeignTargets = true;
+                for (file_list::iterator it = files.begin(); it != files.end(); it++)
+                {
+                    Node *node = client->nodebyhandle((*it)->h);
+                    if (node)
+                    {
+                        handle rootnode = client->getrootnode(node)->nodehandle;
+                        if (client->rootnodes[0] == rootnode)
```
mattw-mega: Probably we should compare against all 3 root nodes here, as it's technically possible to upload/download from //bin and //in node trees.    Perhaps we should have a function `bool MegaClient::isAccountRootNode(handle)` to make that check easy?  thanks

src/transfer.cpp
```diff
@@ -356,29 +356,78 @@ void Transfer::failed(error e, dstime timeleft)
...
+        // Remove files with foreign targets, if transfer failed with a (foreign) storage overquota
+        auto auxit = it++;
+        Node *node = client->nodebyhandle((*auxit)->h);
+        if (e == API_EOVERQUOTA && !timeleft && node)
         {
-            bt.backoff(timeleft ? timeleft : NEVER);
-            client->activateoverquota(timeleft);
-            if (!slot)
+            handle rootnode = client->getrootnode(node)->nodehandle;
+            if (client->rootnodes[0] != rootnode)
```
mattw-mega: and same again here?  thanks

src/transfer.cpp
jgandres: You are right, I've created the method isAccountRootNode to check it.

src/commands.cpp
```diff
@@ -324,6 +324,30 @@ CommandPutFile::CommandPutFile(MegaClient* client, TransferSlot* ctslot, int ms)
...
+    set<handle> targetRoots;
+    beginarray("t");
+    for (auto &file : tslot->transfer->files)
+    {
+        if (!ISUNDEF(file->h))
+        {
+            Node *node = client->nodebyhandle(file->h);
+            if (node)
+            {
+                handle rootnode = client->getrootnode(node)->nodehandle;
```
sergiohs84: @mattw-mega do you think we should avoid to scale the tree up to the rootnode several times for each file? It is done here, but also when the transfer fails. We could save whether the `Transfer::file` is a foreign one or not by adding a `bool isForeign` to the `File` object. We would also need to restore the status upon serialization, or add it to the cached data.
I've checked and only `rename()` seems to use the `getrootnode()`, despite I guess there are plenty of code going upwards to the rootnode for multiple purposes, so probably it is not worthy. If you feel the same way, I think we can merge

mattw-mega: I think that's a very minor cost, I wouldn't worry about it unless it shows up when profiling in some scenario.  I think it's fine to merge (I did spot one other thing which I'll comment, but it's not a showstopper, could be adjusted anytime later on)

src/megaclient.cpp
```diff
@@ -549,6 +549,23 @@ Node *MegaClient::getrootnode(Node *node)
...
+        return false;
+    }
+
+    handle rootnode = getrootnode(node)->nodehandle;
+    return (rootnode == rootnodes[0] || rootnode == rootnodes[1] || rootnode == rootnodes[2]);
+}
+
+bool MegaClient::isForeignNode(handle h)
+{
+    return !isPrivateNode(h);
```
mattw-mega: this one will return true if the node for the handle can't be found, though - shouldn't it return false in that case?  thanks

sergiohs84: Good catch!! thanks, we will adjust it and merge. It's a minor thing :)

PR #1598, opened 2019-09-11
#13160. Force a new file upload if thumbnail or preview are missing

alber2510:
Add new method to force upload new file if node is duplicated and thumbnail or preview are missing

Risk area/s:
 - Upload duplicates files

### DIFF
--- a/include/megaapi.h
+++ b/include/megaapi.h
@@ -10577,10 +10577,12 @@ class MegaApi
         void startUpload(const char* localPath, MegaNode* parent, const char* fileName, int64_t mtime, MegaTransferListener *listener = NULL);
 
         /**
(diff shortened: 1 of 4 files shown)


sergiohs84:
> You are welcome :)
> 
> I had overlooked this is specific to the chat and/or an special/edge case, so likely no big deal i'd say...at least for now ;)

Yes, Diego. The issue is that CUv2 may create the node while the thumbnail/preview are not ready. In that case, the image might have missing file attributes, but it's expected that sooner or later the CUv2 itself puts the missing attributes. That's why we discarded the idea of automatically restore the missing attributes from the local file as result of the `startUpload()`, since it may end in overwritten fa (and the attachment in chat with a broken/obsolete preview/thumbnail again).

PR #1604, opened 2019-09-18
#13250 Add performance mode to logging

cb-mega:
This is to facilitate high performance logging.

Performance mode can be enabled at compile time via defining:
`ENABLE_LOG_PERFORMANCE`

In performance mode, the `SimpleLogger` does not lock mutexes nor does it heap-allocate.
Only `loglevel` and `message` of `Logger`/`MegaLogger` are populated where `message` will include file/line. It assumed that the subclass of `Logger`/`MegaLogger` provides timing information.

In performance mode, output streams (via `addOutput` or `setAllOutputs`) are not supported.

In addition, this change removes the `LOGn_*` macros because they were using a `SimpleLogger` constructor that doesn't exist, hence, those macros are unused.


include/megaapi_impl.h
```diff
@@ -2511,13 +2513,13 @@ class MegaApiImpl : public MegaApp
         void enableGeolocation(MegaRequestListener *listener = NULL);
         void isGeolocationEnabled(MegaRequestListener *listener = NULL);
         bool isChatNotifiable(MegaHandle chatid);
+#endif
```
cb-mega: Needed to move this endif down to be able to compile without chat enabled.

include/mega/logging.h
```diff
@@ -109,41 +119,223 @@ enum LogLevel {
...
+    }
+
 public:
     static Logger *logger;
 
     static enum LogLevel logCurrentLevel;
 
-    SimpleLogger(enum LogLevel ll, char const* filename, int line);
-    ~SimpleLogger();
+    SimpleLogger(const enum LogLevel ll, const char* filename, const int line)
```
cb-mega: Constructor moved to header for inlining

include/mega/logging.h
```diff
@@ -109,41 +119,223 @@ enum LogLevel {
...
+        std::ostringstream oss;
+        oss << filename;
+        if(line >= 0)
+        {
+            oss << ":" << line;
+        }
+        fname = oss.str();
+    }
+
+    ~SimpleLogger()
```
cb-mega: Destructor moved to header for inlining

include/mega/logging.h
```diff
@@ -199,66 +431,48 @@ class SimpleLogger {
...
+    }
+
     // Synchronizes all registered stream buffers with their controlled output sequence
     static void flush();
 };
 
-// output VERBOSE log with line break
 #define LOG_verbose \
-    if (SimpleLogger::logCurrentLevel < logMax) ;\
+    if (::mega::SimpleLogger::logCurrentLevel < ::mega::logMax) ;\
```
cb-mega: Added explicit namespace here to ensure that one can call these macros without issuing a `using namespace mega` first.

include/mega/logging.h
```diff
@@ -158,24 +350,64 @@ class SimpleLogger {
...
-            ostr << "(NULL)";
+            ostr << null;
 
         return *this;
     }
 
-    template <typename T>
-    SimpleLogger& operator<<(T const& obj)
+    template <typename T, typename = typename std::enable_if<std::is_arithmetic<T>::value>::type>
+    SimpleLogger& operator<<(const T obj)
```
cb-mega: Arithmetic values (int, double, etc) should be passed by value and not by ref.

examples/megacli.cpp
```diff
@@ -6264,7 +6264,15 @@ void DemoApp::exportnode_result(handle h, handle ph)
             return;
         }
 
-        const char *key = (n->type == FILENODE) ? Base64Str<FILENODEKEYLENGTH>((const byte*)n->nodekey.data()) : Base64Str<FOLDERNODEKEYLENGTH>(n->sharekey->key);
+        const char* key;
```
cb-mega: This hunk fixes the build on windows.

include/mega/logging.h
```diff
@@ -109,41 +119,223 @@ enum LogLevel {
...
+    std::array<char, 256>::iterator mBufferIt; // used in performance mode
+
     // logging can occur from multiple threads, so we need to protect the lists of loggers to send to
     // though the loggers themselves are presumed to be owned elsewhere, and the pointers must remain valid
     // actual output to the loggers is not synchronised (at least, not by this class)
     static std::mutex outputs_mutex;
     static OutputMap outputs;
     static OutputStreams getOutput(enum LogLevel ll);
 
+    static bool performanceMode;
```
polmr: = false;
missing?

include/mega/logging.h
```diff
@@ -109,41 +119,223 @@ enum LogLevel {
...
+    logValue(const T value)
+    {
+        char str[20];
+        const auto size = std::sprintf(str, "%f", value);
+        copyToBuffer(str, size);
+    }
+
+    void logValue(const char* value)
+    {
+        copyToBuffer(value, std::strlen(value));
```
polmr: I'm wondering if we should consider implementing this by looping until `\0`, to avoid one extra loop (strlen) before the reading loop that copyToBuffer entails. We should gain a little bit in performance and this should be (one of ) the most used cases. Shouldn't it?

include/mega/logging.h
```diff
@@ -158,24 +350,64 @@ class SimpleLogger {
     template <typename T>
     SimpleLogger& operator<<(T* obj)
     {
+        const char* null = "(NULL)";
```
polmr: could this go static into SimpleLogger?

include/mega/logging.h
cb-mega: Yes, good point. Will change accordingly.

include/mega/logging.h
cb-mega: It's initialized in the cpp. But I agree, we should make this a compile time decision so we avoid the runtime checks (Matt also agrees). How about we call this: `ENABLE_LOG_PERFORMANCE_MODE` to indicate it's related to logging?

include/mega/logging.h
cb-mega: This is just a stack-based pointer to a static string so should be fine, no? This is essentially free code AFAIK because the compiler knows the address of `null` at compile time, which just references a string from the static storage section of the binary.

include/mega/logging.h
cb-mega: Actually, I am not convinced anymore: `std::strlen` and `std::copy` will be highly optimized and possibly vectorized. Also, the check `if (mBufferIt == mBuffer.end() - 1)` only has to be done after the bulk copy but if we copy each element then this check has to be done every time.
What do you think?

Btw, this is how I'd implement what you suggested:
```cpp
    void logValue(const char* value)
    {
        while (*value)
        {
            *mBufferIt = *value;
            ++mBufferIt;
            if (mBufferIt == mBuffer.end() - 1)
            {
                outputBuffer();
            }
            ++value;
        }
    }
```

cb-mega: I think to get a definite answer on this we'd have to measure across different string sizes. However, the real big fish to fry will be in MEGAsync itself where logging seems quite inefficient.

include/mega/logging.h
cb-mega: With the introduction of `ENABLE_LOG_PERFORMANCE` it made more sense to use "(NULL)" inline again so I got rid of `null`.

PR #1607, opened 2019-09-18
#13248. Do not track non-contacts keys

sergiohs84:
Additionally, it solves the fetching of user attributes, when it fails, for non-contact users.

**Risk area/s:**
- Retrieval of non-existing user attributes (firstname, avatar, lastname...) for non-contacts (excluded ex-contacts)


alber2510:
trigger compilation

PR #1610, opened 2019-09-19
Fix set user alias

jgandres:
Fix MegaStringMap construction

**Risk area/s:**
- Set user alias

### DIFF
--- a/src/megaapi_impl.cpp
+++ b/src/megaapi_impl.cpp
@@ -9950,9 +9950,9 @@ void MegaApiImpl::getUserAlias(MegaHandle uh, MegaRequestListener *listener)
 void MegaApiImpl::setUserAlias(MegaHandle uh, const char *alias, MegaRequestListener *listener)
 {
     MegaRequestPrivate *request = new MegaRequestPrivate(MegaRequest::TYPE_SET_ATTR_USER, listener);
-    unique_ptr<MegaStringMap> stringMap(new MegaStringMapPrivate());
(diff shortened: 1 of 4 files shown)


src/megaapi_impl.cpp
```diff
@@ -9950,9 +9950,9 @@ void MegaApiImpl::getUserAlias(MegaHandle uh, MegaRequestListener *listener)
 void MegaApiImpl::setUserAlias(MegaHandle uh, const char *alias, MegaRequestListener *listener)
 {
     MegaRequestPrivate *request = new MegaRequestPrivate(MegaRequest::TYPE_SET_ATTR_USER, listener);
-    MegaStringMap stringMap;
-    stringMap.set(Base64Str<MegaClient::USERHANDLE>(uh), alias ? alias : "");
-    request->setMegaStringMap(&stringMap);
+    unique_ptr<MegaStringMap> stringMap(new MegaStringMapPrivate());
```
sergiohs84: Good catch!

Actually, the old declaration of the `stringMap` variable should result in a compilation error. We should define an default ctor as protected, to avoid this kind of issues. Please, add it to every public object that provides a `createInstance()` method, since those are the risky ones (for the other types, it's always the SDK who creates the private object and provides it to the app via a pointer to the public object).

Then, the right declaration here would become:
```suggestion
    unique_ptr<MegaStringMap> stringMap(MegaStringMap::createInstance());
```

sergiohs84:
> megasdk_PR FAILED (1290) -1

Again the test for streaming cloud-raid files, it timed out. Unrelated to this PR

PR #1611, opened 2019-09-19
#13259. Webdav server compliance with RFC 7233 - 2.1

polmr:
limit the end of a requested range to the total size of a file for HTTP streaming

no longer returns 416 for request that petition for out of range bytes
(which is compliant with RFC 7233 - 2.1 Byte Ranges)


polmr:
Thanks @khand100 !

PR #1613, opened 2019-09-19
#13260. New event to notify user flags availability/update

sergiohs84:


src/megaapi.cpp
```diff
@@ -5931,6 +5931,11 @@ MegaHandle MegaEvent::getHandle() const
     return INVALID_HANDLE;
 }
 
+const char *MegaEvent::getEventString() const
+{
+    return MegaEvent::getEventString();
```
jnavarrom: Shouldn't this method return NULL?

sergiohs84: Agree. It is fixed here: https://github.com/meganz/sdk/pull/1618/commits/ccb8f0661583e731d79ce56af2363cbc3db459dd

PR #1616, opened 2019-09-20
#13273: prevent fopen from following symbolic links

polmr:
**Risk area/s:**
- Synchronization in POSIX systems
- Any operation with files/folders in POSIX systems

### DIFF
--- a/include/mega/posix/megafs.h
+++ b/include/mega/posix/megafs.h
@@ -80,7 +80,7 @@ public:
     int defaultfilepermissions;
     int defaultfolderpermissions;
 
-    FileAccess* newfileaccess();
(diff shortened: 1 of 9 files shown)


include/mega/posix/megafs.h
```diff
@@ -80,7 +80,7 @@ class MEGA_API PosixFileSystemAccess : public FileSystemAccess
     int defaultfilepermissions;
     int defaultfolderpermissions;
 
-    FileAccess* newfileaccess();
+    FileAccess* newfileaccess(bool followSymLinks = true);
```
sergiohs84: I will add some `override`s for the declaration of this virtual methods here, just to ensure we don't make a mistake changing the signature of the method at the parent class.

PR #1617, opened 2019-09-22
Smaller downloads for Soak test

mattw-mega:
To try to avoid test failures when download speeds are slow.

### DIFF
--- a/tests/integration/SdkTest_test.cpp
+++ b/tests/integration/SdkTest_test.cpp
@@ -3859,7 +3859,7 @@ TEST_F(SdkTest, SdkCloudraidStreamingSoakTest)
 
     transferFlags[0][MegaTransfer::TYPE_DOWNLOAD] = false;
     megaApi[0]->startDownload(nimported, filename2.c_str());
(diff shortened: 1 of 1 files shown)


tests/integration/SdkTest_test.cpp
```diff
@@ -3859,7 +3859,7 @@ TEST_F(SdkTest, SdkCloudraidStreamingSoakTest)
 
     transferFlags[0][MegaTransfer::TYPE_DOWNLOAD] = false;
     megaApi[0]->startDownload(nimported, filename2.c_str());
-    ASSERT_TRUE(waitForResponse(&transferFlags[0][MegaTransfer::TYPE_DOWNLOAD], 60)) << "Download transfer failed after " << maxTimeout << " seconds";
+    ASSERT_TRUE(waitForResponse(&transferFlags[0][MegaTransfer::TYPE_DOWNLOAD], 600)) << "Setup transfer failed after " << 60 << " seconds";
```
sergiohs84: Actually, 600s is the value used by default, so you can skip the param. What is good to fix is the log message: the test will fail after 600 seconds, not 60 seconds. It could be misleading during a log's investigation.
```suggestion
    ASSERT_TRUE(waitForResponse(&transferFlags[0][MegaTransfer::TYPE_DOWNLOAD])) << "Setup transfer failed after " << maxTimeout << " seconds";
```

PR #1620, opened 2019-09-24
#12268 Fix fs ID handling for FAT filesystems (core changes)

cb-mega:
On all FAT filesystem variants it may happen that filesystem IDs may change between mounts of that filesystem (e.g. SD cards for cameras and such). Because the sync algorithm heavily relies on those IDs to persist, this can lead to unnecessary transfers and quickly consume a user's transfer quota. 

To address this we came up with an algorithm that, on add of a sync, would look at the filesystem and assign the real fs IDs to the cached local nodes. This is based on file fingerprints and applying a heuristic that matches on paths. Please see `sync.cpp`.


src/sync.cpp
```diff
@@ -92,6 +487,9 @@ Sync::Sync(MegaClient* cclient, string* crootpath, const char* cdebris,
         fsfp = dirnotify->fsfingerprint();
     }
 
+    fsstableids = dirnotify->fsstableids();
+    LOG_info << "Filesystem IDs are stable: " << std::boolalpha << fsstableids;
```
sergiohs84: Nice, I didn't know about `std::boolalpha`. I'll start using it :smiley:

src/sync.cpp
```diff
@@ -36,6 +36,401 @@ const int Sync::FILE_UPDATE_DELAY_DS = 30;
...
+            LOG_err << "Invalid fingerprint: " << node.localname;
+            return;
+        }
+        ffp = node;
+    }
+    else if (node.type == FOLDERNODE)
+    {
+        combinedFingerprint(ffp, node.children);
+    }
+    else
```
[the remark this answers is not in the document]
cb-mega: Yeah, good point. Will rename.

src/sync.cpp
```diff
@@ -36,6 +36,401 @@ const int Sync::FILE_UPDATE_DELAY_DS = 30;
...
+            if (score > bestScore)
+            {
+                bestScore = score;
+                bestNodeIt = nodeIt;
+            }
+        }
+
+        bestNodeIt->second->setfsid(fsId);
+        fingerprints.erase(bestNodeIt);
+    }
```
polmr: I might foresee some possible issue here. 
If I haven't misunderstood the idea here: 
in case there are more than one file with the same fingerprint as me, I will chose the one closer to my location, take it to dance and remove it from the list of available fingerprints.
But, what if that was a perfect match for another file? I might be stealing it from the rightful node. Aren't I?

src/sync.cpp
```diff
@@ -36,6 +36,401 @@ const int Sync::FILE_UPDATE_DELAY_DS = 30;
 const int Sync::FILE_UPDATE_MAX_DELAY_SECS = 60;
 const dstime Sync::RECENT_VERSION_INTERVAL_SECS = 10800;
 
+namespace {
+
+// Collects all syncable filesystem paths in the given folder under `localpath`
+set<string> collectAllPathsInFolder(Sync& sync, MegaApp& app, FileSystemAccess& fsaccess, string localpath,
+                                    const string& localdebris, const string& localseparator, const bool followsymlinks)
+{
+    auto fa = std::unique_ptr<FileAccess>{fsaccess.newfileaccess()};
```
polmr: This and others newfileaccess calls should pass false to followSymLinks. Otherwise we would be collecting data for the wrong files. As a matter of fact, if fa->mIsSymLink evals true, we shouldn't add the path until symlinks are supported

src/sync.cpp
cb-mega: Thanks for sticking your head into this :) 
The thing is we don't know which node is the best match for fa's fingerprint. So when we have multiple nodes with the same fingerprint we pick the one that best matches fa's path. It's simply a heuristic that I thought applies in most circumstances but may sometimes be wrong (very rarely I think because fingerprints have to match). If the fingerprints on disk changed the fsid will simply stay invalid and reassigned later in the sync algo and the file considered new and re-transferred. 
I will update the comments in this functions (the debris one is stale).

src/sync.cpp
cb-mega: This should be addressed now. Fun fact: You can't actually create links on FAT :)
Still good that we've got this part covered I think. Who knows, maybe we'll use the functionality for something else in the future, though I hope we don't have to...

PR #1624, opened 2019-09-25
SMS verification iOS Binding

simonwng:
SMS verification iOS Binding project for https://github.com/meganz/iOS_dev/pull/360

### DIFF
--- a/bindings/ios/MEGASdk.mm
+++ b/bindings/ios/MEGASdk.mm
@@ -2171,6 +2171,11 @@ using namespace mega;
     return (SMSState)self.megaApi->smsAllowedState();
 }
 
+- (NSString *)smsVerifiedPhoneNumber {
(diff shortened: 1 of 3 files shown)


bindings/ios/MEGASdk.mm
```diff
@@ -2161,6 +2161,11 @@ - (SMSState)smsAllowedState {
     return (SMSState)self.megaApi->smsAllowedState();
 }
 
+- (NSString *)smsVerifiedPhoneNumber {
+    char *number = self.megaApi->smsVerifiedPhoneNumber();
```
jnavarrom: Memory leak here.
smsVerifiedPhoneNumber  -> You take the ownership of the returned value.
We should release the memory.

lancy-mega: It's handled now. Please have a look

PR #1636, opened 2019-10-01
#13043: iOS binding for startUploadForChat method.

lancy-mega:
#13043: iOS binding for `startUploadForChat` method.

### DIFF
--- a/bindings/ios/MEGASdk.mm
+++ b/bindings/ios/MEGASdk.mm
@@ -1418,6 +1418,18 @@ using namespace mega;
     self.megaApi->startUploadWithTopPriority((localPath != nil) ? [localPath UTF8String] : NULL, (parent != nil) ? [parent getCPtr] : NULL, (appData !=nil) ? [appData UTF8String] : NULL, isSourceTemporary);
 }
 
+- (void)startUploadForChatWithLocalPath:(NSString *)localPath
+                                 parent:(MEGANode *)parent
+                                appData:(NSString *)appData
+                      isSourceTemporary:(BOOL)isSourceTemporary
+                               delegate:(id<MEGATransferDelegate>)delegate {
+    self.megaApi->startUploadForChat((localPath != nil) ? [localPath UTF8String] : NULL,
+                                     (parent != nil) ? [parent getCPtr] : NULL,
+                                     (appData !=nil) ? [appData UTF8String] : NULL,
+                                     isSourceTemporary,
+                                     [self createDelegateMEGATransferListener:delegate singleListener:YES]);
+}
+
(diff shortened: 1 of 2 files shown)


bindings/ios/MEGASdk.mm
```diff
@@ -1418,6 +1418,18 @@ - (void)startUploadTopPriorityWithLocalPath:(NSString *)localPath parent:(MEGANo
     self.megaApi->startUploadWithTopPriority((localPath != nil) ? [localPath UTF8String] : NULL, (parent != nil) ? [parent getCPtr] : NULL, (appData !=nil) ? [appData UTF8String] : NULL, isSourceTemporary);
 }
 
+- (void)startUploadForChatWithLocalPath:(NSString *)localPath
+                                 parent:(MEGANode *)parent
+                                appData:(NSString *)appData
+                      isSourceTemporary:(BOOL)isSourceTemporary
+                               delegate:(id<MEGATransferDelegate>)delegate {
+    self.megaApi->startUploadForChat((localPath != nil) ? [localPath UTF8String] : NULL,
```
jnavarrom: Usually, we used to do this check with strings `(localPath != nil) ? [localPath UTF8String] : NULL` but we can migrate to `localPath.UTF8String`, it should be enough. It is not necessary to check if it is `nil`, technically `nil` and `NULL` are the same and both equal 0. `UTF8String` should be guaranteed to return a `const char *` type pointer. That means, even `localPath` is `nil`, it will return a `const char *` type empty, which is `NULL`.

What do you think @lancy-mega ?

Note: For the old places where we used `string != nil` we can leave it as it is, but for the new ones we will add we can use the new way to do it, that is: don't check if the string is `nil`, just use `string.UTF8String`

bindings/ios/MEGASdk.mm
```diff
@@ -1418,6 +1418,18 @@ - (void)startUploadTopPriorityWithLocalPath:(NSString *)localPath parent:(MEGANo
...
+                                     (parent != nil) ? [parent getCPtr] : NULL,
+                                     (appData !=nil) ? [appData UTF8String] : NULL,
```
jnavarrom: Same here.

bindings/ios/MEGASdk.mm
lancy-mega: @jnavarrom, Yes you are right. Sending a message to nil is legal and returns nil (NULL or 0). I think it is more of developer coding style. Since the other `startupload` method had the check, I too used the same style . I have removed it now as it adds simplicity to the code. Please have a look.

bindings/ios/MEGASdk.mm
lancy-mega: Changed.

PR #1637, opened 2019-10-01
Enable SDK clients to lock the sdkMutex for performance reasons

mattw-mega:
Sometimes clients need to make many calls in a row that each lock the sdkMutex.  If the SDK is busy, many interruptions may occur in that sequence.   This allows that to occur all in one quick go.

MEGAsync in particular can benefit form this because it does a sequence of locking in its paint functions when transfers are first processed there.

The object returned manages the lock on the mutex, ensuring it is locked max once (at a time), and is unlocked on object destruction.

### DIFF
--- a/include/megaapi.h
+++ b/include/megaapi.h
@@ -6627,6 +6627,43 @@ public:
 class MegaApiImpl;
 
 /**
+ * @brief Allows calling many synchronous operations on MegaApi without being blocked by SDK activity.
+ *
+ * Call MegaApi::getMegaApiLock() to get an instance of this class to use.
+ */
+class MegaApiLock
+{
+    MegaApiImpl* api;
+    bool locked = false;
+    MegaApiLock(MegaApiImpl*, bool lock);
+    friend class MegaApi;
+
+public:
+    /**
+     * @brief Lock the MegaApi if this instance does not currently have a lock on it yet.
+     * 
+     * There is no harm in calling this more than once, the MegaApi will only be locked 
+     * once, and the first unlock() call will release it.
+     */
+    void lockOnce();
+
+    /**
+     * @brief Release the lock on the MegaApi if one is still held by this instance
+     *
+     * The MegaApi will be unable to continue work until all MegaApiLock objects release
+     * their locks.  Only use multiple of these if you need nested locking.  The destructor
+     * of the object will release the lock, so it is sufficient to delete it when finished.
+     * However, when using it from a garbage collected language it may be prudent to call unlock() directly.
+     */
+    void unlock();
+    
+    /**
+     * @brief Destructor.  This will call unlock() if the MegaApi is still locked by this instance.
+     */
+    ~MegaApiLock();
+};
+
+/**
  * @brief Allows to control a MEGA account or a shared folder
  *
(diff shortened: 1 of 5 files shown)


include/megaapi.h
```diff
@@ -6626,6 +6626,52 @@ class MegaInputStream
...
+     */
+    void unlockOnce();
+
+    /**
+     * @brief Destructor.  This will call unlock() if the MegaApi is still locked by this instance.
+     */
+    ~MegaApiLock();
+
+private:
+    MegaApiImpl* api;
```
cb-mega: Can be a reference?

include/megaapi.h
```diff
@@ -6626,6 +6626,52 @@ class MegaInputStream
...
+    bool locked = false;
+    MegaApiLock(MegaApiImpl*, bool lock);
+    MegaApiLock(const MegaApiLock&) = delete;
+    void operator=(const MegaApiLock&) = delete;
```
cb-mega: Should return `MegaApiLock&`

src/megaapi.cpp
```diff
@@ -5706,6 +5711,38 @@ MegaInputStream::~MegaInputStream()
 
 }
 
+MegaApiLock::MegaApiLock(MegaApiImpl* ptr, bool lock) : api(ptr)
```
cb-mega: Can we pass-by-ref here and name it `api`, not `ptr`?

include/megaapi.h
mattw-mega: done :-)

src/megaapi.cpp
mattw-mega: done :-)

include/megaapi.h
mattw-mega: it only matches on the parameters, not the return type - but I've made the change anyway :-)   cheers

include/megaapi.h
```diff
@@ -6654,13 +6653,23 @@ class MegaApiLock
      * their locks.  Only use multiple of these if you need nested locking.  The destructor
      * of the object will release the lock, so it is sufficient to delete it when finished.
      * However, when using it from a garbage collected language it may be prudent to call unlock() directly.
+     *
+     * This function must be called from the same thread that called MegaApiLock::lockOnce().
```
polmr: The same should apply to the destruction of the object

src/win32/fs.cpp
```diff
@@ -1455,7 +1455,7 @@ WinDirNotify::~WinDirNotify()
 #endif
 }
 
-FileAccess* WinFileSystemAccess::newfileaccess(bool followSymLinks = true)
+FileAccess* WinFileSystemAccess::newfileaccess(bool followSymLinks)
```
polmr: why? it doesn't matter since it's unused, but I can't help to wonder why the change? ;-)

include/megaapi.h
mattw-mega: thanks, I'll add that comment

src/win32/fs.cpp
mattw-mega: The MSVC compiler won't accept that on the function definition, only on the declaration, even if it matches the declaration exactly.  It throws an error, otherwise I would have left it :-)  cheers

PR #1638, opened 2019-10-02
#13382. When operating on several or many transactions, commit to the db just…

mattw-mega:
… once for the batch.

This provides a big speedup when MEGASync is operating on many files at once, eg. uploading/downloading a large folder.


src/megaclient.cpp
```diff
@@ -12364,7 +12360,8 @@ bool MegaClient::syncdown(LocalNode* l, string* localpath, bool rubbish)
 
                     rit->second->syncget = new SyncFileGet(l->sync, rit->second, localpath);
                     nextreqtag();
-                    startxfer(GET, rit->second->syncget);
+                    DBTableTransactionCommitter committer(tctable); // TODO: use one committer for all files in the loop, without calling syncdown() recursively
```
polmr: Missing beginOnce? a `filecacheadd` that leads to `tctable->put` might be executed within starxfer.  In any case, I'm missing where the tctable->begin was called in the past in this case. I wonder if this requires a committer at all.

src/db.cpp
```diff
@@ -80,6 +80,23 @@ bool DbTable::next(uint32_t* type, string* data, SymmCipher* key)
...
 
+void DbTable::checkTransaction()
+{
+    if (mCheckAlwaysTransacted)
+    {
+        assert(mCurrentTransaction);  // if this fails, we should have started a DBTableTransactionCommitter higher in the call stack
+        if (mCurrentTransaction)
+        {
+            mCurrentTransaction->beginOnce();
+        }
```
polmr: The assert should be `mCurrentTransaction && mCurrentTransaction.started`, shouldn't it?
Similarly the `if`, should do `mCurrentTransaction && !mCurrentTransaction.started`, although that's already checked within `beginOnce`

mattw-mega: Usually that would be the case, but on the first operation since the committer was created, `started` would be false, and the `beginOnce()` will start the transaction at this point.   The programmer can choose to start it themselves by calling `beginOnce` manually of course, but sometimes (maybe usually?) it's better to defer that, and let the first database operation start the transaction (for those tables with checkAlwaysTransacted turned on) in case the transaction is not needed at all - there may be some conditionals on the database operation, or a loop that might have 0 iterations, and if we can avoid the begin/commit entirely in those cases, it's much better, for performance and to avoid the begin/commit logging also.     One other point, keeping that logic in the committer object allows those to still be used occasionally for a db table that doesn't have the checkAlwaysTransacted turned on, should that be useful.  cheers!

src/megaclient.cpp
mattw-mega: excellent point, and I'm working on extending the changes to cover those cases.  thanks!

src/megaclient.cpp
```diff
@@ -10673,7 +10670,7 @@ void MegaClient::enabletransferresumption(const char *loggedoutid)
     // postpone the resumption until the filesystem is updated
     if ((!sid.size() && !loggedinfolderlink()) || statecurrent)
     {
-        tctable->begin();
+        DBTableTransactionCommitter committer(tctable);
```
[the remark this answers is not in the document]
mattw-mega: Thanks for pointing out these cases, I'm extending the PR to cover them.  thanks!

src/megaclient.cpp
```diff
@@ -3970,7 +3964,7 @@ bool MegaClient::procsc()
 
                         if (tctable && cachedfiles.size())
                         {
-                            tctable->begin();
+                            DBTableTransactionCommitter committer(tctable);
```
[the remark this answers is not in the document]
mattw-mega: Yes, and the new changes as a result of these points will make sure that transactions are begun (and committed) in all these cases.  cheers!

src/transferslot.cpp
```diff
@@ -227,7 +227,7 @@ TransferSlot::~TransferSlot()
 
         if (cachetransfer)
         {
-            transfer->client->transfercacheadd(transfer);
+            transfer->client->transfercacheadd(transfer, nullptr);
```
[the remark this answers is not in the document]
mattw-mega: Yes, I would much rather have made that a reference, but since database operations are being called from a destructor, we can't have a committer as a parameter (or any parameters at all, of course).   Requiring one as a parameter very useful in general I think, in order to remind ourselves that there are database operations going on in the function call, which otherwise is often not obvious.  And it helps a lot with finding the code paths that need committer objects too.   Though, passing them as parameters is not strictly needed, since the database operations will be processed whether they are in a transaction or not - it's just a matter of knowing and reminding ourselves to batch them in transactions for efficiency where possible.   So anyway, making the parameter a pointer instead of a reference here means that we can pass the object when it's available, but pass null from destructors.   So we are reminded that we should be making transactions, but we can also choose to side-step actually passing the transaction object when it's inconvenient to do so.   I hope that makes sense!  cheers

PR #1641, opened 2019-10-03
#13380. Open files just once during fingerprinting.

mattw-mega:
Opening and closing them for each piece of the file read is much slower, on windows at least.

### DIFF
--- a/include/mega/filesystem.h
+++ b/include/mega/filesystem.h
@@ -102,7 +102,7 @@ struct MEGA_API FileAccess
     // open for reading, writing or reading and writing
     virtual bool fopen(string*, bool, bool) = 0;
 
-    // open by name only. returns false for folders setting type to FOLDERNODE
+    // Only prepares for opening.  Actually stats the file/folder, getting mtime, size, type.  Call openf() afterwards to actually open it if required.  For folders, retunrs false with type==FOLDERNODE.
     bool fopen(string*);
 
     // check if a local path is a folder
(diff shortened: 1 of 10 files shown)


src/filefingerprint.cpp
```diff
@@ -190,7 +190,7 @@ bool FileFingerprint::genfingerprint(FileAccess* fa, bool ignoremtime)
                 if (!fa->frawread(block, sizeof block,
                                   (size - sizeof block)
                                   * (i * blocks + j)
-                                  / (sizeof crc / sizeof *crc * blocks - 1)))
+                                  / (sizeof crc / sizeof *crc * blocks - 1), i || j, true))
                 {
                     size = -1;
                     return true;
```
polmr:
```suggestion
                    fa->closef();
                    return true;
```

polmr: actually this should only be done in case the failure does not come from fopen. We could add a isOpened()  and do `if (fa->isOpened()) fa->closef();`

mattw-mega: I saw in your PR for this branch, and that the file is always already opened before calling this function fingerprinting!   Great spotting, I should have seen that too.  Thanks very much for the fix :-)

include/mega/filesystem.h
```diff
@@ -117,19 +117,15 @@ struct MEGA_API FileAccess
     // absolute position read to byte buffer
     bool frawread(byte *, unsigned, m_off_t, bool skipopen = false, bool keepopen = false);
 
-    // non-locking ops: open/close temporary hFile
+    // After a successful fopen(), call fopen() to really open the file (by localname) (this is a lazy-type approach in case we don't actually need to open the file after finding out type/size/mtime).  If the size or mtime changed, it will fail.
```
polmr: `call fopen()` -> `call openf()`?

PR #1643, opened 2019-10-04
#13615 Fix fs ID handling for FAT filesystems (improved algo + unit testing)

cb-mega:
This PR contains an improved algorithm for assigning fs IDs from the filesystem to cached localnodes. In particular, the best-path match heuristic has been extended to be symmetric between files and localnodes. In order to speed up computation, we're now using a light-weight fingerprint.

This PR changes 8 production sources files. The remaining changes are build config, megacli and unit tests.

It was attempted to unit test this functionality. And for this, some changes had to be made to the production code. Most notably:
• Add default initializers to LocalNode and Sync so that, when default constructed, objects would be in a well defined state. This should not affect production because before member fields would contain random values.
• Some checks for a null `client` had to be added because the `client` is currently null in unit tests. This was deemed to be the least intrusive way. Alternatively one could add a client interface but that would mean even more changes.
• Some functions had their signature changed to receive a parameter so the `client` doesn't have to be accessed.
• The `FileFingerprint`'s `crc` was changed from a C-style array to a `std::array` to make it easier to 1) define copy constructor/assignment and 2) make unit testing easier. Note that `FileFingerprint` now has almost full unit test coverage (97%).


contrib/QtCreator/MEGAtests/MEGAtest_integration/MEGAtest_integration.pro
```diff
@@ -16,6 +16,7 @@ CONFIG += USE_MEDIAINFO
 CONFIG += USE_LIBRAW
 CONFIG += USE_FFMPEG
 CONFIG -= qt
+CONFIG += object_parallel_to_source
```
cb-mega: Ensures we don't get clashes between object files from prod and test code.

contrib/cmake/CMakeLists.txt
```diff
@@ -607,19 +607,66 @@ if(WIN32)
...
+        find_program( CODECOV_LCOV lcov )
+        find_program( CODECOV_GENHTML genhtml )
+        add_definitions( -fprofile-arcs -ftest-coverage )
+        link_libraries( gcov )
+        set( CMAKE_EXE_LINKER_FLAGS ${CMAKE_EXE_LINKER_FLAGS} --coverage )
+        add_custom_target( coverage_init ALL ${CODECOV_LCOV} --base-directory .  --directory ${CMAKE_BINARY_DIR} --output-file ${CODECOV_OUTPUTFILE} --capture --initial )
+        add_custom_target( coverage ${CODECOV_LCOV} --base-directory .  --directory ${CMAKE_BINARY_DIR} --output-file ${CODECOV_OUTPUTFILE} --capture COMMAND genhtml -o ${CODECOV_HTMLOUTPUTDIR} ${CODECOV_OUTPUTFILE} )
+endif ( CMAKE_COMPILER_IS_GNUCXX )
+
+endif (ENABLE_CODECOVERAGE )
```
cb-mega: This block is to enable code coverage through unit tests.

examples/megacli.cpp
```diff
@@ -511,7 +511,7 @@ AppFileGet::AppFileGet(Node* n, handle ch, byte* cfilekey, m_off_t csize, m_time
 
         if (!cfingerprint->size() || !unserializefingerprint(cfingerprint))
         {
-            memcpy(crc, filekey, sizeof crc);
+            memcpy(crc.data(), filekey, sizeof crc);
```
cb-mega: This change has been made in a few places because `crc` is now a `std::array` (see PR description).

include/mega/node.h
```diff
@@ -60,7 +60,7 @@ struct MEGA_API NewNode : public NodeCore
 
     handle ovhandle;
     handle uploadhandle;
-    byte uploadtoken[UPLOADTOKENLEN];
+    byte uploadtoken[UPLOADTOKENLEN]{};
```
cb-mega: This array was previously not initialized in the `Node`s constructor.

include/mega/sync.h
```diff
@@ -115,47 +112,48 @@ class MEGA_API Sync
...
     // values related to possible files being updated
-    m_off_t updatedfilesize;
-    m_time_t updatedfilets;
-    m_time_t updatedfileinitialts;
-    
+    m_off_t updatedfilesize = ~0;
+    m_time_t updatedfilets = 0;
+    m_time_t updatedfileinitialts = 0;
+
+    Sync() = default;
```
cb-mega: This constructor is required for unit tests

src/filefingerprint.cpp
```diff
@@ -165,10 +165,10 @@ bool FileFingerprint::genfingerprint(FileAccess* fa, bool ignoremtime)
             return true;
         }
 
-        for (unsigned i = 0; i < sizeof crc / sizeof *crc; i++)
+        for (unsigned i = 0; i < crc.size(); i++)
```
cb-mega: Because `crc` is now a `std::array` we can get its size through the `.size()` member function

src/node.cpp
```diff
@@ -1498,7 +1493,7 @@ void LocalNode::setnotseen(int newnotseen)
 }
 
 // set fsid - assume that an existing assignment of the same fsid is no longer current and revoke
-void LocalNode::setfsid(handle newfsid)
+void LocalNode::setfsid(handle newfsid, handlelocalnode_map& fsidnodes)
```
cb-mega: This is also necessary for unit testing to avoid having to access the `client` directly.

src/node.cpp
```diff
@@ -1540,6 +1535,11 @@ LocalNode::~LocalNode()
         return;
     }
 
+    if (!sync->client)
```
cb-mega: Here and in other places we have to check for `client` being null because it is null in unit tests.

src/sync.cpp
```diff
@@ -508,7 +534,7 @@ Sync::Sync(MegaClient* cclient, string* crootpath, const char* cdebris,
     }
 
     fsstableids = dirnotify->fsstableids();
-    LOG_info << "Filesystem IDs are stable: " << std::boolalpha << fsstableids;
+    LOG_info << "Filesystem IDs are stable: " << fsstableids;
```
cb-mega: Removed boolalpha because it's actually sticky

src/posix/fs.cpp
```diff
@@ -1787,7 +1787,7 @@ bool PosixDirNotify::fsstableids() const
         return true;
     }
 
-    LOG_info << "Filesystem type: 0x" << std::hex << statfsbuf.f_type;
+    LOG_info << "Filesystem type: " << statfsbuf.f_type;
```
cb-mega: Removed std::hex because it's actually sticky

src/sync.cpp
```diff
@@ -149,10 +195,12 @@ void combinedFingerprint(FileFingerprint& ffp, FileSystemAccess& fsaccess, const
         if (fa->type == FILENODE)
         {
             FileFingerprint faFfp;
-            faFfp.genfingerprint(fa.get());
+            const auto index = fsaccess.lastpartlocal(const_cast<string*>(&path));
```
cb-mega: Need to get the leaf name of the path here. This needs to be equivalent to `localname` of `LocalNode` so that the light-weight fingerprints become comparable.
Please let me know if there's a better more doing this.

include/mega/node.h
```diff
@@ -249,42 +249,42 @@ struct MEGA_API Node : public NodeCore, FileFingerprint
...
 
     // number of iterations since last seen
-    int notseen;
+    int notseen = 0;
 
     // global sync reference
-    handle syncid;
+    handle syncid = mega::UNDEF;
 
     struct
```
cb-mega: Sadly, bit fields cannot be initialized at declaration (coming with c++20 though).

include/mega/file.h
```diff
@@ -87,14 +87,14 @@ struct MEGA_API File: public FileFingerprint
     char *chatauth;
 
     // if !hprivate, filekey and size must be valid
-    byte filekey[FILENODEKEYLENGTH];
+    byte filekey[FILENODEKEYLENGTH]{};
```
cb-mega: This array was previously not initialized in the `File`'s constructor.

src/sync.cpp
```diff
@@ -331,6 +323,85 @@ void assignFilesystemIdsImpl(bool& success, Sync& sync, MegaApp& app, handleloca
...
+            auto l = nodeIt->second;
+            if (l != &l->sync->localroot) // never assign fs ID to the root localnode
+            {
+                nodePath.clear();
+                l->getlocalpath(&nodePath, false, &localseparator);
+                for (auto fileIt = fileRange.first; fileIt != fileRange.second; ++fileIt)
+                {
+                    const auto& filePath = fileIt->second.path;
+                    const auto score = computeReversePathMatchScore(accumulated, nodePath, filePath, localseparator);
+                    if (score > 0) // leaf name must match
```
polmr: why? don't we want a match of a renamed file? what am I missing?

src/sync.cpp
```diff
@@ -620,7 +685,7 @@ void Sync::addstatecachechildren(uint32_t parent_dbid, idlocalnode_map* tmap, st
 
         l->parent_dbid = parent_dbid;
         l->size = size;
-        l->setfsid(fsid);
+        l->setfsid(fsid, l->sync->client->fsidnode);
```
polmr: lost check (l->sync) != null included in former `setfsid`
Might not be required though

src/sync.cpp
```diff
@@ -1203,7 +1268,7 @@ LocalNode* Sync::checkpath(LocalNode* l, string* localpath, string* localname, d
                         {
                             if (fa->fsidvalid && l->fsid != fa->fsid)
                             {
-                                l->setfsid(fa->fsid);
+                                l->setfsid(fa->fsid, l->sync->client->fsidnode);
```
polmr: lost check (l->sync) != null included in former `setfsid`

src/sync.cpp
```diff
@@ -1242,7 +1307,7 @@ LocalNode* Sync::checkpath(LocalNode* l, string* localpath, string* localname, d
                         // content scan anyway)
                         if (fa->fsidvalid && fa->fsid != l->fsid)
                         {
-                            l->setfsid(fa->fsid);
+                            l->setfsid(fa->fsid, l->sync->client->fsidnode);
```
polmr: lost check (l->sync) != null included in former `setfsid`

src/sync.cpp
```diff
@@ -1429,7 +1494,7 @@ LocalNode* Sync::checkpath(LocalNode* l, string* localpath, string* localname, d
 
                     if (fa->fsidvalid)
                     {
-                        l->setfsid(fa->fsid);
+                        l->setfsid(fa->fsid, l->sync->client->fsidnode);
```
polmr: lost check (l->sync) != null included in former `setfsid`

src/sync.cpp
```diff
@@ -1470,7 +1535,7 @@ LocalNode* Sync::checkpath(LocalNode* l, string* localpath, string* localname, d
                 {
                     if (fa->fsidvalid && l->fsid != fa->fsid)
                     {
-                        l->setfsid(fa->fsid);
+                        l->setfsid(fa->fsid, l->sync->client->fsidnode);
```
polmr: lost check (l->sync) != null included in former `setfsid`

include/mega/filefingerprint.h
```diff
@@ -59,4 +65,26 @@ struct MEGA_API FileFingerprintCmp
...
+{
+    m_off_t size = -1;
+    m_time_t mtime = 0;
+
+    LightFileFingerprint() = default;
+
+    MEGA_DEFAULT_COPY_MOVE(LightFileFingerprint)
+
+    // Generates a new fingerprint not involving I/O
+    bool genfingerprint(m_off_t filesize, m_time_t filemtime);
```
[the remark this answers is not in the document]
cb-mega: I am not sure if the client needs to know what this function really does under the hood? Maybe later we'll change it to something else. Plus I think it's nice to have a consistent method name across the fingerprint classes. Speaking of which, maybe we could change all `genfingerprint` methods to be named just `from` or `make`?

src/sync.cpp
```diff
@@ -96,49 +142,41 @@ set<string> collectAllPathsInFolder(Sync& sync, MegaApp& app, FileSystemAccess&
...
+    bool success = false;
     for (const auto& path : paths)
     {
         auto fa = fsaccess.newfileaccess(false);
         if (!fa->fopen(const_cast<string*>(&path), true, false))
         {
             LOG_err << "Unable to open path: " << path;
-            ffp.isvalid = false;
+            success = false;
             break;
```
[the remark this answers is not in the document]
polmr: Sure. Agreed

src/sync.cpp
polmr: Thanks for the the explanation Christian. As a matter of fact I was thinking that would be the reason later on, after posting the comment. We might want to add a word at the definition of `assignFilesystemIds`, or at `assignFilesystemIdsImpl` or `computeReversePathMatchScore` explaining that _renames_ are considered misses due to the heuristics computation  (and that `if` prevents them from being assigned). Anyway: no showstopper

include/mega/filefingerprint.h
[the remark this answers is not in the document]
polmr: I like `from`, but since those are called with not too neat named variables (ll, f, fp, ...) I guess `genfingerprint` does give more information. Let's keep that one.

PR #1652, opened 2019-10-09
#13436: Converted newfileaccess() to return a unique_ptr to get compiler supp…

mattw-mega:
…ort for making sure we always delete the object.

There were at least a couple of places that manual deletes were missed in the break; or continue; clauses of loops/switches.

### DIFF
--- a/include/megaapi_impl.h
+++ b/include/megaapi_impl.h
@@ -2924,6 +2924,10 @@ protected:
         bool sync_syncable(Sync *, const char*, string *, Node *) override;
         bool sync_syncable(Sync *, const char*, string *) override;
         void syncupdate_local_lockretry(bool) override;
+
+        // for the exclusive use of sync_syncable
+        unique_ptr<FileAccess> mSyncable_fa;
+        std::mutex mSyncable_fa_mutex;
 #endif
 
 protected:
--- a/include/megaapi_impl.h
+++ b/include/megaapi_impl.h
@@ -3260,7 +3260,7 @@ public:
     std::string host;
     std::string destination;
     bool overwrite;
-    FileAccess *tmpFileAccess;
+    std::unique_ptr<FileAccess> tmpFileAccess;
     std::string tmpFileName;
     std::string newname; //newname for moved node
     MegaHandle nodeToMove; //node to be moved after delete
@@ -3532,7 +3532,7 @@ public:
     m_off_t rangeWritten;
 
     std::string tmpFileName;
-    FileAccess* tmpFileAccess;
+    std::unique_ptr<FileAccess> tmpFileAccess;
     size_t tmpFileSize;
 
     bool controlRespondedElsewhere;
--- a/src/megaapi_impl.cpp
+++ b/src/megaapi_impl.cpp
@@ -27625,10 +27625,8 @@ MegaHTTPContext::MegaHTTPContext()
 MegaHTTPContext::~MegaHTTPContext()
 {
     delete node;
-    if (tmpFileAccess)
+    if (tmpFileName.size())
     {
-        delete tmpFileAccess;
-
         string localPath;
         server->fsAccess->path2local(&tmpFileName, &localPath);
         server->fsAccess->unlinklocal(&localPath);
@@ -30142,7 +30140,6 @@ MegaFTPDataContext::MegaFTPDataContext()
 MegaFTPDataContext::~MegaFTPDataContext()
 {
     delete transfer;
-    delete tmpFileAccess;
     delete node;
 }
 
--- a/src/megaapi_impl.cpp
+++ b/src/megaapi_impl.cpp
@@ -8182,7 +8182,7 @@ int MegaApiImpl::syncPathState(string* path)
                 fsAccess->local2name(&name);
                 if (is_syncable(sync, name.c_str(), path))
                 {
-                    FileAccess *fa = fsAccess->newfileaccess();
+                    auto fa = fsAccess->newfileaccess();
                     if (fa->fopen(path, false, false) && (fa->type == FOLDERNODE || is_syncable(fa->size)))
                     {
                         state = MegaApi::STATE_PENDING;
@@ -8191,7 +8191,6 @@ int MegaApiImpl::syncPathState(string* path)
                     {
                         state = MegaApi::STATE_IGNORED;
                     }
-                    delete fa;
                 }
                 else
                 {
@@ -11190,14 +11189,13 @@ char *MegaApiImpl::getFingerprint(const char *filePath)
     string localpath;
     fsAccess->path2local(&path, &localpath);
 
-    FileAccess *fa = fsAccess->newfileaccess();
+    auto fa = fsAccess->newfileaccess();
     if(!fa->fopen(&localpath, true, false))
         return NULL;
 
     FileFingerprint fp;
-    fp.genfingerprint(fa);
+    fp.genfingerprint(fa.get());
     m_off_t size = fa->size;
-    delete fa;
     if(fp.size < 0)
         return NULL;
 
@@ -11388,13 +11386,12 @@ char *MegaApiImpl::getCRC(const char *filePath)
     string localpath;
     fsAccess->path2local(&path, &localpath);
 
-    FileAccess *fa = fsAccess->newfileaccess();
+    auto fa = fsAccess->newfileaccess();
     if(!fa->fopen(&localpath, true, false))
         return NULL;
 
     FileFingerprint fp;
-    fp.genfingerprint(fa);
-    delete fa;
+    fp.genfingerprint(fa.get());
     if(fp.size < 0)
         return NULL;
 
@@ -12733,12 +12730,18 @@ bool MegaApiImpl::sync_syncable(Sync *sync, const char *name, string *localpath,
 
 bool MegaApiImpl::sync_syncable(Sync *sync, const char *name, string *localpath)
 {
-    static FileAccess* f = fsAccess->newfileaccess();
-    if (!sync || !sync->appData
-            || ((syncLowerSizeLimit || syncUpperSizeLimit)
-                && f->fopen(localpath) && !is_syncable(f->size)))
     {
-        return false;
+        std::lock_guard<std::mutex> g(mSyncable_fa_mutex);
+        if (!mSyncable_fa)
+        {
+            mSyncable_fa = fsAccess->newfileaccess();
+        }
+        if (!sync || !sync->appData
+            || ((syncLowerSizeLimit || syncUpperSizeLimit)
+                && mSyncable_fa->fopen(localpath) && !is_syncable(mSyncable_fa->size)))
+        {
+            return false;
+        }
     }
 
     sdkMutex.unlock();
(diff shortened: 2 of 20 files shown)


src/sync.cpp
```diff
@@ -1226,8 +1219,6 @@ LocalNode* Sync::checkpath(LocalNode* l, string* localpath, string* localname, d
 
                             statecacheadd(l);
 
-                            delete fa;
```
cb-mega: I think we should replace this with `fa.reset()` to ensure fa is deleted before calling `notify`

src/win32/fs.cpp
```diff
@@ -1476,9 +1476,9 @@ WinDirNotify::~WinDirNotify()
 #endif
 }
 
-FileAccess* WinFileSystemAccess::newfileaccess(bool followSymLinks)
+std::unique_ptr<FileAccess> WinFileSystemAccess::newfileaccess(bool followSymLinks)
 {
-    return new WinFileAccess(waiter);
+    return std::unique_ptr<FileAccess>(new WinFileAccess(waiter));
```
cb-mega: Use {} to call constructors

include/megaapi_impl.h
```diff
@@ -2924,6 +2924,10 @@ class MegaApiImpl : public MegaApp
         bool sync_syncable(Sync *, const char*, string *, Node *) override;
         bool sync_syncable(Sync *, const char*, string *) override;
         void syncupdate_local_lockretry(bool) override;
+
+        // for the exclusive use of sync_syncable
+        unique_ptr<FileAccess> mSyncable_fa;
+        std::mutex mSyncable_fa_mutex;
```
cb-mega: Normally we don't use snake case according to our guidelines..

src/megaapi_impl.cpp
```diff
@@ -11190,14 +11189,13 @@ char *MegaApiImpl::getFingerprint(const char *filePath)
...
-    FileAccess *fa = fsAccess->newfileaccess();
+    auto fa = fsAccess->newfileaccess();
     if(!fa->fopen(&localpath, true, false))
         return NULL;
 
     FileFingerprint fp;
-    fp.genfingerprint(fa);
+    fp.genfingerprint(fa.get());
     m_off_t size = fa->size;
-    delete fa;
```
cb-mega: replace this with fa.reset()?

src/megaapi_impl.cpp
```diff
@@ -11388,13 +11386,12 @@ char *MegaApiImpl::getCRC(const char *filePath)
...
     fsAccess->path2local(&path, &localpath);
 
-    FileAccess *fa = fsAccess->newfileaccess();
+    auto fa = fsAccess->newfileaccess();
     if(!fa->fopen(&localpath, true, false))
         return NULL;
 
     FileFingerprint fp;
-    fp.genfingerprint(fa);
-    delete fa;
```
cb-mega: replace this with fa.reset()?

src/megaapi_impl.cpp
```diff
@@ -24446,12 +24424,11 @@ void MegaFolderDownloadController::downloadFolderNode(MegaNode *node, string *pa
 
     string localpath;
     client->fsaccess->path2local(path, &localpath);
-    FileAccess *da = client->fsaccess->newfileaccess();
+    auto da = client->fsaccess->newfileaccess();
     if (!da->fopen(&localpath, true, false))
     {
         if (!client->fsaccess->mkdirlocal(&localpath))
         {
-            delete da;
```
cb-mega: Replace with da.reset()?

src/transfer.cpp
```diff
@@ -544,51 +542,53 @@ void Transfer::complete()
...
+                    LOG_err << "Fingerprint mismatch";
+
+                    // enforce the verification of the fingerprint for sync transfers only
+                    if (syncxfer && (!badfp.isvalid || !(badfp == fingerprint)))
                     {
-                        fixfingerprint = true;
+                        badfp = fingerprint;
+                        chunkmacs.clear();
+                        client->fsaccess->unlinklocal(&localfilename);
+                        fa.reset();
```
cb-mega: Let's move this back to before `chunkmacs.clear()`?

mattw-mega: good catch! done

src/megaapi_impl.cpp
mattw-mega: in this case the fopen failed so it won't have an open handle - I've replaced it anyway since checkcomplete() will be called, just in case.

src/megaapi_impl.cpp
mattw-mega: nothing significant occurs before the end of the function here, so I think it's ok

src/megaapi_impl.cpp
mattw-mega: nothing significant occurs before the end of the function here, so I think it's ok

include/megaapi_impl.h
mattw-mega: 'fa' is a well established name for FileAccess objects though, and would look odd with a capital F... if someone else wants to adjust it, I won't object :-)

src/win32/fs.cpp
mattw-mega: I prefer the original syntax, unless it's something that needs curly braces.  It's more backward compatible too.

src/sync.cpp
mattw-mega: replaced - I see notify() can open a path so best to make it work as it did.  thanks!

src/megaapi_impl.cpp
cb-mega: agreed

src/megaapi_impl.cpp
cb-mega: agreed

src/win32/fs.cpp
cb-mega: fair enough

src/megaapi_impl.cpp
```diff
@@ -13275,28 +13278,18 @@ void MegaApiImpl::fa_complete(handle, fatype, const char* data, uint32_t len)
...
-            delete f;
-            fireOnRequestFinish(request, MegaError(API_EWRITE));
-            continue;
-        }
+        bool success = f->fopen(&localPath, false, true)
+                    && f->fwrite((const byte*)data, len, 0);
 
-        delete f;
-        fireOnRequestFinish(request, MegaError(API_OK));
+        f.reset();
```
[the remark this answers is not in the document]
mattw-mega: Yes, I thought it's better to close the file handle before calling `fireOnRequestFinish`, since the 'fire' functions call the application back, and the client may want to open the file with its own handle etc.

src/megaapi_impl.cpp
```diff
@@ -14474,27 +14461,23 @@ void MegaApiImpl::getua_result(byte* data, unsigned len, attr_t type)
...
-                {
-                    delete f;
-                    fireOnRequestFinish(request, MegaError(API_EWRITE));
-                    return;
-                }
+                
+                bool success = f->fopen(&localPath, false, true)
+                            && f->fwrite((const byte*)data, len, 0);
+
+                f.reset();
```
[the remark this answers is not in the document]
mattw-mega: same here, I think it's better to close the file handle before calling the app back, in case it wants to open the file itself.

src/megaapi_impl.cpp
```diff
@@ -17437,9 +17419,9 @@ void MegaApiImpl::sendPendingTransfers()
                 FileFingerprint fp;
                 if (type == FILENODE)
                 {
-                    fp.genfingerprint(fa);
+                    fp.genfingerprint(fa.get());
                 }
-                delete fa;
+                fa.reset();
```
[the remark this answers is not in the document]
mattw-mega: yes, and we do have some 'fire' functions being called shortly after, and with the reset() called we are doing exactly what we used to do.  We could make a { } code block to limit `fa`'s scope instead, I thought this was a smaller change though.

src/megaapi_impl.cpp
```diff
@@ -17743,11 +17725,10 @@ void MegaApiImpl::sendPendingTransfers()
                             transfer->setMeanSpeed(0);
                             transfer->setState(MegaTransfer::STATE_COMPLETED);
                             fireOnTransferFinish(transfer, MegaError(API_OK));
-                            delete fa;
                             break;
                         }
                     }
-                    delete fa;
+                    fa.reset();
```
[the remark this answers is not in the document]
mattw-mega: yes, this one makes sure we close the file handle exactly where we did before, which is very low risk.  I think we need to close it before the call to `startxfer` which makes its own FileAccess and opens the file again, and doing so in the same place as before should avoid any problems.

src/megaapi_impl.cpp
```diff
@@ -18942,21 +18923,19 @@ void MegaApiImpl::sendPendingRequests()
...
                     }
 
                     if (!f->fread(&attrvalue, unsigned(f->size), 0, 0))
                     {
-                        delete f;
                         e = API_EREAD;
                         break;
                     }
-                    delete f;
+                    f.reset();
```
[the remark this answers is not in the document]
mattw-mega: this one closes the handle before we call putua(), which might call the application back, so it may be necessary - safer to do exactly what it did before in this case, I think.

src/sync.cpp
```diff
@@ -1109,11 +1105,9 @@ LocalNode* Sync::checkpath(LocalNode* l, string* localpath, string* localname, d
             {
                 LOG_verbose << "New file. FaType: " << fa->type << "  FaSize: " << fa->size << "  FaMtime: " << fa->mtime;
             }
-            delete fa;
             return NULL;
         }
 
-        delete fa;
         fa = client->fsaccess->newfileaccess(false);
```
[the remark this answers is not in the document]
mattw-mega: yes, the effect is exactly the same between those two syntax - I think the assignment looks a bit cleaner - the assign (which will resolve to a move-assign since copy is disallowed for unique_ptr) will first delete any ownership it currently has, and then take ownership of the FileAccess* passed in from the assignment.  Happy to change it though if you prefer.

src/transfer.cpp
```diff
@@ -544,51 +542,53 @@ void Transfer::complete()
             }
         }
 
-        if (!fixedfingerprint && success && fa->fopen(&localfilename, true, false))
+        if (auto fa = client->fsaccess->newfileaccess())
```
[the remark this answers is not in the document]
mattw-mega: Yes, I used it to limit the scope of `fa`  - though in retrospect, maybe it's less of a change to do it the other way - I'll change this one.

src/transferslot.cpp
```diff
@@ -175,12 +174,10 @@ TransferSlot::~TransferSlot()
             }
 
             // Open the file in synchonous mode
-            delete fa;
             fa = transfer->client->fsaccess->newfileaccess();
```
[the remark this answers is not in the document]
mattw-mega: The reset is done implicitly in the move-assignment on the next line - the unique_ptr will make sure anything it was previously owning is deleted before completing the assignment.

src/sync.cpp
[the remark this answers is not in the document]
polmr: nah, it's ok. I didn't notice that was the behaviour for `operator=`. Apologies for the noise.

PR #1655, opened 2019-10-10
#13437 Faster WinWaiter::wait().

mattw-mega:
This one gets called very frequently when the system is busy.  We can avoid deallocating and reallocating the handle and flag vectors each time, and the profiler shows that reduces CPU and time spent.

### DIFF
--- a/src/win32/waiter.cpp
+++ b/src/win32/waiter.cpp
@@ -102,7 +102,7 @@ int WinWaiter::wait()
     }
 
     addhandle(externalEvent, NEEDEXEC);
-    DWORD dwWaitResult = WaitForMultipleObjectsEx((DWORD)handles.size(), &handles.front(), FALSE, maxds * 100, TRUE);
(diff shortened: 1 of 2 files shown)


src/win32/waiter.cpp
```diff
@@ -125,8 +125,7 @@ int WinWaiter::wait()
         r |= flags[dwWaitResult - WAIT_OBJECT_0];
     }
 
-    handles.clear();
-    flags.clear();
+    index = 0;
```
sergiohs84: I don't know if there could be a huge amount of handles in `handles` temporary, but in that case maybe we want to shrink the vector from time to time. Imagine that only temporary we wait for 20.000 objects, but the average is in the hundreds. We would keep keep memory for the maximum amount of handles ever tracked, which is not that much, anyway (20k*sizeof(HANDLE)=156KB is probably acceptable, anyway)
Apart from that memory comsumption concern, the PR looks totally fine :)

mattw-mega: Yes, I thought about that too, but thought keeping the PR simple and just concentrating on performance for now is the way to go, memory for this one won't be too bad and we can adjust it later if it turns out to be an issue.  thanks!

PR #1658, opened 2019-10-11
#13463. Transfer and TransferSlot backoff timer efficency.

mattw-mega:
Only check the backoff timers that we need to when preparing a wait.   We used to iterate the whole list of transferslots and transfers checking each one, which cost a lot of CPU when the system was busy.
This change maintains a set of just those timers that are in play, and only checks that set (for each of the cases:  slots, put transfers, get transfers).
Also address a bug with the `retrying` flag not getting set true when a slot needs to back off, and a bug where that backoff was being set to a very large value when it shouldn't (which didn't cause any problem due to the first bug)


src/backofftimer.cpp
```diff
@@ -136,4 +136,85 @@ TimerWithBackoff::TimerWithBackoff(PrnGen &rng, int tag)
     this->tag = tag;
 }
 
+
+void BackoffTimerGroupTracker::update(dstime* waituntil, bool transfers)
+{
+    vector<BackoffTimerTracked*> v;
```
sergiohs84: Perhaps calling it `expiredTimers` or `armedBackoffs` could improve readability :)

src/megaclient.cpp
```diff
@@ -3582,23 +3576,11 @@ void MegaClient::freeq(direction_t d)
 }
 
 // determine next scheduled transfer retry
-// FIXME: make this an ordered set and only check the first element instead of
-// scanning the full map!
 void MegaClient::nexttransferretry(direction_t d, dstime* dsmin)
 {
-    for (transfer_map::iterator it = transfers[d].begin(); it != transfers[d].end(); it++)
+    if (!xferpaused[d])   // avoid setting the timer's next=1 if it won't be processed
```
sergiohs84: If some transfers are paused and we avoid to update its backoff timer, do we arm the timer when they are un-paused?

src/megaclient.cpp
```diff
@@ -2841,13 +2841,7 @@ int MegaClient::preparewait()
...
 
         // retry transferslots
-        for (transferslot_list::iterator it = tslots.begin(); it != tslots.end(); it++)
-        {
-            if (!(*it)->retrybt.armed())
-            {
-                (*it)->retrybt.update(&nds);
-            }
-        }
+        tslotsbackoff.update(&nds, false);
```
sergiohs84: Don't we do the opposite than before in here?
It used to update the timer when the backoff was not armed, but the `BackoffTimerGroupTracker::update()`, for slots, does update for those which are armed.

src/backofftimer.cpp
```diff
@@ -136,4 +136,85 @@ TimerWithBackoff::TimerWithBackoff(PrnGen &rng, int tag)
...
+        // used to be:
+        //for (transferslot_list::iterator it = tslots.begin(); it != tslots.end(); it++)
+        //{
+        //    if (!(*it)->retrybt.armed())
+        //    {
+        //        (*it)->retrybt.update(&nds);
+        //    }
+        //}
+
+        for (auto t : timeouts)
```
sergiohs84:
```suggestion
        // put the ones to work on in a vector, as working on them changes their position in the map
        for (auto t : timeouts)
```

include/mega/megaclient.h
```diff
@@ -1073,6 +1074,9 @@ class MEGA_API MegaClient
     // transfer tslots
     transferslot_list tslots;
 
+    // keep track of next transfer slot timeout
+    BackoffTimerGroupTracker tslotsbackoff;
```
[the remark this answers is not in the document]
mattw-mega: I was going for the strong association with the 'tslots' member - but I don't mind the update, done :-)  cheers

src/backofftimer.cpp
mattw-mega: I think we should keep calling this one `udpate`, since it's doing the same thing as BackoffTimer::update, but for the whole group - I've added comments to explain that aspect.  thanks

src/megaclient.cpp
mattw-mega: Those two actually operate independently - xferpaused just corresponds to whether the user has clicked 'pause' for puts/gets.  If the timer was in a backoff state when those are unpaused, then we'll start evaluating it again, and process it if the specifed backoff time has been reached.   This check was needed specifically because the timers have a built in check that we are using them correctly, if the backoff is expired when we call preparewait(), it sets its internal flag next=1, expecting that the next exec() will do something with that backed-off object.   If it doesn't, then we get an assertion on the next preparewait().   So with the pause turned on, we would not make that adjustment in exec(), and so we need to avoid considering those backofftimers in preparewait().  cheers

src/megaclient.cpp
mattw-mega: I think that check for !armed might be an attempt at efficiency, or a way to avoid that next=1 but not processed due to the `xferpaused` flag issue discussed above.   The thing is, the profiler showed this loop took a lot of CPU when we push the number of active transfers up to say, 1024, as the entire list was iterated on every wait.   So now, instead of checking every tslot, we specifically only keep a list of those tslots that are in a backoff state - so this loop only costs us when something goes wrong, and not in normal operation.   Also the items in the list correspond to that test, they will be in the list if `!armed()`  (with some adjustments  -eg. NEVER items are not in the list either, as we know those won't set the `nds`.   Also, there could be some whose `next` is set, and the Waiter::ds has ticked over so the transfer should now be checked - and we should call update() on those so `nds` goes to 0, but in fact the old loop didn't do that)

src/backofftimer.cpp
sergiohs84: Oh, I meant the short-named variable `v`, not the method itself :rofl: 
Thanks for the explanatory comments, anyway.

PR #1667, opened 2019-10-15
#13442. Add support to retrieve business pricing plans

jgandres:
**Risk area/s:**
- Information about pricing plans, including business

### DIFF
--- a/src/commands.cpp
+++ b/src/commands.cpp
@@ -2339,8 +2339,8 @@ void CommandUpdatePendingContact::procresult()
 CommandEnumerateQuotaItems::CommandEnumerateQuotaItems(MegaClient* client)
 {
     cmd("utqa");
-    arg("f", 1);
-
+    arg("nf", 1);
+    arg("b", 1);
     tag = client->reqtag;
 }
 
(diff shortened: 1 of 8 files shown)


sergiohs84:
trigger compilation

src/commands.cpp
```diff
@@ -2351,36 +2351,95 @@ void CommandEnumerateQuotaItems::procresult()
...
+                            || !ios
+                            || !android)
+                    {
+                        return client->app->enumeratequotaitems_result(API_EINTERNAL);
+                    }
+
+                    // If pricing plan is business, storage and bandwith are unlimited
+                    if (type == 1)
+                    {
+                        gbstorage = 0;
```
sergiohs84: I think we should honor the value returned by API: `-1` to indicate the storage/bandwidth is unlimited (which is different than `0` --> no allowance at all).
If we do the change above, we need to change the public interfaces of `MegaPricing` for `getGbStorage()` and `getGbBandwidth()`, but I feel the subtle difference between `-1` and `0` is worthy.

src/megaapi_impl.cpp
```diff
@@ -22318,15 +22318,15 @@ int MegaPricingPrivate::getProLevel(int productIndex)
 
 unsigned int MegaPricingPrivate::getGBStorage(int productIndex)
 {
-    if((unsigned)productIndex < gbStorage.size())
+    if((unsigned)productIndex < gbStorage.size() && !type[productIndex])
```
sergiohs84: Since we will change the variables `gbStorage`/`gbBandwidth` to signed integers, we don't need the `type` check anymore.

src/megaapi_impl.cpp
```diff
@@ -22318,15 +22318,15 @@ int MegaPricingPrivate::getProLevel(int productIndex)
...
         return gbStorage[productIndex];
 
     return 0;
 }
 
 unsigned int MegaPricingPrivate::getGBTransfer(int productIndex)
 {
-    if((unsigned)productIndex < gbTransfer.size())
+    if((unsigned)productIndex < gbTransfer.size() && !type[productIndex])
```
sergiohs84: Same than above

src/megaapi_impl.cpp
```diff
@@ -22380,27 +22380,45 @@ const char *MegaPricingPrivate::getAndroidID(int productIndex)
...
+{
+    if((unsigned)productIndex < type.size())
+        return type[productIndex];
+
+    return false;
+}
+
+int MegaPricingPrivate::getAmountMonth(int productIndex)
+{
+    if((unsigned)productIndex < amountMonth.size())
```
sergiohs84: Castings with `static_cast<size_t>`, please.
```suggestion
    if (static_cast<size_t>(productIndex) < amountMonth.size())
```

src/megaapi_impl.cpp
```diff
@@ -22380,27 +22380,45 @@ const char *MegaPricingPrivate::getAndroidID(int productIndex)
     return NULL;
 }
 
+bool MegaPricingPrivate::isBusinessType(int productIndex)
+{
+    if((unsigned)productIndex < type.size())
```
```suggestion
    if (static_cast<size_t>(productIndex) < type.size())
```

sergiohs84:
This PR introduces not only new interfaces at `MegaPricing`, but it also changes the interface for `getGbStorage()` and `getGbBandwidth()`, which will return a signed integer instead of an unsigned integer.
Do not merge it until the apps are ready, so we don't break the compilation.

sergiohs84:
> This PR introduces not only new interfaces at `MegaPricing`, but it also changes the interface for `getGbStorage()` and `getGbBandwidth()`, which will return a signed integer instead of an unsigned integer.
> Do not merge it until the apps are ready or it is confirmed it doesn't create issues with their `develop` code.
> Android: [meganz/android2#987](https://github.com/meganz/android2/pull/987)
> iOS: [meganz/android2#987](https://github.com/meganz/android2/pull/987)
> MEGAsync: (not PR available yet)

- iOS does not retrieve pricing plan details from SDK, but from iTunes store.
- Android is ready, just need to solve some merge conflicts.
- MEGAsync will update the interface before the next release (and probably hide the business plan, unless UI designs are ready for the new plan) (Ticket ID: 12620)

PR #1670, opened 2019-10-16
#13567: isSyncing to return if any sync is in syncing state

polmr:
convenient function. No risk areas

### DIFF
--- a/src/megaapi_impl.cpp
+++ b/src/megaapi_impl.cpp
@@ -1355,7 +1355,7 @@ bool MegaApiImpl::isSyncing()
 
     for (auto & sync : client->syncs)
     {
-        if (sync->localroot->ts == TREESTATE_SYNCING)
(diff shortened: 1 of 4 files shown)


src/megaapi_impl.cpp
```diff
@@ -1345,6 +1345,24 @@ bool MegaApiImpl::isIndexing()
...
+{
+    if (!client->syncs.size())
+    {
+        return false;
+    }
+    SdkMutexGuard g(sdkMutex);
+
+    for (auto & sync : client->syncs)
+    {
+        if (sync->localroot->ts == TREESTATE_SYNCING)
```
sergiohs84: Not sure, but you might want to consider the abnormal state `TREESTATE_PENDING` here too, since it's also syncing, but waiting for stable `mtime`/`size` or waiting for the upload to complete and put the node.

It all depends on the planned usage for this new function.

polmr: agree

PR #1671, opened 2019-10-17
#13592. Avoid unnecesary syncdown if API fails with overquota

jgandres:
**Risk area/s:**
- Sync transfers failed with overquota or Business account expired

### DIFF
--- a/include/megaapi.h
+++ b/include/megaapi.h
@@ -12801,7 +12801,7 @@ class MegaApi
         /**
          * @brief Get the user relative to an incoming share
          *
-         * This function will return NULL if the node is not found.
+         * This function will return NULL if the node is not found
          *
          * If recurse is true, it will return NULL if the root corresponding to
          * the node received as argument doesn't represent the root of an incoming share.
--- a/include/megaapi.h
+++ b/include/megaapi.h
@@ -12783,15 +12783,20 @@ class MegaApi
         /**
          * @brief Get the user relative to an incoming share
          *
-         * This function will return NULL if the node is not found or doesn't represent
+         * This function will return NULL if the node is not found.
+         *
+         * If recurse is true, it will return NULL if the root corresponding to
+         * the node received as argument doesn't represent the root of an incoming share.
+         * Otherwise, it will return NULL if the node doesn't represent
          * the root of an incoming share.
          *
          * You take the ownership of the returned value
          *
-         * @param node Incoming share
+         * @param node Node to look for inshare user.
+         * @param recurse use root node corresponding to the node passed
          * @return MegaUser relative to the incoming share
          */
-        MegaUser *getUserFromInShare(MegaNode *node);
+        MegaUser *getUserFromInShare(MegaNode *node, bool recurse = false);
 
         /**
           * @brief Check if a MegaNode is being shared by/with your own user
--- a/include/megaapi.h
+++ b/include/megaapi.h
@@ -12803,10 +12803,10 @@ class MegaApi
          *
          * This function will return NULL if the node is not found
          *
-         * If recurse is true, it will return NULL if the root corresponding to
-         * the node received as argument doesn't represent the root of an incoming share.
-         * Otherwise, it will return NULL if the node doesn't represent
-         * the root of an incoming share.
+         * When recurse is true and the root of the specified node is not an incoming share,
+         * this function will return NULL.
+         * When recurse is false and the specified node doesn't represent the root of an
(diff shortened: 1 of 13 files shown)


jgandres:
trigger compilation

src/node.cpp
```diff
@@ -1364,6 +1364,11 @@ void LocalNode::init(Sync* csync, nodetype_t ctype, LocalNode* cparent, string*
     if (cparent)
     {
         setnameparent(cparent, cfullpath);
+
+        if (parent->node)
+        {
+            h = parent->node->nodehandle;
```
mattw-mega: I think we should do this assignment at megaclient.cpp:13151  instead, at the moment the PUT transfer is started, to ensure we pick the up to date handle (as actionpackets may later rearrange the parent pointers etc). thanks!

sergiohs84: I agree. Thanks, Matt ;)

src/sync.cpp
```diff
@@ -640,7 +640,10 @@ Sync::~Sync()
 
     // Create a committer to ensure we update the transfer database in an efficient single commit,
     // if there are transactions in progress.
-    DBTableTransactionCommitter committer(client->tctable);
+    if (client)
+    {
+        DBTableTransactionCommitter committer(client->tctable);
```
sergiohs84: This commiter becomes out of scope in the next line, so it will not be valid at the point where it's needed (`proctree()` and the delete of `localroot`. 
Since it's very critical to manage the lifetime of be object to properly manage the transaction, you must avoid a `unique_ptr`. There are 2 options, create one with `new` and ensure you call `delete`, or simply declare a committer in each place as required

src/transfer.cpp
```diff
@@ -401,6 +401,10 @@ void Transfer::failed(error e, DBTableTransactionCommitter& committer, dstime ti
                 }
             }
         }
+        else if (e == API_EARGS)
+        {
+            client->app->transfer_failed(this, e);
+        }
         else
```
sergiohs84: Maybe here we could do the:
```else if (e != API_EBUSINESSPASTDUE)
```
It looks like easier to read, but not a biggie.

src/transfer.cpp
```diff
@@ -423,6 +427,24 @@ void Transfer::failed(error e, DBTableTransactionCommitter& committer, dstime ti
             continue;
         }
 
+        /*
+         * If The transfer failed with API_EARGS and the file is part of a sync transfer,
+         * defer transfer, otherwise remove file from transfer.
```
sergiohs84:
```
If the transfer failed with API_EARGS, the target handle is invalid. For a sync-transfer,
the actionpacket will eventually remove the target and the sync-engine will force to 
disable the synchronization of the folder. For non-sync-transfers, remove the file directly.
```

include/megaapi.h
```diff
@@ -12801,15 +12801,20 @@ class MegaApi
         /**
          * @brief Get the user relative to an incoming share
          *
-         * This function will return NULL if the node is not found or doesn't represent
+         * This function will return NULL if the node is not found.
```
sergiohs84:
```suggestion
         * This function will return NULL if the node is not found
```

include/megaapi.h
```diff
@@ -12801,15 +12801,20 @@ class MegaApi
...
+         *
+         * If recurse is true, it will return NULL if the root corresponding to
+         * the node received as argument doesn't represent the root of an incoming share.
+         * Otherwise, it will return NULL if the node doesn't represent
          * the root of an incoming share.
```
sergiohs84:
```
         * When recurse is true and the root of the specified node is not an incoming share,
         * this function will return NULL.
         * When recurse is false and the specified node doesn't represent the root of an 
         * incoming share, this function will return NULL.
```

src/sync.cpp
```diff
@@ -656,8 +659,7 @@ Sync::~Sync()
         client->syncactivity = true;
 
         {
-            // now recursively delete all the associated LocalNodes, and their associated transfer and file objects. 
-            // If any have transactions in progress, the committer will ensure we update the transfer database in an efficient single commit.
+            // Create a committer and recursively delete all the associated LocalNodes, and their associated transfer and file objects.
```
sergiohs84:
```suggestion
            // Create a committer and recursively delete all the associated LocalNodes, and their associated transfer and file objects.
            // If any have transactions in progress, the committer will ensure we update the transfer database in an efficient single commit.
```

PR #1677, opened 2019-10-21
#13636. Allow params embedded in public links

sergiohs84:
Some websites modify MEGA public links to include tracking params. We need to skip them in order to parse the public-handle and the key.

This PR also solves the ticket #13637, which fixes the public links protected by password.

**Risk area/s**:
- Create/open/import/download a file/folder link with or without key, and password-protected links.


aabellagm:
trigger compilation

sergiohs84:
trigger compilation

PR #1679, opened 2019-10-21
#13644 Fix for setCameraUploadsFolder and setCameraUploadsFolderSecondary.

mattw-mega:
The secondary version should have request->getFlag() return true
And request->getNodeHandle() should return the relevant handle for both versions.
Some cases were missing when the value was already present, or when the value was being overwritten.


mattw-mega:
trigger compilation

mattw-mega:
I agree, that's much simpler.   I've approved that one and will close this one.  cheers

PR #1680, opened 2019-10-22
#13651. Efficient removal of sync transfers from database when a sync is remo…

mattw-mega:
…ved.

Closing a sync means we need to remove any active transfers from that sync from the transfer table in the database.
This change means we will just use one begin/commit to update all those transfers, which means it's much faster.

### DIFF
--- a/src/sync.cpp
+++ b/src/sync.cpp
@@ -411,7 +411,7 @@ int computeReversePathMatchScore(const string& path1, const string& path2, const
 bool assignFilesystemIds(Sync& sync, MegaApp& app, FileSystemAccess& fsaccess, handlelocalnode_map& fsidnodes,
                          const string& localdebris, const string& localseparator)
 {
-    auto rootpath = sync.localroot.localname;
+    auto rootpath = sync.localroot->localname;
 
     auto fa = fsaccess.newfileaccess(false);
     if (!fa->fopen(&rootpath, true, false))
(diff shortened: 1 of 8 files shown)


src/sync.cpp
```diff
@@ -436,11 +436,11 @@ bool assignFilesystemIds(Sync& sync, MegaApp& app, FileSystemAccess& fsaccess, h
     // Ensures that unmatched nodes (local nodes that don't have a fingerprint that's
     // the same as a file on disk) have invalid IDs.
     size_t invalidatedCount = 0;
-    invalidateFilesystemIds(fsidnodes, sync.localroot, invalidatedCount);
+    invalidateFilesystemIds(fsidnodes, *sync.localroot, invalidatedCount);
```
sergiohs84: Why not with `sync->localroot`?? here and 4 lines below too

sergiohs84:
Remember to create and associate a ticket, please :pray: 
Thanks

src/sync.cpp
mattw-mega: `localroot` is a pointer now, so that we can destroy its object separately in ~Sync (with the db transaction around that).  But the function called here takes a LocalNode& so we need to dereference it now.  cheers

sergiohs84: Oh, right. I thought the pointer here was the `sync` and that `invalidateFilesystemIds()` too a pointer to `LocalNode`. Nothing else to say :)

PR #1685, opened 2019-10-23
#13646 Allow to resend email

jgandres:


src/megaapi_impl.cpp
```diff
@@ -12021,6 +12030,15 @@ void MegaApiImpl::getemaillink_result(error e)
     fireOnRequestFinish(request, MegaError(e));
 }
 
+void MegaApiImpl::resendverificationemail_result(error e)
+{
+    if (requestMap.find(client->restag) == requestMap.end()) return;
+    MegaRequestPrivate *request = requestMap.at(client->restag);
```
sergiohs84: There is a new way to avoid multiple lookups in the `requestMap`. Please, keep the iterator returned by `find()` and use it to get the pointer to the `request`.

src/commands.cpp
```diff
@@ -5450,6 +5450,25 @@ void CommandConfirmCancelLink::procresult()
...
+{
+    cmd("era");
+    tag = client->reqtag;
+}
+
+void CommandResendVerificationEmail::procresult()
+{
+    if (client->json.isnumeric())
+    {
+        return client->app->resendverificationemail_result((error)client->json.getint());
```
sergiohs84: This method returns `void`. Actually, you can leave it fall through, and it will return anyway (as it is the code today).

PR #1722, opened 2019-11-06
#13679. Add startUploadForSupport that uploads to support inbox

polmr:
Have user handle gatthered from pubk response used when putting nodes

**Notes**
- Folder upload is not possible when sending to Inbox:
   API won't return handle for folder creation, and even if that was the case
   doing a put nodes with t = userhandle & the corresponding handle as parent p, API returns EACCESS

**Risk areas**
- Uploads & transfers in general

### DIFF
--- a/src/commands.cpp
+++ b/src/commands.cpp
@@ -3560,12 +3560,9 @@ void CommandPubKeyRequest::procresult()
                     if (!ISUNDEF(uh))
                     {
                         client->mapuser(uh, u->email.c_str());
-                        if (u->isTemporary && u->uid == u->email)
+                        if (u->isTemporary && u->uid == u->email) //update uid with the received USERHANDLE (will be used as target for putnodes)
                         {
-                            char uid[12];
-                            Base64::btoa((byte*)&uh, MegaClient::USERHANDLE, uid);
-                            uid[11] = '\0';
-                            u->uid = uid;
+                            u->uid = Base64Str<MegaClient::USERHANDLE>(uh);
                         }
                     }
 
--- a/src/commands.cpp
+++ b/src/commands.cpp
@@ -3560,6 +3560,13 @@ void CommandPubKeyRequest::procresult()
                     if (!ISUNDEF(uh))
                     {
                         client->mapuser(uh, u->email.c_str());
+                        if (u->isTemporary && u->uid == u->email)
+                        {
+                            char uid[12];
+                            Base64::btoa((byte*)&uh, MegaClient::USERHANDLE, uid);
+                            uid[11] = '\0';
+                            u->uid = uid;
+                        }
                     }
 
                     if (client->fetchingkeys && u->userhandle == client->me && len_pubk)
--- a/src/megaapi_impl.cpp
+++ b/src/megaapi_impl.cpp
@@ -17559,7 +17559,7 @@ unsigned MegaApiImpl::sendPendingTransfers()
                 const char *inboxTarget = uploadToInbox ? transfer->getParentPath() : nullptr;
 
                 if (!localPath || !fileName || !(*fileName)
-                        || (!inboxTarget && (!parent || parent->type == FILENODE) ) )
+                        || (!uploadToInbox && (!parent || parent->type == FILENODE) ) )
                 {
                     e = API_EARGS;
                     break;
@@ -17662,7 +17662,7 @@ unsigned MegaApiImpl::sendPendingTransfers()
                                 tc.nn->ovhandle = client->getovhandle(parent, &sname);
                             }
 
-                            if (inboxTarget)
+                            if (uploadToInbox)
                             {
                                 client->putnodes(inboxTarget, tc.nn, nc);
                             }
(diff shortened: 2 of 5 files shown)


src/commands.cpp
```diff
@@ -3560,6 +3560,13 @@ void CommandPubKeyRequest::procresult()
                     if (!ISUNDEF(uh))
                     {
                         client->mapuser(uh, u->email.c_str());
+                        if (u->isTemporary && u->uid == u->email)
+                        {
+                            char uid[12];
+                            Base64::btoa((byte*)&uh, MegaClient::USERHANDLE, uid);
+                            uid[11] = '\0';
+                            u->uid = uid;
```
sergiohs84: We have a much nicer interface to convert from Binary to B64: 
```
u->uid = Base64Str<MegaClient::USERHANDLE>(uh);
```
:)

src/megaapi_impl.cpp
```diff
@@ -7934,6 +7934,11 @@ void MegaApiImpl::startUpload(bool startFirst, const char *localPath, MegaNode *
         transfer->setParentHandle(parent->getHandle());
     }
 
+    if (targetUser)
+    {
+        transfer->setParentPath(targetUser); //TODO: copy?
```
sergiohs84: Yes, `setParentPath()` does a copy of the input parameter. What do you mean by `//TODO`??

src/megaapi_impl.cpp
```diff
@@ -7968,6 +7976,11 @@ void MegaApiImpl::startUpload(const char *localPath, MegaNode *parent, int64_t m
 void MegaApiImpl::startUpload(const char* localPath, MegaNode* parent, const char* fileName, MegaTransferListener *listener)
 { return startUpload(false, localPath, parent, fileName, -1, 0, false, NULL, false, false, listener); }
 
+void MegaApiImpl::startUploadForSupport(const char *localPath, MegaTransferListener *listener)
+{
+    return startUpload(false, localPath, nullptr, nullptr, "supportdrop@mega.nz", -1, 0, false, nullptr, false, false, listener);
```
sergiohs84: Considering the relevance of this upload, perhaps we want to use the `startFirst = true`, indeed.
Also, it could be useful to give the ownership of the file to the SDK, so it deletes it after the upload, by setting the `isSourceFileTemporary = true`.
```suggestion
    return startUpload(true, localPath, nullptr, nullptr, "supportdrop@mega.nz", -1, 0, false, nullptr, true, false, listener);
```

src/megaapi_impl.cpp
```diff
@@ -17650,7 +17663,7 @@ unsigned MegaApiImpl::sendPendingTransfers()
 
                     currentTransfer = transfer;                    
                     string wFileName = fileName;
-                    MegaFilePut *f = new MegaFilePut(client, &wLocalPath, &wFileName, transfer->getParentHandle(), "", mtime, isSourceTemporary);
+                    MegaFilePut *f = new MegaFilePut(client, &wLocalPath, &wFileName, transfer->getParentHandle(), transfer->getParentPath()?transfer->getParentPath():"", mtime, isSourceTemporary);
```
```suggestion
                    MegaFilePut *f = new MegaFilePut(client, &wLocalPath, &wFileName, transfer->getParentHandle(), transfer->getParentPath() ? transfer->getParentPath() : "", mtime, isSourceTemporary);
```

src/megaapi_impl.cpp
polmr: `startFirst` is a good idea. For FileTemporary, let's keep it false: we might want to upload something need to stay alive afterwards.

sergiohs84:
trigger compilation

src/megaapi_impl.cpp
```diff
@@ -17542,7 +17555,11 @@ unsigned MegaApiImpl::sendPendingTransfers()
                 Node *parent = client->nodebyhandle(transfer->getParentHandle());
                 bool startFirst = transfer->shouldStartFirst();
 
-                if (!localPath || !parent || parent->type == FILENODE || !fileName || !(*fileName))
+                bool uploadToInbox = !parent && transfer->getParentPath() && strchr(transfer->getParentPath(), '@');
+                const char *inboxTarget = uploadToInbox ? transfer->getParentPath() : nullptr;
+
+                if (!localPath || !fileName || !(*fileName)
+                        || (!inboxTarget && (!parent || parent->type == FILENODE) ) )
```
```suggestion
                        || (!uploadToInbox && (!parent || parent->type == FILENODE) ) )
```

src/megaapi_impl.cpp
```diff
@@ -17636,7 +17661,15 @@ unsigned MegaApiImpl::sendPendingTransfers()
                             {
                                 tc.nn->ovhandle = client->getovhandle(parent, &sname);
                             }
-                            client->putnodes(parent->nodehandle, tc.nn, nc);
+
+                            if (inboxTarget)
```
```suggestion
                            if (uploadToInbox)
```

src/megaapi_impl.cpp
```diff
@@ -17650,7 +17683,7 @@ unsigned MegaApiImpl::sendPendingTransfers()
 
                     currentTransfer = transfer;                    
                     string wFileName = fileName;
-                    MegaFilePut *f = new MegaFilePut(client, &wLocalPath, &wFileName, transfer->getParentHandle(), "", mtime, isSourceTemporary);
+                    MegaFilePut *f = new MegaFilePut(client, &wLocalPath, &wFileName, transfer->getParentHandle(), inboxTarget ? inboxTarget : "", mtime, isSourceTemporary);
```
```suggestion
                    MegaFilePut *f = new MegaFilePut(client, &wLocalPath, &wFileName, transfer->getParentHandle(), uploadToInbox ? inboxTarget : "", mtime, isSourceTemporary);
```

PR #1724, opened 2019-11-07
#13781. Putnodes tctable transaction, megacli logging improvements, factoring.

mattw-mega:
CMake updates for linking with dynamic CRT on windows
megacli support for specifying a file to log to with the `debug` commands.cpp
Additional report logging for cs requests when MEGA_MEASURE_CODE is on
Remove unused function nextpos()
Added error code DAEMON_EFAILED = -4, use of which makes the code much clearer
Make sure we begin a tctable transaction for PutNodes procresult
Factor some code into checkTransferFinished(), (preparing for changes in that area)
Use the more compact form of sendevent() in transferslot.cpp (preparing for changes in that area)


mattw-mega:
trigger compilation

PR #1726, opened 2019-11-07
#13784. Improve tctable update times when executing many same-type requests

mattw-mega:
Start one DBTableTransactionCommitter that will be used for consecutive loops in sendPendingRequests().
The transaction is committed if we exit the loop, process 1024 requests, or a different type of request is processed (and a new transaction will be started if there is any tctable activity)
This improvement makes a really huge difference for things like pausing or cancelling all transfers when there are a large number of them.  Those actions are almost instant now (in combination with some other PRs under review).


mattw-mega:
trigger compilation

mattw-mega:
trigger compilation

src/megaapi_impl.cpp
```diff
@@ -18186,16 +18186,37 @@ void MegaApiImpl::sendPendingRequests()
...
+
+
+    while(MegaRequestPrivate *request = requestQueue.front())
     {
-        sdkMutex.lock();
 
-        MegaRequestPrivate *request = requestQueue.pop();
-        if (!request)
+        // also we avoid yielding for consecutive transaction cancel operations (we used to yeild every time, but we need to keep the sdkMutex lock while the database transaction is ongoing)
+        if ((lastRequestType == -1 || lastRequestType == request->getType()) && lastRequestConsecutive < 1024)
```
sergiohs84: Why do we check the `lastRequestType` here? :thinking: 
Perhaps it's cleaner to reset the value when we break the loop.

src/megaapi_impl.cpp
```diff
@@ -18186,16 +18186,37 @@ void MegaApiImpl::sendPendingRequests()
...
+        {
+            committer.commitNow();
             sdkMutex.unlock();
+            yield();
+            sdkMutex.lock();
+            lastRequestConsecutive = 0;
+        }
+
+        request = requestQueue.pop();
+        if (!request)
```
sergiohs84: If the `front()` didn't return `null`, the `pop()` will neither. This check is unnecessary.

polmr: why not ` while(MegaRequestPrivate *= requestQueue.pop())` ?

mattw-mega: Yes, I think that arrangement was left over from a version where we might exit the loop before all requests were processed (to give the SDK a go periodically on this thread, but we decided against that for now in a long-ago review, I think).    I'll make this simplification, thank you both for the review!  cheers

src/megaapi_impl.cpp
mattw-mega: The change from the comment below will fix this up - we won't break the loop at all, but rather it finishes when we run out of items on the request queue - so this line is specifically to decide if we need to commit and start a new database transaction (but lazily, so we only actually begin the transaction if any database work gets started).   And we do that if we are processing a different request type to the last loop iteration (if there was one), or if we have done 1024 of the same request type in a row.  thanks!

sergiohs84: Oh, I see now. My bad, I didn't notice, at first, that a different type of request should commit the ongoing transaction. All good :)

PR #1727, opened 2019-11-07
#12290 Add syncable flag to Node and LocalNode

cb-mega:
This is in preparation for one-way-sync. The syncable flag of both `Node` and `LocalNode` is not used apart from serialization and deserialization.

The client is now constructed in unit tests so checks for a null client are now removed from our production code, client is always valid.


include/mega/node.h
```diff
@@ -209,53 +209,62 @@ struct MEGA_API Node : public NodeCore, FileFingerprint
...
     // parent
-    Node* parent;
+    Node* parent = nullptr;
 
     // children
     node_list children;
 
     // own position in parent's children
-    node_list::iterator child_it;
+    node_list::iterator child_it{};
```
mattw-mega: Do we need {} here?  Since it's an object of a class, if it has a default constructor then it would be called anyway.

include/mega/node.h
```diff
@@ -209,53 +209,62 @@ struct MEGA_API Node : public NodeCore, FileFingerprint
...
     // handle of public link for the node
-    PublicLink *plink;
+    PublicLink* plink = nullptr;
 
     void setpubliclink(handle, m_time_t, m_time_t, bool);
 
     bool serialize(string*);
     static Node* unserialize(MegaClient*, string*, node_vector*);
 
+    Node();
```
mattw-mega: Do we need a default constructor for node?  If used, it would presumably have a null `client` pointer, which is supposed to never happen.

include/mega/node.h
```diff
@@ -291,6 +300,9 @@ struct MEGA_API LocalNode : public File
     // stored to rebuild tree after serialization => this must not be a pointer to parent->dbid
     int32_t parent_dbid = 0;
 
+    // whether this node can be synced to the remote tree
+    bool syncable = true;
```
mattw-mega: `syncable` and `mSyncable`?  perhaps these two could be distinguished a bit more - or should there only be one?

src/node.cpp
```diff
@@ -338,7 +338,10 @@ Node* Node::unserialize(MegaClient* client, string* d, node_vector* dp)
     hasLinkCreationTs = MemAccess::get<char>(ptr);
     ptr += sizeof(hasLinkCreationTs);
 
-    for (i = 6; i--;)
+    const bool syncable = MemAccess::get<bool>(ptr);
```
mattw-mega: from cppreference:  "bool - type, capable of holding one of the two values: true or false. The value of sizeof(bool) is implementation defined and might differ from 1."

src/node.cpp
```diff
@@ -529,7 +533,9 @@ bool Node::serialize(string* d)
     char hasLinkCreationTs = plink ? 1 : 0;
     d->append((char*)&hasLinkCreationTs, 1);
 
-    d->append("\0\0\0\0\0", 6);
+    d->append((char*)&mSyncable, sizeof(mSyncable));
```
mattw-mega: Can we switch Node's serialize and unserialize to use CacheableReader/CacheableWriter which is much easier to maintain, since we are making changes in these functions?  thanks

src/node.cpp
```diff
@@ -1078,6 +1084,18 @@ Node* Node::firstancestor()
...
 
+#ifdef ENABLE_SYNC
+void Node::setSyncable(const bool syncable)
+{
+    mSyncable = syncable;
+}
+
+bool Node::isSyncable() const
+{
+    return mSyncable;
```
mattw-mega: since the getter/setter functions don't have any side effects, isn't this the same as just having the member public, but with extra steps? ;-)

src/node.cpp
```diff
@@ -1863,6 +1901,22 @@ LocalNode* LocalNode::unserialize(Sync* sync, string* d)
...
+        syncable = MemAccess::get<bool>(ptr);
+        ptr += sizeof(syncable);
+
+        for (int i = 7; i--;)
+        {
+            if (ptr + (unsigned char)*ptr < end)
+            {
+                ptr += (unsigned char)*ptr + 1;
+            }
+        }
```
mattw-mega: this function would benefit a lot from using CacheableReader/CacheableWriter too

src/node.cpp
cb-mega: Wouldn't that require a lot of changes? I am not sure this is necessary right now

src/node.cpp
cb-mega: I used setters and getters here because they will have side effects later on in future one-way-sync PRs. So please bare with me :)

include/mega/node.h
cb-mega: what do you mean? syncable on `LocalNode` refers to syncing to the remote. syncable on `Node` refers to syncing to local.

include/mega/node.h
cb-mega: I suppose we don't need it. Will remove

include/mega/node.h
cb-mega: That's just needed for unit testing similar to `LocalNode` and `Sync`. I will later try to refactor the tests to not require that but it may be hard.. Is this ok for now?

src/node.cpp
cb-mega: I didn't know it could be different from 1 but do you see a problem here?

cb-mega: Ah got you. We'd have to take up more extension bytes of course. Ok, hm. Maybe best to write out an 8bit integer then. I'll change that.

include/mega/node.h
mattw-mega: ah, I see - the gihub diff layout made it look like these two were in the same class

src/node.cpp
mattw-mega: I think it would make it a lot easier to review (as the current format is quite fiddly which makes it easy to miss problems), and easier to maintain in future.  It would also mean we automatically avoid issues with the sizes of types etc.

include/mega/node.h
cb-mega: Removed.

src/node.cpp
```diff
@@ -338,7 +335,10 @@ Node* Node::unserialize(MegaClient* client, string* d, node_vector* dp)
     hasLinkCreationTs = MemAccess::get<char>(ptr);
     ptr += sizeof(hasLinkCreationTs);
 
-    for (i = 6; i--;)
+    const auto syncableInt = MemAccess::get<int8_t>(ptr);
```
mattw-mega: Shouldn't nodes serialized by prior SDK versions and deserialized by this code end up with syncable == true?   I think they would end up with that field false here?  thanks

cb-mega: Good point. Will fix

src/node.cpp
```diff
@@ -1800,6 +1825,11 @@ bool LocalNode::serialize(string* d)
     d->append((char*)&ll, sizeof ll);
     d->append(localname.data(), ll);
 
+    const auto syncableInt = static_cast<int8_t>(!syncable);
```
mattw-mega: perhaps a formulation such as `char syncableInt = syncable ? 0 : 1;`  similar to lines above would be a lot clearer?
Also, on reanalysis of this and the corresponding deserializer,  this code serializing and then the older SDK code deserializing would break because the old code expects not just a flag, but a number of bytes to skip over.   It would read the new flag as 1, skip 1 byte further, and expect 5 more flags (actually the number of bytes to skip each time), but there would only be 4, and everything after would be misaligned.  thanks

src/node.cpp
```diff
@@ -1831,6 +1861,13 @@ LocalNode* LocalNode::unserialize(Sync* sync, string* d)
     m_off_t size = MemAccess::get<m_off_t>(ptr);
     ptr += sizeof(m_off_t);
 
+    const bool hasExtensionBytes = size == LOCALNODE_EXTENSION_BYTES_FLAG;
```
mattw-mega: I think `LocalNode::unserialize` is only called from `Sync::readstatecache`, and it passes an independent string from the database each time.   The old code does not check if it got to the end of the string, so we could just append new fields, and older SDK versions will ignore them.   I think we can get away without the extension bytes flag, and retain backward compatibility?  thanks

cb-mega: We could do that if `mtime` was serialized the same way as other fields with a fixed number of bytes. However, `Serialize64::serialize` only writes out as many bytes as it needs so if we appended to the end then we wouldn't know when `mtime` ends and our extension begins?

include/mega/node.h
sergiohs84: The new member, anyway, should have been named `mSyncable` too, @cb-mega , don't you think so?

PR #1728, opened 2019-11-07
 #13795. Don't timeout a whole transfer if some large chunks are making progress

mattw-mega:
There are separate timeouts per chunk (cURL checks for insufficient data rate per connection).
Only time out the whole transfer if no chunks are moving
Also factored some chunkmac_map loops into member functions of chunkmac_map
Also moved chunkmac_map declaration to Utils.h since its members are defined in Utils.cpp
Also allowed useragent to be specified by command line for integration tests

### DIFF
--- a/include/mega/raid.h
+++ b/include/mega/raid.h
@@ -23,6 +23,7 @@
 #define MEGA_RAID_H 1
 
 #include "http.h"
+#include "utils.h"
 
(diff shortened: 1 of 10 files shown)


mattw-mega:
trigger compilation

mattw-mega:
trigger compilation

PR #1729, opened 2019-11-08
Ensure that access to MegaLoggers is thread-safe in log performance mode

cb-mega:
This comes at the cost of logging a mutex every time we issue a log but is necessary if we want to add/remove MegaLoggers while actively logging.


polmr:
This will not be required for the moment.

PR #1730, opened 2019-11-08
#13796. Adjust name for singup links

sergiohs84:
Webclient uses the fullname here, so native apps should behave the same way.

### DIFF
--- a/include/megaapi.h
+++ b/include/megaapi.h
@@ -7834,7 +7834,7 @@ class MegaApi
          * The associated request type with this request is MegaRequest::TYPE_SEND_SIGNUP_LINK.
          *
          * @param email Email for the account
-         * @param name Fullname of the user (firstname + fullname)
+         * @param name Fullname of the user (firstname + lastname)
          * @param password Password for the account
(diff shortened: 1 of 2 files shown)


include/megaapi.h
```diff
@@ -7834,7 +7834,7 @@ class MegaApi
          * The associated request type with this request is MegaRequest::TYPE_SEND_SIGNUP_LINK.
          *
          * @param email Email for the account
-         * @param name Firstname of the user
+         * @param name Fullname of the user (firstname + fullname)
```
jnavarrom:
```suggestion
         * @param name Fullname of the user (firstname + lastname)
```

include/megaapi.h
```diff
@@ -7847,7 +7847,7 @@ class MegaApi
          * email address, in case the user mistyped the email at the registration form.
          *
          * @param email Email for the account
-         * @param name Firstname of the user
+         * @param name Fullname of the user (firstname + fullname)
```
jnavarrom:
```suggestion
         * @param name Fullname of the user (firstname + lastname)
```

PR #1735, opened 2019-11-11
Remove full paths for sources in logs (when ENABLE_LOG_PERFORMANCE)

polmr:


include/mega/logging.h
```diff
@@ -292,7 +292,15 @@ class SimpleLogger
 #endif
     {
 #ifdef ENABLE_LOG_PERFORMANCE
-        logValue(filename);
+        const char * actualFileName = strrchr(filename, '/'); //The project part of the path uses `/` for Windows too.
+        if (actualFileName)
```
sergiohs84: Perhaps protect against filenames like: `mypath/to/some/folder/`:
```suggestion
        if (actualFileName && *actualFileName != '\0')
```

PR #1744, opened 2019-11-12
#13826. Fix serialization + skip new member

sergiohs84:
- Do not use more than one extension byte for the new flag
- Add the new member to Node only for sync-enabled clients


sergiohs84:
This branch has been superseded by https://github.com/meganz/sdk/pull/1745

PR #1745, opened 2019-11-12
#13826 Use ENABLE_SYNC for mSyncable; use char instead of int8_t for consistency

cb-mega:
This is PR does the following:
- Remove `mSyncable` de-/serialization from `Node` (this will be managed in a different data structure)
- Use `ENABLE_SYNC` for `Node`s `mSyncable`
- Use char instead of int8_t for consistency

### DIFF
--- a/include/mega/node.h
+++ b/include/mega/node.h
@@ -262,9 +262,10 @@ struct MEGA_API Node : public NodeCore, FileFingerprint
     ~Node();
 
 private:
+#ifdef ENABLE_SYNC
     // whether this node can be synced to the local tree
     bool mSyncable = true;
-
+#endif
     // full folder/file key, symmetrically or asymmetrically encrypted
     // node crypto keys (raw or cooked -
     // cooked if size() == FOLDERNODEKEYLENGTH or FILEFOLDERNODEKEYLENGTH)
--- a/include/mega/node.h
+++ b/include/mega/node.h
@@ -301,7 +301,7 @@ struct MEGA_API LocalNode : public File
     int32_t parent_dbid = 0;
 
     // whether this node can be synced to the remote tree
-    bool syncable = true;
+    bool mSyncable = true;
 
     // children by name
     localnode_map children;
--- a/include/mega/node.h
+++ b/include/mega/node.h
@@ -236,12 +236,6 @@ struct MEGA_API Node : public NodeCore, FileFingerprint
     // location in the tounlink node_set
     // FIXME: merge todebris / tounlink
     node_set::iterator tounlink_it;
-
-    // sets whether this node can be synced to the local tree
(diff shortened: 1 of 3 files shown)


include/mega/node.h
```diff
@@ -262,9 +262,10 @@ struct MEGA_API Node : public NodeCore, FileFingerprint
     ~Node();
 
 private:
+#ifdef ENABLE_SYNC
     // whether this node can be synced to the local tree
     bool mSyncable = true;
```
polmr: This is no longer used. And ENABLE_SYNC macros should be moved to LocalNode instead. And control deserialization in that case, to skip the corresponding byte without assigning its value

src/node.cpp
```diff
@@ -1925,16 +1906,16 @@ LocalNode* LocalNode::unserialize(Sync* sync, const string* d)
...
-    if (ptr < end)
+    char syncable = 1;
+    if (ptr + 1 < end)
     {
-        const bool hasSyncableInt = (unsigned char)*ptr == sizeof(syncableInt);
+        const bool hasSyncable = (unsigned char)*ptr == sizeof(syncable);
         ptr += 1;
 
-        if (hasSyncableInt)
+        if (hasSyncable && ptr + 1 < end)
```
polmr:
```suggestion
        if (hasSyncable && ptr + sizeof(syncable) < end)
```

include/mega/node.h
sergiohs84: It is still used, indeed, by the `Node::setSyncable()` and `Node::isSyncable()`, which I guess Christian will eventually use for the one-way-sync.

sergiohs84: Furthermore, the `LocalNode::mSyncable` is already between the macros, as it is the full `LocalNode` declaration :)

src/node.cpp
```diff
@@ -1837,12 +1818,10 @@ bool LocalNode::serialize(string* d)
...
 
-    const int8_t syncableInt = syncable ? 1 : 0;
-    const int8_t syncableIntByteCount = sizeof(syncableInt);
-    d->append((char*)&syncableIntByteCount, 1);
-    d->append((char*)&syncableInt, syncableIntByteCount);
+    const char syncable = mSyncable ? 1 : 0;
+    d->append(&syncable, sizeof(syncable));
 
-    d->append("\0\0\0\0\0\0", 7); // Use these bytes for extensions
+    d->append("\0\0\0\0\0\0", 8); // Use these bytes for extensions
```
sergiohs84: We think that better to add 8 bytes for extensions, rather than 7, now that we include them for localnodes too :)

src/node.cpp
sergiohs84: Corrected already! Thanks

PR #1746, opened 2019-11-13
#13836. Sync config

cb-mega:
This adds the `SyncConfig` object. Remains unused until the remaining one-way-sync functionality is implemented.

### DIFF
--- a/include/mega/types.h
+++ b/include/mega/types.h
@@ -678,7 +678,7 @@ struct SyncConfig
     {
         TYPE_UP = 0x01, // sync up from local to remote
         TYPE_DOWN = 0x02, // sync down from remote to local
-        TYPE_DEFAULT = TYPE_UP | TYPE_DOWN,
+        TYPE_DEFAULT = TYPE_UP | TYPE_DOWN, // Two-way sync
     };
 
     // type of the sync, defaults to bidirectional
--- a/include/mega/types.h
+++ b/include/mega/types.h
@@ -671,6 +671,26 @@ namespace CodeCounter
     };
 }
 
+// Holds the config of a sync. Can be extended with future config options
+struct SyncConfig
+{
+    enum Type
+    {
+        TYPE_UP = 0x01, // sync up from local to remote
+        TYPE_DOWN = 0x02, // sync down from remote to local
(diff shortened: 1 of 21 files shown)


include/mega/types.h
```diff
@@ -671,6 +671,26 @@ namespace CodeCounter
...
+        TYPE_UP = 0x01, // sync up from local to remote
+        TYPE_DOWN = 0x02, // sync down from remote to local
+        TYPE_DEFAULT = TYPE_UP | TYPE_DOWN, // Two-way sync
+    };
+
+    // type of the sync, defaults to bidirectional
+    Type syncType = TYPE_DEFAULT;
+
+    // whether deletions are synced (only relevant for one-way-sync)
+    bool syncDeletions = false;
```
mattw-mega: wouldn't syncDeletions be true for two-way sync?  I it would probably keep the code simpler if we don't have to check the sync type before checking this flag?  And likewise for the one below?  thanks

src/sync.cpp
```diff
@@ -513,9 +513,10 @@ bool assignFilesystemIds(Sync& sync, MegaApp& app, FileSystemAccess& fsaccess, h
 
 // new Syncs are automatically inserted into the session's syncs list
 // and a full read of the subtree is initiated
-Sync::Sync(MegaClient* cclient, string* crootpath, const char* cdebris,
+Sync::Sync(MegaClient* cclient, SyncConfig config, string* crootpath, const char* cdebris,
            string* clocaldebris, Node* remotenode, fsfp_t cfsfp, bool cinshare, int ctag, void *cappdata)
-    : localroot(new LocalNode)
+: localroot{new LocalNode}
```
mattw-mega: the round brackets that were already here definitely call a constructor - curlies add a lot more possibilities.  Why change something that's perfectly fine, and possibly introduce problems?

cb-mega: Will revert.

include/mega/types.h
cb-mega: It is actually true for two-way. The function we'd use is `Sync::syncDeletions`. The config object should never be used directly when making sync decisions. I'll add some doc explaining that.

PR #1748, opened 2019-11-13
#13849. Fix for detecing statecurrent correctly with spoonfeeding on

mattw-mega:


src/megaclient.cpp
```diff
@@ -4094,7 +4101,11 @@ bool MegaClient::procsc()
                             useralerts.begincatchup = true;
                         }
                     }
-                    app->catchup_result();
+
+                    if (statecurrent)
```
```suggestion
                    if (statecurrent && !insca_notlast)
```
This event should be triggered only if we are actually up to date, so I think it should also check that there's no more pending actionpackets.

mattw-mega: I agree - thanks very much!

src/megaclient.cpp
```diff
@@ -4094,7 +4101,11 @@ bool MegaClient::procsc()
                             useralerts.begincatchup = true;
                         }
                     }
-                    app->catchup_result();
+
+                    if (statecurrent && !insca_notlast)
+                    {
+                        app->catchup_result();
+                    }
```
polmr: If I'm not mistaken this is the same as `if (!insca_notlast)` : if `statecurrent` had evalved `false` and it was the last (i.e. !nsca_notlast == true),  the above if would have set statecurrent as `true`.   Thus, app needs to catchup if the sc is the last one here. We might want to change it for better reading.

sergiohs84: That is true. We can skip to check the `statecurrent`. Please, feel free to remove it and I'll re-approve it.

PR #1752, opened 2019-11-14
WIP: #13845. One-Way-Sync: add unsyncable nodes set and table

cb-mega:


sergiohs84:
This PR must be updated after merging https://github.com/meganz/sdk/pull/1746

src/megaclient.cpp
```diff
@@ -4506,6 +4519,17 @@ void MegaClient::updatesc()
                     {
                         break;
                     }
+#ifdef ENABLE_SYNC
+                    if (!(*it)->isSyncable())
+                    {
+                        mUnsyncableNodes.insert((*it)->nodehandle);
+                        auto nodehandle = std::to_string((*it)->nodehandle);
```
mattw-mega: Can we be sure the dbid has been assigned at this point, since it's for a different table?   I think it would be better to keep them independent, eg. in case of total node tree reload?

src/node.cpp
```diff
@@ -1020,6 +1047,15 @@ bool Node::setparent(Node* p)
     if (parent)
     {
         child_it = parent->children.insert(parent->children.end(), this);
+
+#ifdef ENABLE_SYNC
+        if (parent->type == FILENODE)
+        {
+            // copy syncable from child (old version) to parent (new version)
+            parent->setSyncable(isSyncable());
```
mattw-mega: For a down-only sync, if a file in the cloud had been un-syncable, but then it gets replaced with an updated file, I think we should then sync that file, rather than maintain the old version's status?

src/node.cpp
```diff
@@ -984,6 +984,33 @@ NodeCounter Node::subnodeCounts() const
     return nc;
 }
 
+#ifdef ENABLE_SYNC
+void Node::setSyncable(const bool syncable, const bool notify)
+{
+    if (syncable != mSyncable)
+    {
+        mSyncable = syncable;
+        if (notify)
```
mattw-mega: I think we would want to always keep the node flag and the map and the db table all exactly 'in sync'?  thanks

src/node.cpp
```diff
@@ -1626,6 +1662,11 @@ LocalNode::~LocalNode()
 
     if (node)
     {
+        if (!sync->isUpSync())
+        {
+            node->setSyncable(false);
```
mattw-mega: should we set syncable true here if the sync is being shut down?  (the test for that is a few lines below)  thanks

src/megaclient.cpp
```diff
@@ -4493,6 +4499,13 @@ void MegaClient::updatesc()
                     if ((*it)->dbid)
                     {
                         LOG_verbose << "Removing node from database: " << (Base64::btoa((byte*)&((*it)->nodehandle),MegaClient::NODEHANDLE,base64) ? base64 : "");
+#ifdef ENABLE_SYNC
+                        mUnsyncableNodes.erase((*it)->nodehandle);
+                        if (!(complete = mUnsyncableNodesTable->del((*it)->dbid)))
```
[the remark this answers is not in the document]
cb-mega: Will fix.

src/megaclient.cpp
cb-mega: Will fix.

src/node.cpp
cb-mega: Nah. Current spec is that the file does not get re-downloaded.

src/node.cpp
cb-mega: Will fix.

src/node.cpp
cb-mega: I dont think so. A remote node should retain its syncable property regardless of the current sync, no?

src/node.cpp
mattw-mega: does !mSyncable mean this side can't be synced to the other side, or that the other side can't update this node?  thanks

cb-mega: `mSyncable` on `Node` indicates whether the remote node can be synced to the local tree.

include/mega/sync.h
```diff
@@ -41,6 +44,38 @@ int computeReversePathMatchScore(string& accumulated, const string& path1, const
...
+
+    // Returns the sync path of the given node handle if the node is not syncable, null otherwise
+    const std::string* syncPath(handle nodeHandle) const;
+
+private:
+    uint32_t mNextTableId = 0; // the next table ID to use
+
+    struct NodeData
+    {
+        std::string mSyncPath; // the sync that the node belongs to
```
mattw-mega: Storing a path per unsyncable node could make the data structure quite large in RAM - eg million node account, user manually removes most of the destination nodes.   Could we get away with just testing the presence/absence of the nodehandle in the set, rather than by path?

src/node.cpp
```diff
@@ -984,6 +984,40 @@ NodeCounter Node::subnodeCounts() const
...
+        {
+            LOG_err << "Incomplete database write for node: " << nodehandle;
+        }
+    }
+}
+
+bool Node::isSyncable() const
+{
+    const auto syncPath = client->unsyncables->syncPath(nodehandle);
+    if (syncPath)
```
mattw-mega: wouldn't it be more efficient to test if the nodehandle is in the unordered_map?  thanks

src/node.cpp
```diff
@@ -1020,6 +1054,19 @@ bool Node::setparent(Node* p)
     if (parent)
     {
         child_it = parent->children.insert(parent->children.end(), this);
+
+#ifdef ENABLE_SYNC
+        if (parent->type == FILENODE)
+        {
+            const auto syncPath = client->unsyncables->syncPath(nodehandle);
```
mattw-mega: could we test child->isSyncable here instead?

include/mega/sync.h
cb-mega: Removed syncpath.

src/node.cpp
cb-mega: yes, fixed now.

src/node.cpp
cb-mega: done.

src/megaclient.cpp
```diff
@@ -13451,6 +13482,12 @@ void MegaClient::delsync(Sync* sync, bool deletecache)
         sync->statecachetable->remove();
         delete sync->statecachetable;
         sync->statecachetable = NULL;
+
+        if (!sync->getConfig().isUpSync() && sync->localroot->node)
```
mattw-mega: Is it enough here to do this for down syncs only?  I think a node in a two way sync would never be !syncable?  thanks

src/node.cpp
```diff
@@ -1020,6 +1045,15 @@ bool Node::setparent(Node* p)
...
     {
         child_it = parent->children.insert(parent->children.end(), this);
+
+#ifdef ENABLE_SYNC
+        if (parent->type == FILENODE && !isSyncable())
+        {
+            // if child (old version) is not syncable then parent must follow suit (new version)
+            parent->setSyncable(false);
+            setSyncable(true); // set old version back to default
+        }
```
mattw-mega: This part looks good, but I think we also need to adjust the flags for nodes moved out of a sync, into a sync, or from one sync to another?  thanks

cb-mega: I think it would make sense to maintain the syncable property of a node across syncs. If you removed a local file from a sync, then went to say the webclient and moved the corresponding node to another down-sync, it should probably still not be down-synced. Same if you moved the node back to the original sync folder. Thoughts?

PR #1756, opened 2019-11-14
Adjust share keys test to cope with server side delays, 500s, -3s.

mattw-mega:
Also megacli debug log timestamps in VS output window.


mattw-mega:
trigger compilation

PR #1759, opened 2019-11-18
#13864 Add sync config to megacli

cb-mega:
This adds the syncconfig to megacli so that we can use two-way and one-way syncing from megacli.

### DIFF
--- a/examples/megacli.cpp
+++ b/examples/megacli.cpp
@@ -167,13 +167,20 @@ static std::pair<bool, SyncConfig> syncConfigFromStrings(std::string type, std::
     toLower(overwrite);
 
     SyncConfig::Type syncType;
-    if (type == "up") {
+    if (type == "up")
+    {
         syncType = SyncConfig::TYPE_UP;
-    } else if (type == "down") {
+    }
+    else if (type == "down")
+    {
         syncType = SyncConfig::TYPE_DOWN;
-    } else if (type == "twoway") {
+    }
+    else if (type == "twoway")
+    {
         syncType = SyncConfig::TYPE_DEFAULT;
-    } else {
+    }
+    else
+    {
         return std::make_pair(false, SyncConfig{});
     }
 
@@ -182,19 +189,29 @@ static std::pair<bool, SyncConfig> syncConfigFromStrings(std::string type, std::
 
     if (syncType != SyncConfig::TYPE_DEFAULT)
     {
-        if (syncDel == "on") {
+        if (syncDel == "on")
+        {
             syncDeletions = true;
-        } else if (syncDel == "off") {
+        }
+        else if (syncDel == "off")
+        {
             syncDeletions = false;
-        } else {
+        }
+        else
+        {
             return std::make_pair(false, SyncConfig{});
         }
 
-        if (overwrite == "on") {
+        if (overwrite == "on")
+        {
             forceOverwrite = true;
-        } else if (overwrite == "off") {
+        }
+        else if (overwrite == "off")
+        {
             forceOverwrite = false;
-        } else {
+        }
+        else
+        {
             return std::make_pair(false, SyncConfig{});
         }
     }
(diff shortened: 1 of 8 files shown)


examples/megacli.cpp
```diff
@@ -3907,6 +3942,26 @@ void exec_sync(autocomplete::ACState& s)
...
+    else if (s.words.size() == 4)
+    {
+        syncConfig.syncType = static_cast<SyncConfig::Type>(atoi(s.words[1].s.c_str()));
+        syncConfig.syncDeletions = atoi(s.words[2].s.c_str());
+        syncConfig.forceOverwrite = atoi(s.words[3].s.c_str());
+        cout << "Successfully applied new sync config!" << endl;
+    }
+    else
+    {
+        assert(false);
```
sergiohs84: An assert here is too critical, IMHO. User might mistype the config, ie. add an extra param by mistake, and we can simply remind the user that's not a valid configuration. In other words, it's an error for the user, not for the developer that requires to halt execution.
In any case, it looks like MEGAcli handles that at the autocomplete side, no calling the this method if the provided params doesn't match the configuration of the command.

examples/megacli.cpp
```diff
@@ -3907,6 +3942,26 @@ void exec_sync(autocomplete::ACState& s)
         }
     }
 }
+
+void exec_syncconfig(autocomplete::ACState& s)
+{
+    if (s.words.size() == 1)
+    {
+        cout << "Current sync config: " << syncConfig.syncType <<
```
sergiohs84: We could use the new method `static std::string syncConfigToString(const Sync& sync)` in here?

examples/megacli.cpp
```diff
@@ -3907,6 +3942,26 @@ void exec_sync(autocomplete::ACState& s)
...
+                " " << syncConfig.syncDeletions << " " << syncConfig.forceOverwrite << endl;
+    }
+    else if (s.words.size() == 4)
+    {
+        syncConfig.syncType = static_cast<SyncConfig::Type>(atoi(s.words[1].s.c_str()));
```
sergiohs84: We could make it more intuitive by avoiding the user to know the enumeration values. In example, instead of `0` for up-sync, `UP` (`DOWN` and `FULL`). Fine with 0 or 1 for the flags ;)

examples/megacli.cpp
```diff
@@ -3907,6 +3942,26 @@ void exec_sync(autocomplete::ACState& s)
...
+        syncConfig.syncDeletions = atoi(s.words[2].s.c_str());
+        syncConfig.forceOverwrite = atoi(s.words[3].s.c_str());
+        cout << "Successfully applied new sync config!" << endl;
```
sergiohs84: I'd say that a sanity check here would make sense: if the sync type is full, the user doesn't need to add the flags, since they will not be used at all. In other words, the flags have to be provided only for `up` and `down` types, and for a `full` sync the command could report an error message to avoid setting an invalid configuration.

examples/megacli.cpp
cb-mega: Yea. The argument parsing already takes care of ensuring we end up with the right number of parameters. So if the assert is hit then it must be a programming error, not a user error.

sergiohs84: Probably, we doesn't need the assertion in that case :stuck_out_tongue: 
But alright, let's keep it

examples/megacli.cpp
sergiohs84: Thank you. It is a dev tool, indeed, but that's a small cost that I feel will be paid back after a few months, when another dev than the author wants to try or needs to debug this new command.

examples/megacli.cpp
```diff
@@ -114,6 +114,102 @@ int responseprogress = -1;
...
+    } else if (type == "down") {
+        config.syncType = SyncConfig::TYPE_DOWN;
+    } else if (type == "twoway") {
+        config.syncType = SyncConfig::TYPE_DEFAULT;
+    } else {
+        return std::make_pair(false, config);
+    }
+
+    config.syncDeletions = syncDel == "on";
+    config.forceOverwrite = overwrite == "on";
```
sergiohs84: I tested with `1` instead of `0` without looking at the code, and MEGAcli said:
> Successfully applied new sync config!
Where it is not true. Why not to check the input is correct as you do for the type? It makes me think that perhaps there should be a smarter interface to setup the SyncConfig, whether at the `Sync` object or the `SyncConfig`, in order to ensure its consistency. Maybe I'm too picky, but I think it could be useful later on for your developments.

cb-mega: Should be fixed now.

PR #1760, opened 2019-11-20
#13885. fix: encrypt/decrypt transfers on threads for performance

mattw-mega:
- in addition to performance, this means the mutex a lot freer during transfers, so less blocking of client apps
- encrypt/decrypt operations made independent so they can run on threads (no dangling raw pointers)
- control of the number of worker threads when creating MegaApi instances (0 allowed)
- chunkmac class moved to utils.h (functions were already in utils.cpp)
- function queue class added to execute std::functions on threads
- default to 1 worker thread for these functions, but configurable.  0 threads works too.
- shared_ptr used in a few place so that queued functions can be running when eg. transfer is destroyed.
- two code blocks changed order in transferslot.cpp, so the changes are smaller than they seem at first.
- for raid downloads, the raid block sizes don't match crc chunk sizes so the non-last portions of a chunk are done (first) on the megaclient thread


src/megaclient.cpp
```diff
@@ -34,6 +35,80 @@ namespace mega {
...
+void MegaClientAsyncQueue::clearQueue()
+{
+    std::lock_guard<std::mutex> g(mMutex);
+    mQueue.clear();
+}
+
+void MegaClientAsyncQueue::asyncThreadLoop()
+{
+    SymmCipher cipher;
+    std::unique_lock<std::mutex> g(mMutex);
```
cb-mega: If you move the guard just before the wait then you dont need unlock/lock further down. Something like:
```cpp
for (;;)
{
    std::function<void(SymmCipher&)> f;
    {
        std::unique_lock<std::mutex> g(mMutex);
        mConditionVariable.wait(g, [this]() { return !mQueue.empty(); });
        f = std::move(mQueue.front());
        if (!f) return;
        mQueue.pop_front();
    }
    f(cipher);
    mWaiter.notify();
}
```

src/megaclient.cpp
```diff
@@ -34,6 +35,80 @@ namespace mega {
...
+            break;
+        }
+    }
+    LOG_debug << "MegaClient Worker threads running: " << mAsyncThreads.size();
+}
+
+MegaClientAsyncQueue::~MegaClientAsyncQueue()
+{
+    clearQueue();
+    push(nullptr);
```
cb-mega: Feels quite hacky to me to push a null function to terminate. I'd use a separate boolean for this.

src/megaclient.cpp
```diff
@@ -34,6 +35,80 @@ namespace mega {
 // FIXME: instead of copying nodes, move if the source is in the rubbish to reduce node creation load on the servers
 // FIXME: prevent synced folder from being moved into another synced folder
 
+
+void MegaClientAsyncQueue::push(std::function<void(SymmCipher&)> f)
+{
+    std::lock_guard<std::mutex> g(mMutex);
```
cb-mega: The lock should only protect access to `mQueue`

include/mega/megaclient.h
```diff
@@ -178,6 +181,31 @@ class MEGA_API FetchNodesStats
...
+    void clearQueue();
+
+    MegaClientAsyncQueue(Waiter& w, unsigned threadCount);
+    ~MegaClientAsyncQueue();
+
+private:
+    Waiter& mWaiter;
+    std::mutex mMutex;
+    std::condition_variable mConditionVariable;
+    std::deque<std::function<void(SymmCipher&)>> mQueue;
```
cb-mega: Why isn't `mQueue` a `std::queue`? All you do is push/pop FIFO-style...

include/mega/megaclient.h
```diff
@@ -178,6 +181,31 @@ class MEGA_API FetchNodesStats
     dstime timeToTransfersResumed;
 };
 
+
+// Helper class for MegaClient.  Executes independent tasks on a separate thread or threads.
+// The number of threads can be 0 (eg. for helper MegaApi) in which case something queued is 
+// immediately executed synchronously on the caller's thread
+// Queueing a nullptr instead of a function to execute will cause the threads to exit.
```
cb-mega: Is that really necessary? IMHO the client can just call the destructor of this class to tear down all threads.

include/mega/megaclient.h
```diff
@@ -178,6 +181,31 @@ class MEGA_API FetchNodesStats
     dstime timeToTransfersResumed;
 };
 
+
+// Helper class for MegaClient.  Executes independent tasks on a separate thread or threads.
```
cb-mega: Can the doc mention that this is a thread pool?

include/mega/megaclient.h
```diff
@@ -178,6 +181,31 @@ class MEGA_API FetchNodesStats
...
+
+    MegaClientAsyncQueue(Waiter& w, unsigned threadCount);
+    ~MegaClientAsyncQueue();
+
+private:
+    Waiter& mWaiter;
+    std::mutex mMutex;
+    std::condition_variable mConditionVariable;
+    std::deque<std::function<void(SymmCipher&)>> mQueue;
+    std::vector<std::thread> mAsyncThreads;
```
cb-mega: That name is a bit confusing to me. How does an "async thread" differ from a "thread"?

src/megaclient.cpp
mattw-mega: I like it, nice one!  will update

src/megaclient.cpp
mattw-mega: I did try that, but it bulked the function up a bit, and didn't look as nice.

include/mega/megaclient.h
mattw-mega: I think I chose that because I know it better - actually, I see std::queue is in fact just a wrapper around std::deque (by default - you can choose the container it wraps)

src/megaclient.cpp
mattw-mega: good point, I think that will be fine, adjusted

include/mega/megaclient.h
mattw-mega: It's an implementation detail so no need to mention it here really - I've moved that comment to the relevant function

include/mega/megaclient.h
mattw-mega: done

include/mega/megaclient.h
mattw-mega: Threads for async operations.  Named in consultation with the department of redundancies department ;-).  updated

src/transferslot.cpp
```diff
@@ -758,12 +769,19 @@ void TransferSlot::doio(MegaClient* client, DBTableTransactionCommitter& committ
...
+                                req->status = REQ_ENCRYPTING;
 
-                                reqs[i]->pos = ChunkedHash::chunkfloor(asyncIO[i]->pos);
-                                reqs[i]->status = REQ_PREPARED;
+                                client->mAsyncQueue.push([req, transferkey, ctriv, finaltempurl, pos, npos](SymmCipher& sc)
+                                    {
+                                        sc.setkey(transferkey.data());
+                                        req->prepare(finaltempurl.c_str(), &sc, ctriv, pos, npos); 
+                                        req->status = REQ_PREPARED;
+                                    });
```
cb-mega: Can we guarantee that `req` is not accessed from another thread at the same time? Is it possible that `doio` runs again at the same time as `prepare` runs on the worker thread?

cb-mega: I wonder if we need to serialize access to members of `req` to avoid race conditions (now and in the future).

mattw-mega: excellent question - doio() can run while prepare does, but the req status has been put to `req->status = REQ_ENCRYPTING` and doio() won't do anything else with that req until its status is put to `REQ_PREPARED` by the lambda.

mattw-mega:
trigger compilation

include/mega/megaclient.h
```diff
@@ -180,6 +183,31 @@ class MEGA_API FetchNodesStats
     dstime timeToTransfersResumed;
 };
 
+
+// Helper class for MegaClient.  
+// Maintains a small thread pool for executing independent operations such as encrypt/decrypt a block of data
+// The number of threads can be 0 (eg. for helper MegaApi that deals with public folder links) in which case something queued is 
+// immediately executed synchronously on the caller's thread
+struct MegaClientAsyncQueue
```
sergiohs84, line 191: I think this class will suit better in `utils.h`/`utils.cpp` rather than in MegaClient. It could be reused, in example, for the upcoming nodes on demand to load nodes from DB asynchronously.

src/raid.cpp
```diff
@@ -292,13 +288,13 @@ void RaidBufferManager::submitBuffer(unsigned connectionNum, FilePiece* piece)
...
-        asyncoutputbuffers[connectionNum] = piece;
+        asyncoutputbuffers[connectionNum].reset(piece);
     }
 }
 
-RaidBufferManager::FilePiece* RaidBufferManager::getAsyncOutputBufferPointer(unsigned connectionNum)
+std::shared_ptr<RaidBufferManager::FilePiece> RaidBufferManager::getAsyncOutputBufferPointer(unsigned connectionNum)
 {
-    std::map<unsigned, FilePiece*>::iterator i = asyncoutputbuffers.find(connectionNum);
+    auto i = asyncoutputbuffers.find(connectionNum);
```
sergiohs84, line 297: No need to add a commit for this `auto`s, but in cases like this one it would be better to keep the full declaration, so the reader knows at first sight what's in the first/second

src/transferslot.cpp
sergiohs84: I was about to do a similar question about concurrent access to `HttpReqXfer` objects. Good to get the answer here already :smile:

mattw-mega:
Thanks very much - I've merged and am retesting and remeasuring now.   I'll make the review adjustments on another PR associated with the same ticket, as we can surely get those done very quickly.  thanks!

include/mega/megaclient.h
mattw-mega: No problem moving it, will do.  I had planned for it to be generic like that initially, until realizing that it needed the SymmCipher as a member for this case, purely because those are very expensive to construct, but very cheap to reuse.   I'm sure there is a way to make it reusable, but I didn't want to further complicate things when it's only used for this case at this stage - anyway, will move it.  cheers

src/utils.cpp
```diff
@@ -317,15 +317,13 @@ m_off_t chunkmac_map::expandUnprocessedPiece(m_off_t pos, m_off_t npos, m_off_t
     return npos;
 }
 
-void chunkmac_map::finishedUploadChunks(m_off_t pos, m_off_t size)
+void chunkmac_map::finishedUploadChunks(m_off_t pos, m_off_t size, chunkmac_map& macs)
```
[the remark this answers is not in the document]
mattw-mega: done, good spotting, thanks

src/raid.cpp
mattw-mega: yes I've been on both sides of this, having both enjoyed the convenience of auto when writing code, and having thought it would be nice to know the exact types when reviewing.    This one isn't too long but some iterator names are ridiculous and auto is certainly better then.   In this case we can see that `find` is looking for a `connectionNum` so we know what the key of the map is (which corresponds to `first` in the pair of course.    And then the name of the map tells us what the value (ie `second`) is in this case - an output buffer (though without the exact type name of course).   In a good GUI you can just mouse over it to get the exact types of course.  Anyway, happy to keep it in mind and good to discuss.  cheers

PR #1767, opened 2019-11-28
WIP #13858. One-Way-Sync: make one-way sync decisions

cb-mega:
This aims at using the sync config to make one-way sync decisions


src/node.cpp
```diff
@@ -202,7 +202,7 @@ Node::~Node()
     // sync: remove reference from local filesystem node
     if (localnode)
     {
-        localnode->deleted = true;
+        localnode->reactToNodeChange(true);
```
mattw-mega: Is it safe to call this function from the end of the Node destructor?  If the localnode tries to access any info in the Node, it may well be referring to deleted (but non-NULL) pointers.

src/node.cpp
```diff
@@ -1343,6 +1428,18 @@ void LocalNode::setnameparent(LocalNode* newparent, string* newlocalpath)
     }
 }
 
+void LocalNode::reactToNodeChange(const bool nodeDeleted)
+{
+    if (!sync->getConfig().isDownSync())
```
mattw-mega: I suggest renaming isDownSync() to syncsDown() or similar.  isDownSync sounds like it's the one-way sync in that direction (excluding two-way), however it tests whether the sync moves changes in that direction (so true for one-way down sync and for two-way sync)

src/node.cpp
cb-mega: I think it's fine. `reactToNodeChange` only access data in localnode and sync.

src/node.cpp
cb-mega: syncsDown is problematic cuz it sounds like "the sync is down" in like not working. We also have `syncsup` of `MegaClient` so using `syncsUp` is problematic too.

cb-mega: Open to other name suggestions tho

src/node.cpp
mattw-mega: can we set localnode->node = NULL first?  So that anyone who tries that in future will find out quickly.  It's the next line anyway

src/node.cpp
mattw-mega: how about doesSyncUp ?  syncsToCloud / syncsToLocal  maybe

cb-mega: The latter sounds. Will change to that.

src/node.cpp
cb-mega: Sure

PR #1777, opened 2019-12-17
#13534. Updated win build script for x86/x64 debug/release pdfium builds

mattw-mega:
3rdParty library pdfium is built statically, direct with a cmake script, as the depot_tools system would not produce working DLLs


bindings/qt/sdk.pri
```diff
@@ -323,7 +368,7 @@ win32 {
     LIBS += -lwinhttp -ladvapi32
     DEFINES += _CRT_SECURE_NO_WARNINGS
 }
-
+else:CONFIG += USE_CURL
```
polmr: why this one? It does not seem to be used afterwards for !win32 anyway ....

bindings/qt/sdk.pri
```diff
@@ -572,32 +648,33 @@ macx {
...
+   !vcpkg:CONFIG(USE_OPENSSL) {
     INCLUDEPATH += $$MEGASDK_BASE_PATH/bindings/qt/3rdparty/include/openssl
     LIBS += $$MEGASDK_BASE_PATH/bindings/qt/3rdparty/libs/libssl.a $$MEGASDK_BASE_PATH/bindings/qt/3rdparty/libs/libcrypto.a
    }
 
-
-
    LIBS += -framework SystemConfiguration
+   
+   LIBS += -liconv -framework CoreServices -framework CoreFoundation -framework AudioUnit -framework AudioToolbox -framework CoreAudio -framework CoreMedia -framework VideoToolbox -framework ImageIO -framework CoreVideo
```
polmr: do we need all these for !vcpkg?

contrib/cmake/build3rdparty.bash
```diff
@@ -1,17 +1,88 @@
...
+# 
+# Copy the folders in contrib/cmake/vcpkg_extra_ports to your 3rdParty/vcpkg/ports folder.
+# Copy the folders in contrib/cmake/vcpkg_extra_triplets to your 3rdParty/vcpkg/triplets folder.
+# 
+# Comment out any libraries that you won't use.
+# If using pdfium, follow the instructions below to get the source code
+#
+# On Mac, go to ports/libraw/CONTROL and comment out the freeglut dependency
+# 
+# Copy this script to your 3rdParty/vcpkg folder, and run it with the desired triplet as the parameter.  (usually x64-windows-mega or x86-windows-mega)
```
polmr: coyping it to the vcpkg folder is not required. x64-windows... does not make sense in this bash script, doesn't it?
We can have the ports & triplet linked from the location of this script (build3rpardy.bash) automatically. Sth like:
```
DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
ln -sf $DIR/vcpkg_extra_ports/* ports/
ln -sf $DIR/vcpkg_extra_triplets/* triplets/
```

bindings/qt/sdk.pri
```diff
@@ -176,16 +204,21 @@ CONFIG(USE_MEDIAINFO) {
 CONFIG(USE_LIBRAW) {
     DEFINES += HAVE_LIBRAW
 
+    vcpkg:LIBS += -lraw$$DEBUG_SUFFIX -ljasper$$DEBUG_SUFFIX
+    vcpkg:win32:LIBS += -ljpeg$$DEBUG_SUFFIX
+    vcpkg:!win32:LIBS += -ljpeg
+    vcpkg:unix:!macx:LIBS += -lgomp
```
polmr: without pdfium this requires `-llcms2$$DEBUG_SUFFIX`

bindings/qt/sdk.pri
mattw-mega: Yes I added this one because for vcpkg on non-win32, there was a build error (I forget if it was include or link), and the missing part was in one of the target-specific blocks.   Turning on USE_CURL means that the block on line 514 (inside vcpkg and protected by USE_CURL) adds include paths and libs for curl, openssl, and c-ares.   I'm not sure if this is the best balance, and taking into consideration mixing vcpkg and locally installed libraries for linux might mean we need some changes for that?  thanks

bindings/qt/sdk.pri
mattw-mega: no, those are vcpkg specific.  I think they may be dependencies of some of the 3rd party libraries though I didn't figure out which ones.  I'll add `vcpkg:` so that the existing prebuilt 3rdParty system is not affected.

contrib/cmake/build3rdparty.bash
```diff
@@ -1,17 +1,88 @@
...
-#./vcpkg install --triplet x64-linux libsodium
+# This script is for getting and building the 3rd party libraries that the MEGA SDK uses (some are optional, and some are only needed only by MEGA apps too)
+# 
+# Your 3rdParty library builds should be outside the SDK repo.  We are moving to use vcpkg to build most of them. You can start it like this:
+#
+# mkdir 3rdParty
+# cd 3rdParty
+# git clone https://github.com/Microsoft/vcpkg.git
+# cd vcpkg
+# .\bootstrap-vcpkg.bat
```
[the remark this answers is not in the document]
mattw-mega: good point, adjusted

contrib/cmake/build3rdparty.bash
mattw-mega: That's a very interesting idea - it would be great for the script to automate as much as possible.  It looks like this assumes that the script would stay in the contrib/cmake folder, and be invoked from the 3rdparty/vcpkg folder.   I've been copying it to the target folder and running from there - but I think that would be fine.  I'll make the update.

bindings/qt/sdk.pri
mattw-mega: thanks, added

contrib/cmake/build3rdparty.bash
mattw-mega: Also we should think about adding an x64-linux-mega triplet file to the extras folder, which could select which libraries will be built for static linking in a standard build, similar to the x64-windows-mega triplet file.

contrib/cmake/build3rdparty.bash
```diff
@@ -1,17 +1,80 @@
...
+#
+# Comment out any libraries that you won't use.
+# If using pdfium, follow the instructions below to get the source code
+#
+# On Mac, go to ports/libraw/CONTROL and comment out the freeglut dependency
+# 
+# From your 3rdParty/vcpkg folder, run this script (in its proper location) with the desired triplet as the parameter.  (usually x64-linux or x64-osx)
 
+export TRIPLET=$1
+
```
vtmateos: echo "" > buildlog

contrib/cmake/build3rdparty.bash
```diff
@@ -1,17 +1,80 @@
...
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+ln -sf $DIR/vcpkg_extra_ports/* ports/
+ln -sf $DIR/vcpkg_extra_triplets/* triplets/
```
vtmateos: It seems that `ln` with wildcard does not work on macOS. Tested with two for loops to create the needed links works.

mattw-mega: Ok, we may need to list those out manually then - those lines are an improvement by @polmr over the prior version by me that just asked the user to copy them manually :-).   There are not many, we can just list them and maintain the list.

contrib/cmake/build3rdparty.bash
mattw-mega: will add

contrib/cmake/build3rdparty.bash
mattw-mega: actually, on consideration, the loop does sound good - could you send me that (or just add directly) - that way it's tested code going into the script?  thanks

contrib/cmake/build3rdparty.bash
mattw-mega: is this to clean prior build records from the buildlog?  Those can be handy to see, eg if you have to run the script several times after adjusting for things that are not working, eg. apt-get install various things (already completed vckpg packages will be skipped by vcpkg).   How about we add a prominent line saying "started build" and the date/time? thanks

vtmateos: Yes, the purpose is to clean prior build records.. I am ok with an info line, up to you. No problem at all.

src/gfx/qt.cpp
```diff
@@ -378,7 +378,8 @@ int GfxProcQT::processEXIF(QByteArray *data, int itemlen){
 }
 /************* END OF EXIF STUFF **************/
 
-#if defined(HAVE_PDFIUM) && defined(_WIN32)
+// turn on PDFIUM_DELAY_LOAD_DLL if you are building XP-compatible apps and using pdfium.dll (which won't load on XP)
+#if PDFIUM_DELAY_LOAD_DLL
```
polmr: Shouldn't we have this defined in `sdk.pri` for the moment as default when win32 & USE_PDFIUM ?? and only remove it after ceasing XP support?

mattw-mega: using the vcpkg/cmake system, we can build pdfium statically, and I think that will work on XP (I have not tried running it, but compiling and linking for XP works).  In case I'm wrong, we can easily turn the delay load back on.

PR #1781, opened 2019-12-19
Remove assert from syncupdate

cb-mega:
This also adds getter and setter for `File::h` to improve maintainability (this helped me debug).


sergiohs84:
The assert is correct, so no need to remove it but to adjust the code to meet the condition in the assert. That is addressed in https://github.com/meganz/sdk/pull/1783 --> closing the PR.

PR #1782, opened 2019-12-19
#13931 iOS transfers change priority

lhr000lhrmega:
provide the ability to change transfer priority

### DIFF
--- a/bindings/ios/MEGATransfer.mm
+++ b/bindings/ios/MEGATransfer.mm
@@ -164,7 +164,7 @@ using namespace mega;
 }
 
 - (unsigned long long)priority {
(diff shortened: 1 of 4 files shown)


bindings/ios/MEGATransfer.mm
```diff
@@ -163,4 +163,8 @@ - (MEGATransferState)state {
     return (MEGATransferState) (self.megaTransfer ? self.megaTransfer->getState() : 0);
 }
 
+- (unsigned long long)priority {
+    return (MEGATransferState) (self.megaTransfer ? self.megaTransfer->getPriority() : 0);
```
jnavarrom: Looks like this casting to MEGATransferState is wrong.

lhr000lhrmega: fixed

PR #1792, opened 2019-12-30
#14159. Auto-resume active syncs after fetchnodes

cb-mega:


src/sync.cpp
```diff
@@ -511,10 +511,127 @@ bool assignFilesystemIds(Sync& sync, MegaApp& app, FileSystemAccess& fsaccess, h
...
+            {
+                assert(false);
+                LOG_err << "Incomplete database del at id: " << syncConfigPair->second.mTableId;
+            }
+        }
+        mSyncConfigs.erase(syncConfig.getLocalPath());
+    }
+    else
+    {
+        assert(false && "sync config does not exist");
```
sergiohs84: Perhaps log an error message is also a good idea for debugging purposes (based on logs, rather on a debugger :) )

src/sync.cpp
```diff
@@ -511,10 +511,127 @@ bool assignFilesystemIds(Sync& sync, MegaApp& app, FileSystemAccess& fsaccess, h
...
+            if (!mTable->put(tableId, &data))
+            {
+                assert(false);
+                LOG_err << "Incomplete database put at id: " << tableId;
+            }
+        }
+    }
+    else
+    {
+        assert(false && "sync config does not exist");
```
sergiohs84: Perhaps log an error message is also a good idea for debugging purposes (based on logs, rather on a debugger :) )

src/utils.cpp
```diff
@@ -2002,21 +2019,166 @@ void NodeCounter::operator -= (const NodeCounter& o)
...
 
+bool SyncConfig::isUpSync() const
+{
+    return mSyncType & TYPE_UP;
+}
 
+bool SyncConfig::isDownSync() const
+{
+    return mSyncType & TYPE_DOWN;
+}
```
sergiohs84: A `bool SyncConfig::isTwoWaySync() const` could be useful too.

src/utils.cpp
```diff
@@ -2002,21 +2019,166 @@ void NodeCounter::operator -= (const NodeCounter& o)
...
 
+std::unique_ptr<SyncConfig> SyncConfig::unserialize(const std::string& data)
+{
+    bool active;
+    std::string localPath;
+    handle remoteNode;
+    fsfp_t fingerprint;
+    uint32_t regExpCount;
+    std::vector<std::string> regExps;
+    int64_t syncType;
```
sergiohs84: I hesitate we need an integer of 64 bits for a simple 3-value variable, at least in DB. Not a big deal, since the number of syncs is small, but perhaps a shorter value will save a bit of space.

src/utils.cpp
```diff
@@ -2002,21 +2019,166 @@ void NodeCounter::operator -= (const NodeCounter& o)
...
+    }
+    if (!reader.unserializebool(forceOverwrite))
+    {
+        return {};
+    }
 
+    auto syncConfig = std::unique_ptr<SyncConfig>{new SyncConfig{std::move(localPath),
+                    remoteNode, fingerprint, std::move(regExps),
+                    static_cast<Type>(syncType), syncDeletions, forceOverwrite}};
+    syncConfig->setActive(active);
```
sergiohs84: The `active` property of a `SyncConfig` should be persisted, shouldn't it? We don't want to remove the `SyncConfig` just because the user has disabled the sync temporary.

include/mega/types.h
```diff
@@ -680,58 +680,71 @@ class SyncConfig
...
-            case TYPE_DOWN: return mForceOverwrite;
-            case TYPE_DEFAULT: return false;
-        }
-        assert(false);
-        return false;
-    }
+    bool forceOverwrite() const;
+
+    // serializes the object to a string
+    std::string serialize() const;
```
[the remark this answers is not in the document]
sergiohs84: You can try (in a separate branch from `develop`) to adjust `serialize` to be a const method, as well as receive a `const string &data` if you like. But I wouldn't change it if it takes more than 10 minutes and specially if it may introduce any kind of error ;)

src/sync.cpp
sergiohs84: Wouldn't it be better to adjust the logout procedure to not reach this point during logout?? Same for `update()`. If you added an `assert` in that place it is because you do expect to find the sync-config when the function is called. If it does not exist, it would represent a error. 
In conclusion, I suggest whether the logout is changed to avoid this failed assertion, or at least to log an error message, so a developer can notice the sync-config was not found.

src/utils.cpp
sergiohs84: Oh, my bad. Sorry!

src/sync.cpp
cb-mega: What I meant is that it's now a possible and expected program flow to erase a sync config that doesn't exist anymore. When you logout the sync config bag is cleared first and then all sync objects are destroyed which in turn attempts to remove their config from the bag (but it's already gone at this point).

include/mega/types.h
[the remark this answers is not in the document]
cb-mega: I'll put it on my todo list for later :)

cb-mega:
trigger compilation

cb-mega:
trigger compilation

cb-mega:
trigger compilation

cb-mega:
trigger compilation

cb-mega:
trigger compilation

cb-mega:
trigger compilation

cb-mega:
trigger compilation

cb-mega:
trigger compilation

cb-mega:
trigger compilation

cb-mega:
trigger compilation

cb-mega:
trigger compilation

cb-mega:
trigger compilation

cb-mega:
trigger compilation

cb-mega:
trigger compilation

cb-mega:
trigger compilation

cb-mega:
trigger compilation

cb-mega:
trigger compilation

cb-mega:
trigger compilation

cb-mega:
trigger compilation

cb-mega:
trigger compilation

cb-mega:
trigger compilation

src/megaapi_impl.cpp
```diff
@@ -7359,6 +7378,105 @@ bool MegaApiImpl::hasToForceUpload(const Node &node, const MegaTransferPrivate &
     return canForceUpload && (isMedia || isPdf) && !(hasPreview && hasThumbnail);
 }
 
+#ifdef ENABLE_SYNC
+void MegaApiImpl::resumeSyncImpl(const char *localFolder, const long long localfp, const handle nodeHandle, MegaRegExp *regExp, MegaRequestListener *listener)
```
cb-mega: This function contains the same body as the original `MegaApiImpl::resumeSync` function.

src/utils.cpp
```diff
@@ -1996,21 +2013,182 @@ void NodeCounter::operator -= (const NodeCounter& o)
...
 
-} // namespace
+SyncConfig::SyncConfig(std::string localPath,
+                       const handle remoteNode,
+                       const fsfp_t localFingerprint,
+                       std::vector<std::string> regExps,
+                       const Type syncType,
+                       const bool syncDeletions,
+                       const bool forceOverwrite)
+    : mLocalPath{std::move(localPath)}
```
sergiohs84: According to the Code style:
_Add a trailing comma to the last element in e.g. enums or maps for more practical diffs and merges._
Usually we put the comma at the end of the line, something like:
```suggestion
    : mLocalPath{std::move(localPath)},
```

sergiohs84: Nevermind. For initializer list, in case of multiple members, the only option is to add the comma at the beginning of the line, since the compiler will complain if there is an trailing comma.

src/commands.cpp
```diff
@@ -1943,6 +1943,10 @@ void CommandLogin::procresult()
                     client->sessionkey.assign((const char *)sek, sizeof(sek));
                 }
 
+#ifdef ENABLE_SYNC
+                client->resetSyncConfigs();
```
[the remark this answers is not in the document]
sergiohs84: Oh, I see. It considers the possibility where a cache already exists for the logged-in user, despite the session has been created right now. Lets discuss first if the cache should survive logouts.

src/megaapi_impl.cpp
```diff
@@ -7359,6 +7378,105 @@ bool MegaApiImpl::hasToForceUpload(const Node &node, const MegaTransferPrivate &
...
+    error e = API_OK;
+    fireOnRequestStart(request);
+
+    const char *localPath = request->getFile();
+    Node *node = client->nodebyhandle(request->getNodeHandle());
+    if (!node)
+    {
+        char base64[12];
+        auto nodeHandle = request->getNodeHandle();
+        LOG_warn << "Node invalid for handle: " << (Base64::btoa((byte*)&nodeHandle, MegaClient::NODEHANDLE, base64) ? base64 : "");
```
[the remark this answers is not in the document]
sergiohs84: Please, resolve the conversation once it's changed. Thanks

src/megaclient.cpp
```diff
@@ -1239,6 +1241,19 @@ MegaClient::~MegaClient()
...
 }
 
+#ifdef ENABLE_SYNC
+void MegaClient::resetSyncConfigs()
+{
+    syncConfigs.reset();
+    if (dbaccess && me != UNDEF)
+    {
+        char buf[16];
+        Base64::btoa((const byte*)&me,  MegaClient::USERHANDLE, buf);
```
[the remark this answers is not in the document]
sergiohs84: Please, resolve the conversation once it's changed. Thanks

src/megaclient.cpp
```diff
@@ -12020,41 +12039,44 @@ error MegaClient::addtimer(TimerWithBackoff *twb)
 // check sync path, add sync if folder
 // disallow nested syncs (there is only one LocalNode pointer per node)
 // (FIXME: perform the same check for local paths!)
-error MegaClient::addsync(const SyncConfig syncConfig, string* rootpath, const char* debris, string* localdebris, Node* remotenode, fsfp_t fsfp, int tag, void *appData)
+error MegaClient::addsync(SyncConfig syncConfig, const char* debris, string* localdebris, int tag, void *appData)
 {
 #ifdef ENABLE_SYNC
+    auto remotenode = nodebyhandle(syncConfig.getRemoteNode());
```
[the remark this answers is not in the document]
sergiohs84: Please, resolve the conversation once it's changed. Thanks

src/megaclient.cpp
```diff
@@ -1239,6 +1241,19 @@ MegaClient::~MegaClient()
...
 
+#ifdef ENABLE_SYNC
+void MegaClient::resetSyncConfigs()
+{
+    syncConfigs.reset();
+    if (dbaccess && me != UNDEF)
+    {
+        char buf[16];
+        Base64::btoa((const byte*)&me,  MegaClient::USERHANDLE, buf);
+        syncConfigs.reset(new SyncConfigBag{*dbaccess, *fsaccess, rng, buf});
```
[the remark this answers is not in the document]
sergiohs84: TBD. You are right, if we want to allow resumption upon a fresh-login of same user, we need to name the DB based on the `me`, not based on the `sid`. Let's first discuss about it ;)

src/megaclient.cpp
```diff
@@ -12020,41 +12039,44 @@ error MegaClient::addtimer(TimerWithBackoff *twb)
...
     bool inshare = false;
     error e = isnodesyncable(remotenode, &inshare);
     if (e)
     {
         return e;
     }
 
-    if (rootpath->size() >= fsaccess->localseparator.size()
-     && !memcmp(rootpath->data() + (int(rootpath->size()) & -int(fsaccess->localseparator.size())) - fsaccess->localseparator.size(),
+    auto localPath = syncConfig.getLocalPath();
```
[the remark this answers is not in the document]
sergiohs84: OK. Please, resolve the conversation once it's changed. Thanks

src/sync.cpp
```diff
@@ -511,12 +511,149 @@ bool assignFilesystemIds(Sync& sync, MegaApp& app, FileSystemAccess& fsaccess, h
...
+            ++mTable->nextid;
+        }
+    }
+    else
+    {
+        const auto tableId = syncConfigPair->second.dbid;
+        if (mTable)
+        {
+            DBTableTransactionCommitter committer{mTable.get()};
+            if (!mTable->del(tableId))
```
sergiohs84: I think you don't need to delete+put here. The `put()` already allows to `insert or update` SQL query :)

My suggestion was to somehow prevent additions that actually update an existing config. The caller should know whether the config is a new one (add) or an update of an existing one (update). Therefore, I was thinking about a `SyncConfigBag::add()` that fails if it already exists, plus a `SyncConfigBag::update()` that fails if the config doesn't exist yet. If you feel it's counterproductive, feel free to go ahead with the new `insert()`.

src/megaclient.cpp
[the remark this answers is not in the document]
cb-mega: I was under the impression that the reviewer should resolved after verifying that the changes actually took place?!

src/sync.cpp
cb-mega: Indeed, put does both! I simplified the insert method. 

I think if the caller really does need to know whether it exists or not then the caller can check using the new `get` method. However, I don't really think it's a problem. Would like to keep the insert method (which makes the class similar to a `set`).

src/megaclient.cpp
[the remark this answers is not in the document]
sergiohs84: Since the sync-config cache will survive logouts, we need the DB's name from `me` and not from `sid`. Solved!

src/commands.cpp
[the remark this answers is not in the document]
sergiohs84: Since the sync-config cache will survive logouts, the cache may already exist upon session creation. Let's keep it the way it is :)

src/sync.cpp
sergiohs84: Well, actually a `std::set::insert()` will behave more like my suggestion, it will fail to insert if there is already an existing item in the set. But alright, let's keep it as it is. It's not a biggie, as long as the caller doesn't overwrite an existing `SyncConfig` in the bag by mistake (we usually opt by protect developers from coding mistakes, when possible).

src/sync.cpp
```diff
@@ -548,19 +548,28 @@ SyncConfigBag::SyncConfigBag(DbAccess& dbaccess, FileSystemAccess& fsaccess, Prn
...
+            assert(false);
+            mTable->abort();
+            return false;
+        }
+        return true;
+    };
+
     auto syncConfigPair = mSyncConfigs.find(syncConfig.getLocalPath());
-    if (syncConfigPair == mSyncConfigs.end())
+    if (syncConfigPair == mSyncConfigs.end()) // syncConfig is new
```
sergiohs84: `syncConfigPair` is an iterator. It should be named `syncConfigIt` or alike, but not `...Pair`, which is misleading. I'll change it.

include/megaapi_impl.h
```diff
@@ -2225,8 +2225,7 @@ class MegaApiImpl : public MegaApp
         //Sync
         int syncPathState(string *path);
         MegaNode *getSyncedNode(string *path);
-        void syncFolder(const char *localFolder, MegaNode *megaFolder, MegaRegExp *regExp = NULL, MegaRequestListener* listener = NULL);
-        void resumeSync(const char *localFolder, long long localfp, MegaNode *megaFolder, MegaRegExp *regExp = NULL, MegaRequestListener *listener = NULL);
+        void syncFolder(const char *localFolder, MegaNode *megaFolder, MegaRegExp *regExp = NULL, MegaRequestListener* listener = NULL, long long localfp = 0);
```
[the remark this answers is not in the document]
cb-mega: done

include/megaapi.h
```diff
@@ -9826,8 +9826,9 @@ class MegaApi
          * - MegaRequest::getNodeHandle - Returns the public handle if logged into a public folder. Otherwise, INVALID_HANDLE
          *
          * @param listener MegaRequestListener to track this request
+         * @param resumeSyncs Whether to auto-resume syncs following a successful fetch
          */
-        void fetchNodes(MegaRequestListener *listener = NULL);
+        void fetchNodes(MegaRequestListener *listener = NULL, bool resumeSyncs = false);
```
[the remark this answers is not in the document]
cb-mega: Added a new function

sergiohs84:
Ready to merge. Next step: enable automatic sync resumption for MEGAcmd :muscle:

PR #1796, opened 2019-12-31
#14155. Fix a memory leak in fetchtimezone_result

mattw-mega:


src/megaapi_impl.cpp
```diff
@@ -15084,11 +15084,13 @@ void MegaApiImpl::fetchtimezone_result(error e, vector<std::string> *timezones,
 
     if (requestMap.find(client->restag) == requestMap.end())
     {
+        delete tzDetails;
         return;
     }
     MegaRequestPrivate* request = requestMap.at(client->restag);
     if (!request || ((request->getType() != MegaRequest::TYPE_FETCH_TIMEZONE)))
     {
+        delete tzDetails;
```
cb-mega: unique_ptr would be ideal here

mattw-mega: I think the whole function could be rearranged to be simpler, but making the change like this is super clear and there can't be accidental side effects

cb-mega: No need to rearrange the function. Just do this:
```cpp
    std::unique_ptr<MegaTimeZoneDetails> tzDetails;
    if (!e)
    {
        tzDetails.reset(new MegaTimeZoneDetailsPrivate(timezones, timezoneoffsets, defaulttz));

        // update the cached timezones for notifications filtering
        delete mTimezones;
        mTimezones = tzDetails->copy();
    }

    if (requestMap.find(client->restag) == requestMap.end())
    {
        return;
    }
    if (!request || ((request->getType() != MegaRequest::TYPE_FETCH_TIMEZONE)))
    {
        return;
    }

    request->setTimeZoneDetails(tzDetails.get());
    fireOnRequestFinish(request, MegaError(e));

```
which is also very simple and sets another good example on using unique_ptr. Explicit `delete` is a code smell we should try to avoid.

mattw-mega: all right; updated

PR #1798, opened 2019-12-31
#14154. Initialize and cleanup c-url and c-ares just once

mattw-mega:
Even when there are multiple MegaApi, hence multiple CurlHttpIO.
Also taking startup on multiple threads into account.

### DIFF
--- a/src/posix/net.cpp
+++ b/src/posix/net.cpp
@@ -275,9 +275,11 @@ CurlHttpIO::CurlHttpIO()
 
 #endif
 
-    curl_global_init(CURL_GLOBAL_DEFAULT);
-    ares_library_init(ARES_LIB_INIT_ALL);
-
+    if (++instanceCount == 1)
+    {
+        curl_global_init(CURL_GLOBAL_DEFAULT);
+        ares_library_init(ARES_LIB_INIT_ALL);
+    }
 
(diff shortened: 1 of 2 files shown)


src/posix/net.cpp
```diff
@@ -718,14 +720,19 @@ CurlHttpIO::~CurlHttpIO()
     closecurlevents(PUT);
 
     curlMutex.lock();
```
cb-mega: We shouldn't need this anymore

cb-mega: Actually, we do to ensure that nothing else runs while we shutdown. But then why do we need the int to be atomic?

mattw-mega: I made the int atomic since threading is relevant for it - probably it's not strictly needed since its use is inside the mutex, but it won't hurt and it's only used on startup/shutdown.  Perhaps we could get away without locking the mutex on exit (in which case we do need the atomic), but I think it's nicer for the startup and shutdown to be symmetric.  I think you'd agree it will work?  thanks

cb-mega: Yes it works but it's confusing to the reader that you lock a mutex and also use an atomic for no apparent reason.

mattw-mega: ok, atomic<> removed

PR #1805, opened 2020-01-07
Cacheable: Fix type of dbid; Use override for serialize; Fix spelling

cb-mega:


include/mega/types.h
```diff
@@ -248,16 +248,15 @@ const int FOLDERNODEKEYLENGTH = 16;
 typedef list<class Sync*> sync_list;
 
 // persistent resource cache storage
-struct Cachable
+class Cacheable
```
mattw-mega: all members are public so struct is more natural than class - why change it?  Save a keyword ;-)

include/mega/types.h
```diff
@@ -248,16 +248,15 @@ const int FOLDERNODEKEYLENGTH = 16;
...
 {
-    virtual bool serialize(string*) = 0;
-
-    int32_t dbid;
+public:
+    virtual ~Cacheable() = default;
```
mattw-mega: the destructor is an implementation detail, why not leave it at the bottom?  serialize() is the most important member

include/mega/types.h
cb-mega: A struct shouldn't have virtual functions IMO.

include/mega/types.h
cb-mega: The destructor is a very important impl. detail for base classes. You need to know if it's virtual or not to know how you can use the base class.

PR #1815, opened 2020-01-14
#14228. Send the fetchnodes "f" command in a separate batch

mattw-mega:
The servers can process that more efficiently
Also, fixed std::tolower call in megacli which expects two parameters - just call ::tolower.

### DIFF
--- a/examples/megacli.cpp
+++ b/examples/megacli.cpp
@@ -158,7 +158,7 @@ static std::pair<bool, SyncConfig> syncConfigFromStrings(std::string type, std::
 {
     auto toLower = [](std::string& s)
     {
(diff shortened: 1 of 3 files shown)


mattw-mega:
trigger compilation

mattw-mega:
Hi @sergiohs84 I've moved those two requests prior to the fetchnodes, which is a bit better for the fetchnods-from-server case, but the fetchnodes-from-cache bypasses the request queue (and happens on login_result), and so those will complete after the fetchnodes_result callback in that case - which is the same as it was before.  I agree we should work on the `ug` ticket to further tidy these cases up, but I think this ticket is ok to merge now if you approve?  thanks

PR #1818, opened 2020-01-15
#14214. Use IOS patch in jsep.h on Linux

alber2510:


sergiohs84:
@alber2510 please, rebuild the SDK for MEGAchat and MEGAchat itself with the last commit in Fedora. It solves the compilation issue for Ubuntu.

alber2510:
It works perfect in Fedora. From my point of view the PR is approved

PR #1820, opened 2020-01-16
#14238. Fire account update upon receiving 'ub' action packet

polmr:
when transition from a non business account


polmr:
Close in favor of release branch: https://github.com/meganz/sdk/pull/1831

PR #1822, opened 2020-01-17
Modified two flags in WebRTC build script to match Android

javitrujillo:
This reduces the final size of the library in a 25%


javisanesp:
I tested this branch and executed the script twice and the `libwebrtc.a` library went from 1,26 GB to 1,2 GB. I do not know if I am missing something. 

Anyway, we do not need that flags, so carry on 🙂

jcmartinac:
I have tested this with same result as Santana. I expected a higher slim down but as the flags are not needed we can go with it.

PR #1823, opened 2020-01-17
#13797. Prevent creation of folders with same name in same path

jgandres:


jgandres:
trigger compilation

src/megaapi_impl.cpp
```diff
@@ -18378,6 +18378,23 @@ void MegaApiImpl::sendPendingRequests()
             const char *name = request->getName();
             if(!name || !(*name) || !parent) { e = API_EARGS; break; }
 
+            bool exists = false;
```
sergiohs84: Something like:
```Node *folder = childnodebyname(parent, name, false);
if (folder && folder->type == FOLDERNODE)
{
    e = API_EEXIST;
    break;
}

File: src/megaapi_impl.cpp [hunk 2]
```diff
@@ -18378,6 +18378,14 @@ void MegaApiImpl::sendPendingRequests()
             const char *name = request->getName();
             if(!name || !(*name) || !parent) { e = API_EARGS; break; }
 
+            // prevent to create a duplicate folder with same name in same path
+            Node *folder = client->childnodebyname(parent, name, false);
+            if (folder && folder->type == FOLDERNODE)
+            {
+                e = API_EEXIST;
+                break;
```
diegocr: Given this is not backward compatible, it will break any existing logic relying on it, won't it?

If this is meant to work like in the webclient, as pointed in the Redmine ticket, please note there we do not fail when the folder exists, we do pretend the operation was successful and we do return the node's handle for the existing folder.

PR #1824, opened 2020-01-17
do include winsock before windows to avoid compilation issues

polmr:
allow logging wstrings


cb-mega:
It'd be great if we could add a unit test for this in `Logging_test.cpp`. Thanks!

PR #1833, opened 2020-01-27
#14318. Notify failed transfer upon overquota

jgandres:
When trying to upload files to a chat in an over quota account , the file appears like trying to upload with spinning wheels and no storage over quota warning

### DIFF
--- a/src/megaapi_impl.cpp
+++ b/src/megaapi_impl.cpp
@@ -11333,7 +11333,7 @@ void MegaApiImpl::transfer_failed(Transfer* t, error e, dstime timeleft, handle 
         // If a Transfer with multiple targets failed, only notify those targets who failed
         if (t->type == PUT
                 && (transfer->getParentHandle() != targetHandle
-                    || targetHandle != UNDEF))
+                    && targetHandle != UNDEF))
         {
             continue;
         }
--- a/src/megaapi_impl.cpp
+++ b/src/megaapi_impl.cpp
@@ -11320,7 +11320,7 @@ char *MegaApiImpl::getFingerprint(MegaNode *n)
     return MegaApi::strdup(n->getFingerprint());
 }
 
-void MegaApiImpl::transfer_failed(Transfer* t, error e, dstime timeleft)
+void MegaApiImpl::transfer_failed(Transfer* t, error e, dstime timeleft, handle targetHandle)
 {
     for (file_list::iterator it = t->files.begin(); it != t->files.end(); it++)
     {
@@ -11329,6 +11329,15 @@ void MegaApiImpl::transfer_failed(Transfer* t, error e, dstime timeleft)
         {
             continue;
         }
+
+        // If a Transfer with multiple targets failed, only notify those targets who failed
+        if (t->type == PUT
+                && (transfer->getParentHandle() != targetHandle
+                    || targetHandle != UNDEF))
+        {
+            continue;
+        }
+
         processTransferFailed(t, transfer, e, timeleft);
     }
 }
--- a/src/megaapi_impl.cpp
+++ b/src/megaapi_impl.cpp
@@ -15794,6 +15794,7 @@ void MegaApiImpl::fireOnTransferFinish(MegaTransferPrivate *transfer, MegaError 
     notificationNumber++;
     transfer->setNotificationNumber(notificationNumber);
     transfer->setLastError(e);
+    transfer->setForeignOverquota(e.getErrorCode() == API_EOVERQUOTA && client->isForeignNode(transfer->getParentHandle()));
 
     if(e.getErrorCode())
     {
--- a/src/megaclient.cpp
+++ b/src/megaclient.cpp
@@ -938,7 +938,6 @@ void MegaClient::activateoverquota(dstime timeleft, handle targetHandle)
                    File *f = (*it++);
                    if (f->h == targetHandle)
                    {
-                       stopxfer(f, mTctableRequestCommitter, API_EOVERQUOTA);
                        continue;
                    }
                 }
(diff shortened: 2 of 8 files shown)


src/megaclient.cpp
```diff
@@ -13603,12 +13603,12 @@ bool MegaClient::startxfer(direction_t d, File* f, DBTableTransactionCommitter&
                 dstime timeleft = dstime(overquotauntil - Waiter::ds);
                 t->failed(API_EOVERQUOTA, committer, timeleft);
             }
-            else if (d == PUT && ststatus == STORAGE_RED)
+            else if (d == PUT && ststatus == STORAGE_RED && isPrivateNode(f->h))
```
sergiohs84: Do not fail the transfer if the upload targets a foreign account (incoming share), since we don't know whether the other account is storage overquota or not (only API knows).

src/megaclient.cpp
```diff
@@ -13712,7 +13712,7 @@ bool MegaClient::startxfer(direction_t d, File* f, DBTableTransactionCommitter&
...
```
sergiohs84: Same than above, but for a transfer that has not started yet

src/transfer.cpp
```diff
@@ -398,29 +398,36 @@ void Transfer::failed(error e, DBTableTransactionCommitter& committer, dstime ti
...
                 bt.backoff(timeleft ? timeleft : NEVER);
-                client->activateoverquota(timeleft);
+                if (client->ststatus == STORAGE_RED && !timeleft)   // already in storage overquota, notify transfer error
+                {
+                    state = TRANSFERSTATE_RETRYING;
+                    slot->retrybt.backoff(NEVER);
+                    slot->retrying = true;
+                    client->app->transfer_failed(this, API_EOVERQUOTA, 0);
+                    ++client->performanceStats.transferTempErrors;
+                }
```
mattw-mega: In order for client->ststatus to be in STORAGE_RED, shouldn't void `MegaClient::activateoverquota` have been called already, which would have iterated through all the transfers and put them in retrying state, and called transfer_failed for the client?   I think we might be doubling up on the notifications here.  Or at least doubling up on notifications for that case.  Perhaps this is intended for a different case where the transfer was started after we are already in overquota state, but I think we should make sure that we don't impact the transition to overquota?  thanks

sergiohs84: That's exactly the case: the `activateoverquota()`, in case there's a transition to storage overquota (not when it's already overquota, when it does nothing), notifies the retrying state for all existing transfers. However, this case is a reported bug: once the account is in overquota, new transfers are not notified via `onTransferTemporaryError(OVERQUOTA)`, so the app doesn't show the error in the UI, but the transfer does not progress :)

mattw-mega: thanks, I see - but for that initial transition to overquota, say we have 30 transfers in progress, the first one of those to discover the account is overquota will call this functoin, activateoverquota() will notify the app for all of them, but then the other 29 will also call this function, I think, and follow the new code path - is it possible for us to differentiate those caess?  thanks

src/megaapi_impl.cpp
```diff
@@ -11329,6 +11329,15 @@ void MegaApiImpl::transfer_failed(Transfer* t, error e, dstime timeleft)
         {
             continue;
         }
+
+        // If a Transfer with multiple targets failed, only notify those targets who failed
```
```suggestion
        // uploads with multiple targets may fail for some targets, but success for other ones --> only notify failed ones
```

src/megaclient.cpp
```diff
@@ -919,26 +919,58 @@ void MegaClient::activateoverquota(dstime timeleft)
                     t->state = TRANSFERSTATE_RETRYING;
                     t->slot->retrybt.backoff(timeleft);
                     t->slot->retrying = true;
-                    app->transfer_failed(t, API_EOVERQUOTA, timeleft);
+                    app->transfer_failed(t, API_EOVERQUOTA, timeleft, targetHandle);
```
sergiohs84: This block is for bandwidth overquota, not for storage. Hence, only for downloads, not uploads.
Do we really need to handle the target handle here? Note that it stores the handle of the node that is being downloaded. I think that whether it's foreign or private is irrelevant here.

src/megaclient.cpp
```diff
@@ -919,26 +919,58 @@ void MegaClient::activateoverquota(dstime timeleft)
...
+                Transfer *t = it->second;
+                for (file_list::iterator it = t->files.begin(); it != t->files.end();)
+                {
+                   File *f = (*it++);
+                   if (f->h == targetHandle)
+                   {
+                       app->transfer_failed(t, API_EOVERQUOTA, 0, targetHandle);
+                       t->removeTransferFile(API_EOVERQUOTA, f, mTctableRequestCommitter);
+                       continue;
+
```
sergiohs84: Empty line

src/megaclient.cpp
```diff
@@ -919,26 +919,58 @@ void MegaClient::activateoverquota(dstime timeleft)
...
+                       ++performanceStats.transferTempErrors;
+                       delete t;
+                   }
+                }
+            }
+            return;
+        }
+        else
+        {
+            setstoragestatus(STORAGE_RED);
```
sergiohs84: Shouldn't we check the return value of `setstoragestatus()` in order to avoid notifying again when we are already on `STORAGE_RED`?? Or is it filtered thanks to the second condition of this `if`??
```
if (t->slot && t->state != TRANSFERSTATE_RETRYING)
```

src/megaclient.cpp
```diff
@@ -919,26 +919,58 @@ void MegaClient::activateoverquota(dstime timeleft)
...
+                       t->finished = true;
+                       app->transfer_removed(t);
+                       ++performanceStats.transferTempErrors;
+                       delete t;
+                   }
+                }
+            }
+            return;
+        }
+        else
```
sergiohs84: If a transfer with private target results in a call to `activateoverquota()` before a call due to a foreign target, all transfers will fail (`transfer_failed()`), including any with a foreign target.

src/megaclient.cpp
jgandres: In this case we are trying to add a new file (with it's target) into an existing transfer. If target is private we need to handle as a local overquota. If the transfer also contains foreign targets, these targets must have been processed when they added and the last parameter in transfer::failed call will prevent to remove any foreign target.

src/megaclient.cpp
jgandres: In this case we there's no existing transfer so the new transfer shouldn't contains any foreign target.

src/megaclient.cpp
jgandres: I think It should be covered by this condition. When you call activateoverquota the first time the storage status will be set as RED and all transfers with slot will be set as retrying. The second time activateoverquota is called when you iterate through the transfers all except the last failed should be in retrying state

src/transfer.cpp
jgandres: In this situation the first failed transfer will call activate overquota and will set storage status as RED, so the other failed transfers will set it's state as retrying, adjust it's backoff and notify the app a temporary error but no call activate overquota again.

PR #1836, opened 2020-01-29
#14399. Allow to move and rename atomically

sergiohs84:
Currently, only the copy operation allows to provide a new name, but the move operation does not. With the improvements for name-collissions, it's desired to allow move and rename in a single request.


src/megaapi_impl.cpp
```diff
@@ -18546,12 +18566,28 @@ void MegaApiImpl::sendPendingRequests()
...
+
+                    string attrstring;
+                    attrs.getjson(&attrstring);
+
+                    SymmCipher key;
+                    key.setkey((const byte*)tc.nn->nodekey.data(), node->type);
+                    client->makeattr(&key, tc.nn->attrstring, attrstring.c_str());
+                }
+
                 client->putnodes(newParent->nodehandle, tc.nn, nc);
```
mattw-mega: Using putnodes here for the file case will still leave the old node, so it's a copy instead of a rename in this case?   I've added 'mv' command to megaclc for easy testing (will make a branch and PR for it) and indeed it copies rather than moves for that case - a pre-existing bug I think.    Also adding some test cases in the test_integration suite would be great (move node is called a couple of times but the changes are not verified, and it doesn't cover both file and folder).

mattw-mega: Acutally, my mistake - I was calling copyNode not moveNode in megaclc.  On retest, that is working fine, I'll approve the ticket.  I'll submit the megaclc branch too as that may be useful in future.  thanks

PR #1838, opened 2020-01-30
removed extra definition of USER_AGENT

aabellagm:
Due to changes in PR 1728 and PR 1758 USER_AGENT ended up declared twice, both in SdkTest_test.cpp and main.cpp, which generates compilation problems.
The one in SdkTest_test.cpp is removed so only the one in main.cpp is left.


mattw-mega:
thanks!

PR #1845, opened 2020-02-04
#14456. cmake for non Windows builds

polmr:
Description
-----------
Improvements in cmake building system
- MEGAsync building with cmake in non win32
- have config.h.in improved with missing cmakedefines & pass some definitions via config.h rather than with hardcoded -DXXXX
- refactoring and lots of changes to provide support for linux & macos
- support for prebuilt libraries in win/macos, vcpkg in win/linux/macos, and system's in linux (&macos?)
- have ffmpeg compiled in !win32 with similar options to currently used
- add option to remove dependency in vcpkg building (to rebuild a dependency) in build3rparty scripts
- fix repeated ifdef. used pragma once instead
- load required qt stuff in sdk if USE_QT preserving upper project config
- windows build will now use generated config.h (enforce -DHAVE_CONFIG_H)
- support openssl >=1.1.0
- correctly support libraw (and included dependent ports)
- define vcpkg subdeps for freeimage, libraw & pdfium, since when not using vcpkg, some subdeps won't be required
- do not include .a if missing (to be able to use system's) for !win

Redmine ticket(s)
-----------------
#14456

Risk area(s)
------------
compilations & all features enabled by third party libs


polmr:
trigger compilation

PR #1846, opened 2020-02-05
#14289. Track affiliate links also during signup

jgandres:


src/megaapi_impl.cpp
```diff
@@ -19724,6 +19718,13 @@ void MegaApiImpl::sendPendingRequests()
             bool resumeProcess = (request->getParamType() == 1);   // resume existing ephemeral account
             bool cancelProcess = (request->getParamType() == 2);
 
+            if (request->getParamType() < mega::MegaApi::AFFILIATE_TYPE_INVALID
```
sergiohs84: We want to provide backwards compatibility. Apps can still create accounts without providing the new parameters, as you still allow the old API without the new params. In consequence, if no public handle is provided, the request should not fail. I suggest to add the `!ISUNDEF(lastPublicHandle)`, something like:
```
if (!ISUNDEF(lastPublicHandle) 
    && lastPublicHandleType <= mega::MegaApi::AFFILIATE_TYPE_INVALID
    && lastPublicHandleType > mega::MegaApi::AFFILIATE_TYPE_CONTACT)
```

include/megaapi.h
```diff
@@ -7747,40 +7747,34 @@ class MegaApi
...
+         * - MegaRequest::getName - Returns the firstname of the user
+         * - MegaRequest::getText - Returns the lastname of the user
          *
          * Valid data in the MegaRequest object received in onRequestFinish when the error code
          * is MegaError::API_OK:
-         * - MegaRequest::getNodeHandle - Returns the nodehandle of the Welcome PDF file, if it
-         * was imported successfully.
+         * - MegaRequest::getSessionKey - Returns the session id to resume the process
+         *
+         * If this request succeeds, a new ephemeral session will be created for the new user
```
sergiohs84:
```suggestion
         * If this request succeeds, a new ephemeral account will be created for the new user
```

PR #1860, opened 2020-02-13
#14515. Include a control character into log messages

polmr:
when ENABLE_LOG_PERFORMANCE
so that the app can discern full from partial messages


sergiohs84:
Jenkins test passed, but it failed due to the race-condition checking useralerts. Fix is in progress for that one.

polmr:
deprecated by https://github.com/meganz/sdk/pull/1871

PR #1862, opened 2020-02-13
#14515. Let the apps decide maximum size of requests payload to log

polmr:


sergiohs84:
trigger compilation

sergiohs84:
Jenkins test passed, but it failed due to the race-condition checking useralerts. Fix is in progress for that one.

sergiohs84:
trigger compilation

PR #1866, opened 2020-02-14
Fix for crash integration tests

sergiohs84:
The `SdkTest` instance is destroyed right after `TearDown`. For tests that instantiate a 3rd instance of `MegaApi`, since it used to not be deleted, the dtor of `SdkTest` destroys it (unique_ptr). During
destruction, it calls back the `SdkTest` --> segfault.


sergiohs84:
trigger compilation

PR #1870, opened 2020-02-17
#13696. Do not create a user alert for permission changes on inshares

sergiohs84:


sergiohs84:
We need to test the case where a new node is added, and confirm whether the node is notified as `TYPE_CHANGE_NEW` or not. If not, we need to adjust the usage of `changed.isnewnode` (set to 1 in ctor, set to 0 in `unserialize()`)

sergiohs84:
> megasdk_PR FAILED (1957) -1

Failed due to the unraided file (unrelated to this PR).

src/node.cpp
```diff
@@ -87,8 +87,8 @@ Node::Node(MegaClient* cclient, node_vector* dp, handle h, handle ph,
 
     plink = NULL;
 
-    memset(&changed,-1,sizeof changed);
-    changed.removed = false;
+    memset(&changed, 0, sizeof changed);
+    changed.newnode = true;
```
polmr: It might be clearer to keep it false in constructor and set the flag at `MegaClient::readnodes` call to new Node.

src/megaclient.cpp
```diff
@@ -5207,9 +5207,11 @@ bool MegaClient::sc_shares()
...
-                        if (!outbound && oh != me && oh && statecurrent)
+                        User* u = finduser(oh);
+                        // only new shares should be notified (skip permissions changes)
+                        bool newShare = u && u->sharing.find(h) == u->sharing.end();
+                        if (!outbound && statecurrent && newShare)
                         {
-                            User* u = finduser(oh);
                             useralerts.add(new UserAlert::NewShare(h, oh, u ? u->email : "", ts, useralerts.nextId()));
                             useralerts.ignoreNextSharedNodesUnder(h);  // no need to alert on nodes already in the new share, which are delivered next
                         }
```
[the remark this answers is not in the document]
sergiohs84: Good spotting. I'll change it.

src/node.cpp
sergiohs84: I like the idea. I'll change it. Thanks

polmr:
at MegaClient::notifynode, when a new node is created (e.g: via remote copy on another client), with the changes, the indicate line is no longer executed: the app does not get notified about the addition on a sync: 
```             if (!n->changed.removed && n->changed.parent)
                {
                    if (!n->localnode)
                    {
                        if (n->type == FOLDERNODE)
                        {
                            app->syncupdate_remote_folder_addition(n->parent->localnode->sync, n);
                        }
                        else
                        {
-------->                app->syncupdate_remote_file_addition(n->parent->localnode->sync, n);
                        }
                    }
                    else
                    {
                        app->syncupdate_remote_move(n->localnode->sync, n,
                            n->localnode->parent ? n->localnode->parent->node : NULL);
                    }
                }
            }```
previously `changed.parent` was true. with the new approach it `if (!n->changed.removed && n->changed.parent)` computes as false

polmr:
https://github.com/meganz/sdk/pull/1874 <- addresses the above

PR #1871, opened 2020-02-17
logging support for partial lines & log without buffering

polmr:
change log call to include parameters that indicate the app that a message is to be continued (partial)
and that needs immediate outputting to disk (after reaching certain threshold size)

### DIFF
--- a/include/mega/logging.h
+++ b/include/mega/logging.h
@@ -179,7 +179,11 @@ class SimpleLogger
         *mBufferIt = '\0';
         if (continues)
         {
-            *(mBufferIt+1) = (char)0x1F;
+            *(mBufferIt + 1) = (char)0x1F;
+        }
+        else
+        {
+            *(mBufferIt + 1) = '\0';
         }
         if (logger)
         {
--- a/include/mega/logging.h
+++ b/include/mega/logging.h
@@ -151,7 +151,7 @@ class SimpleLogger
     static OutputStreams getOutput(enum LogLevel ll);
 #else
     std::array<char, 256> mBuffer; // will be stack-allocated since SimpleLogger is stack-allocated
-    std::array<char, 257>::iterator mBufferIt;
+    std::array<char, 256>::iterator mBufferIt;
     using DiffType = std::array<char, 256>::difference_type;
     using NumBuf = std::array<char, 24>;
     const char* filenameStr;
@@ -163,9 +163,9 @@ class SimpleLogger
         DiffType start = 0;
         while (currentSize > 0)
         {
-            const auto size = std::min(currentSize, std::distance(mBufferIt, mBuffer.end() - 1));
+            const auto size = std::min(currentSize, std::distance(mBufferIt, mBuffer.end() - 2));
             mBufferIt = std::copy(dataIt + start, dataIt + start + size, mBufferIt);
-            if (mBufferIt == mBuffer.end() - 1)
+            if (mBufferIt == mBuffer.end() - 2)
             {
                 outputBuffer(true);
             }
--- a/include/mega/logging.h
+++ b/include/mega/logging.h
@@ -156,6 +156,9 @@ class SimpleLogger
     using NumBuf = std::array<char, 24>;
     const char* filenameStr;
     int lineNum;
+    int outputted = 0;
+    static const int immediate_threshold = 1024;
+    bool immediate = false;
 
     template<typename DataIterator>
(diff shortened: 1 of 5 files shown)


include/mega/logging.h
```diff
@@ -124,8 +124,13 @@ class Logger
 {
 public:
     virtual ~Logger() = default;
+#ifdef ENABLE_LOG_PERFORMANCE
+    // Note: `time` and `source` are null in performance mode
+    virtual void log(const char *time, int loglevel, const char *source, const char *message, bool partial = false, bool requiresDirectOutput = false) = 0;
+#else
     // Note: `time` and `source` are null in performance mode
```
cb-mega: we should remove this comment

include/mega/logging.h
```diff
@@ -124,8 +124,13 @@ class Logger
 {
 public:
     virtual ~Logger() = default;
+#ifdef ENABLE_LOG_PERFORMANCE
+    // Note: `time` and `source` are null in performance mode
+    virtual void log(const char *time, int loglevel, const char *source, const char *message, bool partial = false, bool requiresDirectOutput = false) = 0;
```
cb-mega: shall we remove `time` and `source` since they're gonna be null anyways?

include/mega/logging.h
```diff
@@ -156,16 +161,19 @@ class SimpleLogger
     using NumBuf = std::array<char, 24>;
     const char* filenameStr;
     int lineNum;
+    int outputted = 0;
+    static const int immediate_threshold = 1024;
+    bool immediate = false;
```
cb-mega: Can we name our variables according to our style as `mMember`?

include/mega/logging.h
```diff
@@ -179,15 +187,18 @@ class SimpleLogger
...
         if (continues)
         {
-            *(mBufferIt + 1) = (char)0x1F;
-        }
-        else
-        {
-            *(mBufferIt + 1) = '\0';
+            if (outputted > immediate_threshold)
+            {
+                immediate = true;
```
cb-mega: Wouldn't we have to reset `outputted` back to 0 here?

Also: Wouldn't the client be able to determine when to immediate output? Do we really need to pass the `immediate` flag?

include/mega/logging.h
cb-mega: Is `partial` really a good parameter name since all messages longer than 255 are partial, even the last chunk? Maybe invert and call it `final` instead?

polmr: ? it's actually the other way around: messages shorter than the buffer are !partial (i.e. the message continues): `outputBuffer` is only called with `continues = true` when copyToBuffer reaches the end of the buffer.

include/mega/logging.h
polmr: sure!

include/mega/logging.h
polmr: why the reset? on the contrary, the idea is that from a certain threshold, the message parts are to be outputed directly.

polmr: The thing is that the client does not know if what's being logged is going to be large or not. that's why I took the decision into the source. The client (i.e. the app, is just consuming some log lines -actually log parts- to have then outputted into the log files). We could consider that the client keeps a track of number of "continuation" messages from the same thread, and move the decision back to it. But it might be overcomplicating that.

cb-mega: Oh ok. I might have misunderstood the idea. Maybe add a comment? I would have expected that the chunks that are being immediately output are all the same size but instead the first one only is 1024 and the following ones are 256 (or less).

include/mega/logging.h
cb-mega: Ok. I guess what confuses me is that the parameter name should be a descriptor of the current message but here it holds information about all messages? Say you have two chunks then the last chunk has `partial = false` even though it is still a partial message.

polmr: I understand: partial is confusing. renaming to `continues`

include/mega/logging.h
polmr: ? all chunks are 256 (or less), but the first 4 would be allowed to be kept in the outputting buffer. The fifth will require immediate output before returning the execution. I am increasing the threshold, since this is something we don't want to happen in UI threads, but only in the rare occasions where we have very long lines (requests output).

polmr:
Deprecated by : https://github.com/meganz/sdk/pull/1889

PR #1873, opened 2020-02-18
#14159. Ensure the sync's resumable flag is not reset

cb-mega:
Previously, if a client would call `fetchNodes` they would cause all sync configs to become unresumable so a subsequent call to `fetchNodesAndResumeSyncs` would not resume syncs. This PR fixes that by adding a simple boolean to MegaClient.


sergiohs84:
trigger compilation

PR #1874, opened 2020-02-19
Use newnode instead of parent when evalving changes to notify sync ap…

polmr:
…ps of new nodes creations

### DIFF
--- a/src/megaclient.cpp
+++ b/src/megaclient.cpp
@@ -9326,7 +9326,7 @@ void MegaClient::notifynode(Node* n)
         // deletion from LocalNodes.
         if (n->localnode && n->localnode->parent && n->parent && !n->parent->localnode)
         {
-            if (n->changed.removed || n->changed.parent)
+            if (n->changed.removed || n->changed.newnode)
             {
                 if (n->type == FOLDERNODE)
                 {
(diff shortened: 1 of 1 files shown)


src/megaclient.cpp
```diff
@@ -9326,7 +9326,7 @@ void MegaClient::notifynode(Node* n)
         // deletion from LocalNodes.
         if (n->localnode && n->localnode->parent && n->parent && !n->parent->localnode)
         {
-            if (n->changed.removed || n->changed.parent)
+            if (n->changed.removed || n->changed.newnode)
```
sergiohs84: I think you actually want to keep the code as it used to be here. The callbacks below are for nodes that are moved outside of the sync --> parent has changed.

sergiohs84: I think you actually want to keep the code as it used to be here. The callbacks below are for nodes that are moved outside of the sync --> parent has changed.
```suggestion
            if (n->changed.removed || n->changed.parent)
```

src/megaclient.cpp
```diff
@@ -9359,7 +9359,7 @@ void MegaClient::notifynode(Node* n)
                     n->localnode->deleted = n->changed.removed;
                 }
 
-                if (!n->changed.removed && n->changed.parent)
+                if (!n->changed.removed && n->changed.newnode)
```
sergiohs84: I think it could also be a new parent. Depending on the nature of the movement, it could result in a `copy`+`delete` or in a `move`. So...
```suggestion
                if (!n->changed.removed && (n->changed.newnode || n->changed.parent))
```

PR #1877, opened 2020-02-20
WIP: #14640. Ensure we don't add/resume syncs with vanished remote nodes

cb-mega:


cb-mega:
Good points! I added the removal of the sync config. The callback to the app should be in the next ticket tho which deals with the move of all sync management to the SDK. Currently, everything is still working. I haven't changed how apps are notified. Cheers

src/megaclient.cpp
```diff
@@ -3718,6 +3718,7 @@ void MegaClient::resumeResumableSyncs()
         if (!nodebyhandle(config.getRemoteNode()))
         {
             // remote node gone
+            syncConfigs->remove(config.getLocalPath());
```
sergiohs84: The sync should not be removed in this case. It would change the current behavior in MEGAsync (just try to remove the node relative to a synced folder and you will see the SyncConfig stays there, but disabled). Removing is not an option here. Please, disable the sync instead of remove.

cb-mega: Ah yes. Will change to that. Thanks!

src/megaclient.cpp
```diff
@@ -3709,12 +3709,19 @@ void MegaClient::resumeResumableSyncs()
...
         {
             continue;
         }
+        if (!nodebyhandle(config.getRemoteNode()))
+        {
+            // remote node gone
+            config.setResumable(false);
+            syncConfigs->insert(config);
+            continue;
+        }
```
sergiohs84: Below this comment, if the sync fails to be resumed (else of `if (e == 0)`), shouldn't we update the status / resumable-flag of the sync (and notify the app accordingly)? That way, the app will learn the sync was not resumed by "reason"

cb-mega: Yes, this is done now.

src/megaapi_impl.cpp
```diff
@@ -12819,30 +12819,41 @@ bool MegaApiImpl::sync_syncable(Sync *sync, const char *name, string *localpath)
...
+        s->appData = sync;
+        sync->setState(s->state);
+        syncMap[-nextTag] = sync;
     }
+    else
+    {
+        sync->setState(SYNC_FAILED);
+    }
+
+    fireOnRequestFinish(request, MegaError(e));
```
sergiohs84: Don't need to create a `MegaRequest` and notify here. Simply report the event on a callback, so the app is aware of the _sync resume result_ for this `sync` :)

cb-mega: Which is what we're doing in the next line. However, for this version I'd like to keep the `fireOnRequestFinish` so that I don't have to change much in megasync. Otherwise it seems there's a lot of immediate changes necessary and I need to get on with one-way-sync.

sergiohs84: We can put this PR/task on hold until you have time to focus back on it. I dislike the idea of adding temporary changes for a feature that presumably will take a while to be completed (so those temporary changes may end up in the code base forever). It's very common that changes in the SDK require changes in the app and the SDK cannot be merged until the apps have been adapted and are ready to merge as well. This is just one more case :)

polmr:
closed in favour of https://github.com/meganz/sdk/pull/1955

PR #1882, opened 2020-02-20
#14641. Reset committer on dbtable destruction to avoid use-after-free

cb-mega:


include/mega/db.h
```diff
@@ -35,9 +35,11 @@ class MEGA_API DbTable
 
 protected:
     bool mCheckAlwaysTransacted = false;
-    DBTableTransactionCommitter* mCurrentTransactionCommiter = nullptr;
+    DBTableTransactionCommitter* mTransactionCommitter = nullptr;
```
cb-mega: Better naming

include/mega/db.h
```diff
@@ -88,7 +90,7 @@ class MEGA_API DBTableTransactionCommitter
     bool mStarted = false;
 
 public:
-    inline void beginOnce()
+    void beginOnce()
```
cb-mega: the methods are already inline. No need for superfluous `inline` keyword.

include/mega/db.h
```diff
@@ -109,30 +111,37 @@ class MEGA_API DBTableTransactionCommitter
...
             }
             else
             {
-                mTable->mCurrentTransactionCommiter = this;
+                mTable->mTransactionCommitter = this;
             }
         }
     }
+
+    MEGA_DISABLE_COPY_MOVE(DBTableTransactionCommitter)
```
cb-mega: Without this the committer is copyable but accidentally copying it would be UB.

src/db.cpp
```diff
@@ -84,17 +84,28 @@ void DbTable::checkTransaction()
...
+            mTransactionCommitter->beginOnce();
         }
     }
 }
 
+void DbTable::resetCommitter()
+{
+    if (mTransactionCommitter)
+    {
+        mTransactionCommitter->commitNow();
```
sergiohs84: We may not want to commit the open transaction in all cases. Remember the `sctable` should only commit upon reception of a new `scsn` via actionpackets (or a bit later when the ongoing cs requests complete).
Please, check that it's not possible to do a commit by mistake, or the account's state may get corrupted. Thanks

cb-mega: Yeah, I think you're right. In fact, I think we shouldn't commit here at all as this function is only called from the table destructor. Anything that wasn't committed until then should be discarded (I assume that's the reason why we're calling `abort` in the table destructor?!).

cb-mega: Ok. Done. And now the method name makes more sense :)

PR #1883, opened 2020-02-21
#14658. Prevent to confirm accounts other than the logged-in

sergiohs84:
The SDK must check the confirmation link does not correspond to a different account than the logged-in account. If so, it should return `API_EACCESS`. Additionally, for a confirmation request when the SDK is logged into the account that wants to be confirmed, it should return `API_EEXPIRED`.


sergiohs84:
Do not merge this PR yet, wait for iOS (and perhaps Android) to finish some testing.

sergiohs84:
trigger compilation

PR #1885, opened 2020-02-24
Prevent the discovered write_after_free by disabling the committer be…

mattw-mega:
…fore calling locallogout()

Rather than trying to cope with locallogout() clearing all the data structures when the call stack may still have an interest in those structures.
Assert added so we don't accidentally cause this again.
One unneeded commitNow() removed, the db->del() below will invoke it anyway, if needed.
A couple of warnings addressed (in 32 bit)

### DIFF
--- a/src/megaapi_impl.cpp
+++ b/src/megaapi_impl.cpp
@@ -19112,6 +19112,7 @@ void MegaApiImpl::sendPendingRequests()
             }
             else
             {
+                committer.commitNow(true);
                 client->locallogout(false);
                 client->restag = nextTag;
                 logout_result(API_OK);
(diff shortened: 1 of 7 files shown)


src/megaapi_impl.cpp
```diff
@@ -19112,6 +19112,7 @@ void MegaApiImpl::sendPendingRequests()
             }
             else
             {
+                committer.commitNow(true);
```
sergiohs84: We may not want to commit the open transaction in all cases. Remember the `sctable` should only commit upon reception of a new `scsn` via actionpackets (or a bit later when the ongoing `cs` requests complete).

In this case, I think to commit the ongoing transaction is incorrect and the account's state may get corrupted. Thanks

mattw-mega: Yes you're right that we must be very careful about committing sctable.   This one would only be used for tctable though.

PR #1888, opened 2020-02-25
Fix mem leaks in integration tests

cb-mega:
Gave move semantics to `NewNode`.

### DIFF
--- a/include/mega/node.h
+++ b/include/mega/node.h
@@ -33,6 +33,8 @@ struct MEGA_API NodeCore
     NodeCore();
     ~NodeCore();
 
+    MEGA_DISABLE_COPY_MOVE(NodeCore)
+
     // node's own handle
     handle nodehandle;
 
@@ -67,8 +69,16 @@ struct MEGA_API NewNode : public NodeCore
 
     bool added;
 
+    MEGA_DISABLE_COPY(NewNode)
+
+    NewNode(NewNode&& o);
+    NewNode& operator=(NewNode&& o);
+
     NewNode();
     ~NewNode();
+
+private:
+    void swap(NewNode&& o);
 };
 
 struct MEGA_API PublicLink
--- a/include/mega/node.h
+++ b/include/mega/node.h
@@ -30,23 +30,17 @@
 namespace mega {
 struct MEGA_API NodeCore
 {
-    NodeCore();
-    ~NodeCore();
-
-    MEGA_DISABLE_COPY_MOVE(NodeCore)
-
     // node's own handle
-    handle nodehandle;
+    handle nodehandle = UNDEF;
 
     // parent node handle (in a Node context, temporary placeholder until parent is set)
-    handle parenthandle;
+    handle parenthandle = UNDEF;
 
     // node type
-    nodetype_t type;
+    nodetype_t type = TYPE_UNKNOWN;
 
     // node attributes
-    string *attrstring;
-
+    std::unique_ptr<string> attrstring;
 };
 
 // new node for putnodes()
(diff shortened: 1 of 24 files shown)


include/mega/node.h
```diff
@@ -33,6 +33,8 @@ struct MEGA_API NodeCore
     NodeCore();
     ~NodeCore();
 
+    MEGA_DISABLE_COPY_MOVE(NodeCore)
```
mattw-mega: Since the aim is to make NewNode moveable, we should not disable move in its base class here, but allow it to be moved also, and have NewNode's move constructor use NodeCore's move constructor.    You can get all that for free (along with implicitly disabling copy), and remove the need for the destructor here by protecting the only non-POD member attrstring by a unique_ptr instead of a raw pointer.  (rule of 0)

include/mega/node.h
```diff
@@ -67,8 +69,16 @@ struct MEGA_API NewNode : public NodeCore
 
     bool added;
 
+    MEGA_DISABLE_COPY(NewNode)
```
mattw-mega: no need to disable copy here; its base class has it disabled so it's disabled here too unless you write one yourself

include/mega/node.h
```diff
@@ -67,8 +69,16 @@ struct MEGA_API NewNode : public NodeCore
...
+
+    NewNode(NewNode&& o);
+    NewNode& operator=(NewNode&& o);
```
mattw-mega: NewNode's only owned member is fileattributes, so you can do the same thing here and just make that one a unique_pointer which automatically disables copy and allows the default move constructor/assign to work. (rule of 0).  No need to write and maintain explicit move/assign when the default is correct anyway.

include/mega/node.h
```diff
@@ -67,8 +69,16 @@ struct MEGA_API NewNode : public NodeCore
...
+
     NewNode();
     ~NewNode();
+
+private:
+    void swap(NewNode&& o);
```
mattw-mega: swap should not be needed - I see it's used from the move constructor, but there's no need to copy this object's members back to the source object

src/node.cpp
```diff
@@ -39,15 +39,57 @@ NewNode::NewNode()
...
+    fileattributes = nullptr;
 }
 
 NewNode::~NewNode()
 {
     delete fileattributes;
 }
 
+NewNode::NewNode(NewNode&& o)
+    : NewNode() // assign defaults
```
mattw-mega: we should use the base class's move constructor when moving this class

src/node.cpp
```diff
@@ -39,15 +39,57 @@ NewNode::NewNode()
...
+{
+    swap(std::move(o));
```
mattw-mega: no need to copy this class's (brand new) members back to the move source

src/node.cpp
```diff
@@ -39,15 +39,57 @@ NewNode::NewNode()
...
+        }
+        localnode = nullptr;
+        if (fileattributes)
+        {
+            delete fileattributes;
+            fileattributes = nullptr;
+        }
+        swap(std::move(o));
+    }
+    return *this;
```
mattw-mega: if we use unique_ptr we don't need to write this function at all, it's provided automatically by the compiler

src/node.cpp
```diff
@@ -39,15 +39,57 @@ NewNode::NewNode()
...
+    std::swap(attrstring, o.attrstring);
+    std::swap(nodekey, o.nodekey);
+    std::swap(source, o.source);
+    std::swap(ovhandle, o.ovhandle);
+    std::swap(uploadhandle, o.uploadhandle);
+    std::swap(uploadtoken, o.uploadtoken);
+    std::swap(syncid, o.syncid);
+    std::swap(localnode, o.localnode);
+    std::swap(fileattributes, o.fileattributes);
+    std::swap(added, o.added);
```
mattw-mega: no need for this function, unique_ptr will provide it automatically.   Also no need to copy members back to the move source.

PR #1889, opened 2020-02-25
#14515: support direct logging

polmr:
have Logger::log include a vector of const char* to pass to the app for
direct logging
add DirectMessage class to wrap const char* messages when sent to log
via LOG_XXXX macros
small messages will still be logged normally
SimpleLogger will gather a vector of all message parts that need direct
outputing and pass it to Logger::log upon its destruction

### DIFF
--- a/include/mega/logging.h
+++ b/include/mega/logging.h
@@ -125,10 +125,42 @@ class Logger
 public:
     virtual ~Logger() = default;
     // Note: `time` and `source` are null in performance mode
-    virtual void log(const char *time, int loglevel, const char *source, const char *message) = 0;
+    virtual void log(const char *time, int loglevel, const char *source, const char *message
+#ifdef ENABLE_LOG_PERFORMANCE)
+          , std::vector<const char *> directMessages = std::vector<const char *>()
+#endif
+                     ) = 0;
 };
 
 typedef std::vector<std::ostream *> OutputStreams;
+
+/**
+ * @brief holds a const char * and its size to pass to SimpleLogger, to use the direct logging logic
+ */
+class DirectMessage{
+public:
+    size_t mSize = 0;
+    const char *mConstChar;
+
+    DirectMessage( const char *constChar, bool force = false) //force will set size as max, so as to stay above directMsgThreshold
+    {
+        mConstChar = constChar;
+        mSize = force?SIZE_MAX:strlen(constChar);
+    }
+
+    template<typename T, typename std::enable_if<std::is_integral<T>::value>::type* = nullptr>
+    DirectMessage( const char *constChar, T size)
+    {
+        mConstChar = constChar;
+        mSize = size;
+    }
+
+    size_t size() const
+    {
+        return mSize;
+    }
+
+};
 
 class OutputMap : public std::array<OutputStreams, unsigned(logMax)+1> {};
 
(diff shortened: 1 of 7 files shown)


include/mega/logging.h
```diff
@@ -150,13 +182,18 @@ class SimpleLogger
     static OutputMap outputs;
     static OutputStreams getOutput(enum LogLevel ll);
 #else
-    std::array<char, 256> mBuffer; // will be stack-allocated since SimpleLogger is stack-allocated
-    std::array<char, 256>::iterator mBufferIt;
-    using DiffType = std::array<char, 256>::difference_type;
+    const static int LOGGER_CHUNKS_SIZE = 1024;
```
cb-mega: On Android your stack is only 8KB

include/mega/logging.h
```diff
@@ -125,11 +125,43 @@ class Logger
...
+    size_t mSize = 0;
+    const char *mConstChar;
+
+    DirectMessage( const char *constChar, bool force = false) //force will set size as max, so as to stay above directMsgThreshold
+    {
+        mConstChar = constChar;
+        mSize = force?SIZE_MAX:strlen(constChar);
+    }
+
+    template<typename T, typename std::enable_if<std::is_integral<T>::value>::type* = nullptr>
```
cb-mega: I believe `bool` is also considered integral

include/mega/logging.h
```diff
@@ -125,11 +125,43 @@ class Logger
...
+#endif
+                     ) = 0;
 };
 
 typedef std::vector<std::ostream *> OutputStreams;
 
+/**
+ * @brief holds a const char * and its size to pass to SimpleLogger, to use the direct logging logic
+ */
+class DirectMessage{
```
cb-mega: Can this class use proper encapsulation?

include/mega/logging.h
```diff
@@ -330,6 +371,11 @@ class SimpleLogger
         logValue(lineNum);
         copyToBuffer("]", 1);
         outputBuffer();
+
+        if (!mDirectMessages.empty() && logger)
+        {
+            logger->log(nullptr, level, nullptr, "", mDirectMessages);
```
cb-mega: `move(mDirectMessages)`

include/mega/logging.h
```diff
@@ -423,6 +469,29 @@ class SimpleLogger
     }
 #endif
 
+
+
+    SimpleLogger& operator<<(const DirectMessage &obj)
```
cb-mega: Very elegant approach :+1:

include/mega/logging.h
polmr: DirectMessage( const char *constChar, bool force = false) is called when passing a `bool`. Otherwise, that second one. The idea is that `DirectMessage( const char *constChar, bool force = false)` is not called when another `is_integral` type is used

include/mega/logging.h
polmr: we should bear in mind if using PERFORMANCE_MODE for android

include/mega/logging.h
polmr: good point. I'll probably use const vector & as parameters instead, to avoid vector copying

polmr:
Deprecated by: https://github.com/meganz/sdk/pull/1895

PR #1890, opened 2020-02-25
makeattr is called with unique_ptr every time but one, so make the sy…

mattw-mega:
…ntax easy

(also it will mean many fewer changed lines to review in the fix/mem_leaks_in_integration_tests PR)

### DIFF
--- a/include/mega/megaclient.h
+++ b/include/mega/megaclient.h
@@ -401,6 +401,9 @@ public:
     error setattr(Node*, const char* prevattr = NULL);
 
     // prefix and encrypt attribute json
+    void makeattr(SymmCipher*, string*, const char*, int = -1) const;
+
+    // convenience version of the above (used for all but one call)
     void makeattr(SymmCipher*, std::unique_ptr<string> const &, const char*, int = -1) const;
 
     // check node access level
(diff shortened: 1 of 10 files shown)


include/mega/megaclient.h
```diff
@@ -403,6 +403,9 @@ class MEGA_API MegaClient
     // prefix and encrypt attribute json
     void makeattr(SymmCipher*, string*, const char*, int = -1) const;
 
+    // convenience version of the above (used for all but one call)
```
cb-mega: your comment in parenthesis will become stale tomorrow. Please remove

src/megaclient.cpp
```diff
@@ -6800,6 +6800,11 @@ void MegaClient::makeattr(SymmCipher* key, string* attrstring, const char* json,
     delete[] buf;
 }
 
+void MegaClient::makeattr(SymmCipher* key, std::unique_ptr<string> const & attrstring, const char* json, int l) const
```
cb-mega: we normally put the `const` before the type: `const std::unique_ptr<string>&`

include/mega/megaclient.h
cb-mega: we normally put the `const` before the type: `const std::unique_ptr<string>&`

PR #1900, opened 2020-03-04
Fix mem leak in MegaSyncEventPrivate

cb-mega:


include/megaapi_impl.h
```diff
@@ -919,20 +919,20 @@ class MegaContactRequestPrivate : public MegaContactRequest
 
 #ifdef ENABLE_SYNC
 
-class MegaSyncEventPrivate: public MegaSyncEvent
+class MegaSyncEventPrivate : public MegaSyncEvent
 {
 public:
+    explicit
     MegaSyncEventPrivate(int type);
```
mattw-mega: this is just one declaration; it should be just one line.

include/megaapi_impl.h
```diff
@@ -919,20 +919,20 @@ class MegaContactRequestPrivate : public MegaContactRequest
...
-    virtual int getType() const;
-    virtual const char *getPath() const;
-    virtual MegaHandle getNodeHandle() const;
-    virtual const char *getNewPath() const;
-    virtual const char* getPrevName() const;
-    virtual MegaHandle getPrevParent() const;
+    int getType() const override;
+    const char* getPath() const override;
+    MegaHandle getNodeHandle() const override;
+    const char* getNewPath() const override;
```
mattw-mega: No need to switch the space and * around; others are switching them to be the other way in other PRs, and it's all unnecessary - just leave those as they are, keep the PR change to a minimum.

src/megaapi_impl.cpp
```diff
@@ -23066,61 +23056,49 @@ int MegaSyncEventPrivate::getType() const
...
+    return path.get();
 }
 
 MegaHandle MegaSyncEventPrivate::getNodeHandle() const
 {
     return nodeHandle;
 }
 
-const char *MegaSyncEventPrivate::getNewPath() const
+const char* MegaSyncEventPrivate::getNewPath() const
```
mattw-mega: no need to be switching space and * around

include/megaapi_impl.h
cb-mega: Is that part of our coding guideline?

include/megaapi_impl.h
cb-mega: `const char*` is in our coding guideline, `const char *` is not so we should change it if we're making changes anyway.

src/megaapi_impl.cpp
cb-mega: `const char*` is in our coding guideline, `const char *` is not so we should change it if we're making changes anyway.

PR #1907, opened 2020-03-10
#15064. Use directory entry info instead of fopen() when possible, to save CPU

mattw-mega:
fopen() uses a lot of CPU, on windows at least, so if we can get all the needed info from the directory iterating data, we can save that entirely.
On windows, for a large sync, that means MEGAsync can do its initial scan on startup much more quickly.


src/posix/fs.cpp
```diff
@@ -1950,7 +1950,18 @@ bool PosixDirAccess::dnext(string* path, string* name, bool followsymlinks, node
         {
             path->append(d->d_name);
 
-            if (followsymlinks ? !stat(path->c_str(), &statbuf) : !lstat(path->c_str(), &statbuf))
+            bool statOk = !lstat(path->c_str(), &statbuf);
+            if (statOk && S_ISLNK(statbuf.st_mode) && followsymlinks)
```
polmr, line 1954: very minor perf improvement:

```suggestion
            if (followsymlinks && statOk && S_ISLNK(statbuf.st_mode))
```

mattw-mega: will make this change shortly, thanks

PR #1914, opened 2020-03-11
add tentative timed lock option for the sdk mutex

polmr:


include/megaapi.h
```diff
@@ -6707,6 +6707,18 @@ class MegaApiLock
...
+    /**
+     * @brief Tries to lock the MegaApi if this instance does not currently have a lock on it yet.
+     *
+     * If the lock is succeeded in the expected time, the behaviour is the same as lockOnce().
+     *
+     * @param time Milliseconds to wait for locking
+     *
+     * @return if the locking succeded
+     */
+    bool try_lock_for(long long time);
```
sergiohs84: Can we assign a name more in line with the rest of methods in `MegaApi`?
```suggestion
    bool tryLockFor(long long time);
```

include/megaapi_impl.h
```diff
@@ -2597,6 +2597,7 @@ class MegaApiImpl : public MegaApp
         void yield();
         void lockMutex();
         void unlockMutex();
+        bool tryLockMutex_for(long long time);
```
sergiohs84:
```suggestion
        bool tryLockMutexFor(long long time);
```

PR #1916, opened 2020-03-12
#14903. Store LocalNode shortnames in the localnode database.

mattw-mega:
The SDK was looking up the shortnames on the loading of a Sync, which doesn't take much time for a small sync but for a million node sync takes around 4 minutes (on SSD).
This PR stores those in the database so they can be retrieved much more quickly on startup - the million node sync localnodes load in about 10 seconds.
The first time a sync is loaded after this change, the localnodes still have to have the shortname looked up, so the first time is still slow.  The localnodes are written back to the database periodically (after every 50000).
Also the shortname is not stored in RAM anymore if it is the same as the localnode's localname, saving some memory.


src/sync.cpp
```diff
@@ -837,13 +837,34 @@ void Sync::addstatecachechildren(uint32_t parent_dbid, idlocalnode_map* tmap, st
...
         l->size = size;
         l->setfsid(fsid, client->fsidnode);
         l->setnode(node);
 
+        if (!l->slocalname_in_db)
+        {
+            statecacheadd(l);
+            if (insertq.size() > 50000)
+            {
+                cachenodes();  // periodically output updated nodes with shortname updates, so people who restart megasync still make progress towards a fast startup
```
sergiohs84, line 864: Just a comment. Is it safe to save intermediate updates in permanent cache? It's a great idea to save shortnames during the initial scanning for a faster startup upon restarts, but I'm concerned about applying some changes too early (before completing the scanning), close the app, and upon restart finding an inconsistent state. Perhaps intermediates are harmless, but just in case, to be considered :thinking:

src/node.cpp
```diff
@@ -1815,77 +1788,30 @@ LocalNode* LocalNode::unserialize(Sync* sync, const string* d)
...
-    if (ptr < end)
+    if (!r.unserializehandle(fsid) ||
+        !r.unserializeu32(parent_dbid) || 
+        !r.unserializenodehandle(h) ||
+        !r.unserializestring(localname) ||
+        (type == FILENODE && !r.unserializebinary((byte*)crc, sizeof(crc))) ||
+        (type == FILENODE && !r.unserializecompressed64(mtime)) ||
+        (r.hasdataleft() && !r.unserializebyte(syncable)) ||
+        (r.hasdataleft() && !r.unserializeexpansionflags(expansionflags, 1)) ||
+        (expansionflags[0] && !r.unserializestring(shortname)))
```
sergiohs84, line 1809: `shortname` was serialized by using a `serializecstring()`. Shouldn't we use `unserializecstr()` here?

src/megaclient.cpp
```diff
@@ -952,6 +952,16 @@ void MegaClient::activateoverquota(dstime timeleft)
     looprequested = true;
 }
 
+std::unique_ptr<string> MegaClient::fsShortname(string& localname)
```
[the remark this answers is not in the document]
mattw-mega: Yes that makes sense, I've moved it, thanks

src/sync.cpp
```diff
@@ -837,13 +837,34 @@ void Sync::addstatecachechildren(uint32_t parent_dbid, idlocalnode_map* tmap, st
...
+        // if we already have the shortname from database, use that, otherwise (db is from old code) look it up
+        std::unique_ptr<string> shortname;
+        if (l->slocalname_in_db)
+        {
+            shortname.reset(l->slocalname.release());
+        }
+        else
+        {
+            shortname.reset(new string());
+            client->fsaccess->getsname(path, shortname.get());
```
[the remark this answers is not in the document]
mattw-mega: Good point, adjusted.

src/utils.cpp
```diff
@@ -94,6 +101,11 @@ void CacheableWriter::serializehandle(handle field)
     dest.append((char*)&field, sizeof(field));
 }
 
+void CacheableWriter::serializenodehandle(handle field)
+{
+    dest.append((const char*)&field, MegaClient::NODEHANDLE);
```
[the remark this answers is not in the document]
mattw-mega: I'm not sure about this one -  I think an alternative is to add functions for those specific cases when needed, so `serializechathandle()` etc, and we could make chat handles (and others) an explicit separate type too, to avoid accidental use of one in the context of another, and also to make doubly sure we write the correct entry to the database now, and enable the compiler to alert us to any mismatch between the type used and the type written to the database when making changes.   Does that sound good?  thanks

src/utils.cpp
```diff
@@ -319,6 +331,21 @@ bool CacheableReader::unserializechunkmacs(chunkmac_map& m)
     return false;
 }
 
+bool CacheableReader::unserializecompressed64(uint64_t& field)
+{
+    int fieldize;
```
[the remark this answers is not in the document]
mattw-mega: updated :-)  thanks for spotting that.

src/sync.cpp
mattw-mega: I thought it was better to do it this way, because for very large syncs it can take a long time.  If the user kills and restarts MEGAsync, it has to start over, unless we do something like this where we update the database as we go.  In the normal case where all the localnodes already have their shortname looked up, it has no effect because there are no changed localnodes as we load.   It's a minor point though so I don't mind removing it, if you prefer?  thanks

src/node.cpp
mattw-mega: Yes good point, I prefer those symmetries too.  Adjusted, thanks

src/utils.cpp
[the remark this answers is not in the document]
sergiohs84: That approach would be even better, yes. There have been small issues in the past due by misusing wrong size. However, my concern is that redefining the datatype for chat_link_handles, contact_link_handles and node_handles would require too many changes here and there. Something that requires time and that I'd prefer to not be part of this PR.
By now, I think we can leave it as your current code in this branch. And perhaps improve all that in a future ;)

src/sync.cpp
sergiohs84: Not sure about this one. I understand the advantage when MEGAsync is killed. I'm unsure whether saving partial changes could create any inconsistency afterwards. I leave it at your will :)

mattw-mega: I think it's ok as it's the same mechanism we use for any other update of a LocalNode's database entry, and the loading code for a LocalNode looks to see if the entry has a shortname present or not.  Let's include it in testing for now and see how it goes, thanks!

PR #1917, opened 2020-03-12
#14965: Reuse previously built dependencies

aabellagm:
If incremental flag is enabled and previously built dependencies are
found in a backup directory, they are copied and reused in the current
build.
In case a dependency MD5 changes (e.g. new version), the new dependency
is built and copied back into the backup directory.

*Risk areas*
----------------
package building mechanism that use the incremental building (param `-I`). So, barely none

### DIFF
--- a/contrib/build_sdk.sh
+++ b/contrib/build_sdk.sh
@@ -930,7 +930,7 @@ mediainfo_pkg() {
     local mediainfolib_dir_extract="MediaInfoLib-$mediainfolib_ver"
     local mediainfolib_dir="MediaInfoLib-$mediainfolib_ver/Project/GNU/Library"
 
-    if [ $incremental -eq 1 ] && [ -e $status_dir/$mediainfolib_name.success -a `cat $status_dir/$mediainfolib_name.success` = $mediainfolib_md5]; then
+    if [ $incremental -eq 1 ] && [ -e $status_dir/$mediainfolib_name.success ] && [ `cat $status_dir/$mediainfolib_name.success` = $mediainfolib_md5]; then
         echo "$mediainfolib_name already built"
         return
     else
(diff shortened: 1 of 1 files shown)


contrib/build_sdk.sh
```diff
@@ -338,7 +340,7 @@ openssl_pkg() {
     fi
     local loc_make_opts=$make_opts
 
-    if [ $incremental -eq 1 ] && [ -e $status_dir/$name.success ]; then
+    if [ $incremental -eq 1 ] && [ -e $status_dir/$name.success ] && [ `cat $status_dir/$name.success` = $openssl_md5 ]; then
```
polmr: There's a little backwards compatibility issue with this & similar changes:
for systems relying on $name.success in the old ways (content = `0`), this will fail.

contrib/build_sdk.sh
```diff
@@ -930,7 +930,7 @@ mediainfo_pkg() {
     local mediainfolib_dir_extract="MediaInfoLib-$mediainfolib_ver"
     local mediainfolib_dir="MediaInfoLib-$mediainfolib_ver/Project/GNU/Library"
 
-    if [ $incremental -eq 1 ] && [ -e $status_dir/$mediainfolib_name.success -a `cat $status_dir/$mediainfolib_name.success` = $mediainfolib_md5]; then
+    if [ $incremental -eq 1 ] && [ -e $status_dir/$mediainfolib_name.success ] && [ `cat $status_dir/$mediainfolib_name.success` = $mediainfolib_md5]; then
```
polmr:
```suggestion
    if [ $incremental -eq 1 ] && [ -e $status_dir/$mediainfolib_name.success ] && [ `cat $status_dir/$mediainfolib_name.success` = $mediainfolib_md5 ]; then
```

PR #1919, opened 2020-03-12
#14906. Process API_EBLOCKED on sc channel

sergiohs84:
Rather than waiting for a cs request that triggers a `API_EBLOCKED` when the account has been blocked, the API sends the correspondieng -16 on the sc channel, so we can immediately handle the situation


sergiohs84:
trigger compilation

sergiohs84:
trigger compilation

PR #1923, opened 2020-03-13
Potential hotfixed release v3.6.6b (patching v3.6.6a for MEGAsync 4.3.0)

polmr:
temporary solution to avoid fopen calls for all paths in syncup
if no symlink is ever found

### DIFF
--- a/src/megaclient.cpp
+++ b/src/megaclient.cpp
@@ -12566,7 +12566,7 @@ bool MegaClient::syncup(LocalNode* l, dstime* nds)
         rit = nchildren.find(&localname);
 
         bool isSymLink = false;
-
+#ifndef WIN32
         if (PosixFileAccess::mFoundASymlink)
         {
             string localpath;
@@ -12577,7 +12577,7 @@ bool MegaClient::syncup(LocalNode* l, dstime* nds)
             isSymLink = fa->mIsSymLink;
             LOG_debug << "checking if localpath is symlink: " << localpath;
         }
-
+#endif
         // do we have a corresponding remote child?
         if (rit != nchildren.end())
         {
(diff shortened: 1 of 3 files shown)


src/megaclient.cpp
```diff
@@ -12565,12 +12565,17 @@ bool MegaClient::syncup(LocalNode* l, dstime* nds)
...
-        ll->getlocalpath(&localpath);
-        fa->fopen(&localpath);
-        bool isSymLink = fa->mIsSymLink;
+        if (PosixFileAccess::mFoundASymlink)
+        {
+            string localpath;
+            unique_ptr<FileAccess> fa(fsaccess->newfileaccess(false));
+
+            ll->getlocalpath(&localpath);
+            fa->fopen(&localpath);
```
polmr: LOG_debug << "checking if localpath is symlink: " << localpath;

src/megaclient.cpp
```diff
@@ -12565,13 +12565,19 @@ bool MegaClient::syncup(LocalNode* l, dstime* nds)
...
+#ifndef WIN32
+        if (PosixFileAccess::mFoundASymlink)
+        {
+            string localpath;
+            unique_ptr<FileAccess> fa(fsaccess->newfileaccess(false));
 
+            ll->getlocalpath(&localpath);
+            fa->fopen(&localpath);
+            isSymLink = fa->mIsSymLink;
+            LOG_debug << "checking if localpath is symlink: " << localpath;
```
mattw-mega: Although it is useful to know if we are checking that, outputting the message here for every check may overwhelm the log, since syncup() will call this for every single file/folder in the sync as it recurses through the tree - for very large accounts syncing everything, that will make it hard to read the logs, and cause more CPU and I/O.

polmr: Agree. We can have the log message only when we first set mFoundASymlink to true instead

PR #1930, opened 2020-03-20
Address warnings from logging.h when trying to build test_integration…

mattw-mega:
… for Mac

Some types were mismatched and caused lots of signed/unsigned warnings (repeating since they were in the header)
Also found lots of areas where the ENABLE_LOG_PERFORMANCE was not checked, and various projects failed to build if that was turned on.
Removed Logging_test.cpp code since that was not building with that on, and we are using and checking logging all the time anyway.
Added some casts in net.cpp to ensure we use the size_t version of DirectMessage() constructor, and highlight that we are truncating the size argument at that point (for those targets that have 32 bit size_t).

### DIFF
--- a/examples/megasimplesync.cpp
+++ b/examples/megasimplesync.cpp
@@ -114,7 +114,11 @@ public:
 
     // Logger interface
 public:
-    void log(const char *time, int loglevel, const char *source, const char *message);
+    void log(const char *time, int loglevel, const char *source, const char *message
+        #ifdef ENABLE_LOG_PERFORMANCE
+            , const char **directMessages = nullptr, size_t *directMessagesSizes = nullptr, unsigned numberMessages = 0
+        #endif
+    );
 };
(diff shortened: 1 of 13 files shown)


mattw-mega:
trigger compilation

examples/megasimplesync.cpp
```diff
@@ -361,7 +369,13 @@ void SyncApp::log(const char *time, int loglevel, const char *source, const char
         message = "";
     }
 
+#ifndef ENABLE_LOG_PERFORMANCE
     cout << "[" << time << "][" << SimpleLogger::toStr((LogLevel)loglevel) << "] " << message << endl;
+#else
+    cout << "[" << time << "][" << SimpleLogger::toStr((LogLevel)loglevel) << "] ";
```
polmr: missing `<< message;` at the end of this line

PR #1932, opened 2020-03-23
#15027. Fix/useralerts on their own sc channel

mattw-mega:
The c=50 useralert request occured on the `wsc` channel but should be on its own `sc` channel without any timeout (and the `wsc` system must wait for it before continuing)
Added a pendingsc50 channel to match the pendingsc channel and moved the relevant logic there.
Using unique_ptr to ensure ownership issues are addressed, so there are some small changes to pendingsc.
Also found a case where btcs should have been btsc.

### DIFF
--- a/include/mega/megaclient.h
+++ b/include/mega/megaclient.h
@@ -802,7 +802,8 @@ private:
     vector<TimerWithBackoff *> bttimers;
 
     // server-client command trigger connection
-    HttpReq* pendingsc;
+    std::unique_ptr<HttpReq> pendingsc;
+    std::unique_ptr<HttpReq> pendingsc50;
     BackoffTimer btsc;
     bool stopsc = false;
 
--- a/include/mega/megaclient.h
+++ b/include/mega/megaclient.h
@@ -806,6 +806,8 @@ private:
     std::unique_ptr<HttpReq> pendingscUserAlerts;
     BackoffTimer btsc;
     bool stopsc = false;
+    bool pendingscTimedOut = false;
+
 
     // badhost report
     HttpReq* badhostcs;
--- a/include/mega/megaclient.h
+++ b/include/mega/megaclient.h
@@ -803,7 +803,7 @@ private:
 
     // server-client command trigger connection
     std::unique_ptr<HttpReq> pendingsc;
-    std::unique_ptr<HttpReq> pendingsc50;
+    std::unique_ptr<HttpReq> pendingscUserAlerts;
     BackoffTimer btsc;
     bool stopsc = false;
 
--- a/src/megaclient.cpp
+++ b/src/megaclient.cpp
@@ -1912,7 +1912,7 @@ void MegaClient::exec()
             break;
         }
 
-        // handle API server-client requests
+        // handle the request for the last 50 UserAlerts
         if (pendingscUserAlerts)
         {
             switch (pendingscUserAlerts->status)
@@ -1934,22 +1934,19 @@ void MegaClient::exec()
 
                 // fall through
             case REQ_FAILURE:
-                if (pendingscUserAlerts->httpstatus != 200)
+                if (pendingscUserAlerts->httpstatus == 200)
                 {
-                    (pendingscUserAlerts->httpstatus == 500 ? fnstats.e500Count : fnstats.eOthersCount)++;
-                }
-                else
-                {
-                    error e = (error)atoi(pendingscUserAlerts->in.c_str());
+                    error e = (error)atoi(pendingscUserAlerts->in.c_str());  
                     if (e == API_EAGAIN || e == API_ERATELIMIT)
                     {
-                        fnstats.eAgainCount++;
                         btsc.backoff();
                         pendingscUserAlerts.reset();
+                        LOG_warn << "Backing off before retrying useralerts request: " << btsc.retryin();
                         break;
                     }
                     LOG_err << "Unexpected sc response: " << pendingscUserAlerts->in;
                 }
+                LOG_err << "Useralerts request failed, continuing without them";
                 if (useralerts.begincatchup)
                 {
                     useralerts.begincatchup = false;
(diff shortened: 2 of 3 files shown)


examples/megacli.cpp
```diff
@@ -2778,11 +2778,20 @@ class MegaCLILogger : public ::mega::Logger {
...
+#ifdef ENABLE_LOG_PERFORMANCE
+                 , const char **directMessages, size_t *directMessagesSizes, unsigned numberMessages
+#endif
+    ) override
     {
         if (mLogFile.is_open())
         {
-            mLogFile << Waiter::ds << " " << SimpleLogger::toStr(static_cast<LogLevel>(loglevel)) << ": " << message << std::endl;
+            mLogFile << Waiter::ds << " " << SimpleLogger::toStr(static_cast<LogLevel>(loglevel)) << ": ";
+            if (message) mLogFile << message;
```
sergiohs84: Please, stick to the code style.

include/mega/megaclient.h
```diff
@@ -802,7 +802,8 @@ class MEGA_API MegaClient
     vector<TimerWithBackoff *> bttimers;
 
     // server-client command trigger connection
-    HttpReq* pendingsc;
+    std::unique_ptr<HttpReq> pendingsc;
+    std::unique_ptr<HttpReq> pendingsc50;
```
sergiohs84:
```suggestion
    std::unique_ptr<HttpReq> pendingscUserAlerts;
```

examples/megacli.cpp
mattw-mega: updated

include/mega/megaclient.h
mattw-mega: updated

src/megaclient.cpp
```diff
@@ -3804,10 +3847,9 @@ void MegaClient::catchup()
     {
         pendingsc->disconnect();
 
-        delete pendingsc;
-        pendingsc = NULL;
+        pendingsc.reset();
     }
-    btcs.reset();
+    btsc.reset();
```
sergiohs84: Oh, good catch!! :man_facepalming:

src/megaclient.cpp
```diff
@@ -2025,11 +2069,12 @@ void MegaClient::exec()
                 break;
 
             case REQ_INFLIGHT:
-                if (Waiter::ds >= (pendingsc->lastdata + HttpIO::SCREQUESTTIMEOUT))
+                if (!pendingscTimedOut && Waiter::ds >= (pendingsc->lastdata + HttpIO::SCREQUESTTIMEOUT))
                 {
                     LOG_debug << "sc timeout expired";
-                    delete pendingsc;
-                    pendingsc = NULL;
+                    // In almost all cases the server won't take more than 40 seconds.  But if it does, break the cycle of endless requests for the same thing
```
```suggestion
                    // In almost all cases the server won't take more than SCREQUESTTIMEOUT seconds.  But if it does, break the cycle of endless requests for the same thing
```
Otherwise, in case we change it in the future, the `40` might be misleading :)

src/megaclient.cpp
```diff
@@ -1915,18 +1913,69 @@ void MegaClient::exec()
...
+                    }
+                    pendingscUserAlerts.reset();
+                    break;
+                }
+
+                // fall through
+            case REQ_FAILURE:
+                if (pendingscUserAlerts->httpstatus != 200)
+                {
+                    (pendingscUserAlerts->httpstatus == 500 ? fnstats.e500Count : fnstats.eOthersCount)++;
```
sergiohs84: This code is to count errors for the fetchnodes statistics. IIRC, the user-alerts are requested after the fetchnodes completes, so we should not update `fnstats` in result of errors getting the useralerts (well, it should be harmless, but useless as well, since the stats have been sent already).

Same for `fnstats.eAgainCount++;` in line 1946.

src/megaclient.cpp
```diff
@@ -1915,18 +1913,69 @@ void MegaClient::exec()
...
+                }
+                else
+                {
+                    error e = (error)atoi(pendingscUserAlerts->in.c_str());
+                    if (e == API_EAGAIN || e == API_ERATELIMIT)
+                    {
+                        fnstats.eAgainCount++;
```
```suggestion
```

src/megaclient.cpp
```diff
@@ -1915,18 +1913,69 @@ void MegaClient::exec()
...
+                        btsc.backoff();
+                        pendingscUserAlerts.reset();
```
sergiohs84: If we reset the `pnedingscUserAlerts`, it means that upon reception of these (retriable) errors from API during retrieval of UAs we will not retry at all, correct? Is it the expected behavior? In case the API is busy and returns a temporal -3, useralerts will not be fetched.

mattw-mega: The reset there is just to get back to the state before we sent this request - after the btsc timer has finished waiting and is armed again, the code to send the next wsc/sc request will be activated, and since useralerts.begincatchup is still true, it will send the UserAlerts request again.   If the error is not -3 (or ratelimit) then we abandon the useralerts request since it's not essential, by also setting useralerts.begincatchup false (and useralerts.catchupdone true).  After adjusting for the above comments, I've stepped through again to verify (and added some log warnings so we can see it in the logs if needed).  thanks!

src/megaclient.cpp
mattw-mega: removed

src/megaclient.cpp
mattw-mega: Good to know, I was just keeping that from the `wsc` handling but in that case we don't need it, I've removed those lines, thanks

src/megaclient.cpp
mattw-mega: updated

src/megaclient.cpp
sergiohs84: Oh, I see. My bad, I didn't think about the `begincatchup`. It makes perfect sense now :)

sergiohs84:
trigger compilation

PR #1933, opened 2020-03-24
#15016. Add 32/64 bit architecture of App and OS to UserAgent

mattw-mega:
eg. "MEGAsync/4.3.1.0 (Windows 10.0.18362/64) MegaClient/3.6.9/32".
/64 and /32 are the additions to the existing format.
/32 on the end denotes whether the app is 32 bit or 64 bit (so could also be /64 (or empty currently, for any other architecture))
And the /64 inside the brackets would only be present when the OS architecture differs from the app (so for windows, only when the app is 32 bit on 64 bit, under the WOW system)

### DIFF
--- a/include/mega/filesystem.h
+++ b/include/mega/filesystem.h
@@ -339,7 +339,7 @@ struct MEGA_API FileSystemAccess : public EventTrigger
     bool target_exists;
 
     // append local operating system version information to string
-    virtual void osversion(string*) const { }
+    virtual void osversion(string*, bool includeArchitecture) const { }
 
     // append id for stats
     virtual void statsid(string*) const { }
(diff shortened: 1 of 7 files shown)


include/mega/filesystem.h
```diff
@@ -339,7 +339,7 @@ struct MEGA_API FileSystemAccess : public EventTrigger
     bool target_exists;
 
     // append local operating system version information to string
-    virtual void osversion(string*) const { }
+    virtual void osversion(string*, bool includeArchitecture) const { }
```
polmr: the name is deceiving I'm afraid: in the end it only includes info when using the Wow64 emulator ...
May I suggest: `includeArchExtraInfo` (`includeArchExtraInfoWhenSignificant` is too large ;-) ), and adding a word on WinFileSystemAccess::osversion prototype regarding `WOW64` ...

PR #1937, opened 2020-03-27
WIP: #15019.Import new messages from a different cache

sergiohs84:
MEGAchat needs to access to an external DB that might be accessed by a different thread.
This PR should be merged together with https://github.com/meganz/MEGAchat/pull/827


sergiohs84:
Actually, that's the option by default, so no need to overwrite. We actually need the WAL mode to write from a thread and read from another concurrently. Closing this PR.

PR #1938, opened 2020-03-30
#14637. Stop escaping filenames when the special characters are supported by the target FS

jgandres:


jgandres:
trigger compilation

src/filesystem.cpp
```diff
@@ -48,14 +48,115 @@ bool FileSystemAccess::islchex(char c) const
...
+                return FS_APPLE;
+            case NTFS_SB_MAGIC:
+                return FS_WIN;
+            default:
+                return FS_DEFAULT;
+        }
+    }
+#elif defined  (__APPLE__)
+    struct statfs fileStat;
+    if (!statfs(dstPath->c_str(),&fileStat))
```
alber2510, line 95:
```suggestion
    if (!statfs(dstPath->c_str(), &fileStat))
```

src/filesystem.cpp
```diff
@@ -111,10 +217,47 @@ void FileSystemAccess::unescapefsincompatible(string* name) const
...
+
 // escape forbidden characters, then convert to local encoding
-void FileSystemAccess::name2local(string* filename) const
+void FileSystemAccess::name2local(string* filename, string *dstPath) const
 {
-    escapefsincompatible(filename);
+    std::string path = dstPath ? (*dstPath) : "";
+    if (filename && dstPath && dstPath->size() > filename->size())
+    {
+        try
```
alber2510, line 245: I think that it's easier to do with `std::string::findr`. In this way you can check if the name of the file is the last token in `dstPath`. Furthermore, you avoid a possible exception

src/filesystem.cpp
```diff
@@ -111,10 +217,41 @@ void FileSystemAccess::unescapefsincompatible(string* name) const
     }
 }
 
+std::string FileSystemAccess::getPathSeparator(const char *path)
```
mattw-mega, line 220: Hi, I think this one should be platform specific - eg. on unix a \ is allowed in a file's name, and is not a separator.  thanks

src/filesystem.cpp
```diff
@@ -70,16 +171,21 @@ void FileSystemAccess::escapefsincompatible(string* name) const
...
     unsigned char c;
+    int fileSystemType = getlocalfstype(dstPath);
 
     // replace all occurrences of a badchar with %xx
     for (size_t i = name->size(); i--; )
     {
         c = (unsigned char)(*name)[i];
 
-        if (!islocalfscompatible(c))
+        if (!islocalfscompatible(c, fileSystemType))
```
mattw-mega, line 181: I think we need to take utf-8 encoding into account here - eg. if a utf-8 character is encoded across two bytes, the second byte could match the char "/", but overall that two-byte character should not be changed.

src/filesystem.cpp
jgandres: This idea behind this method is look for the first occurrence of \ or / so:
- If you are in Windows the paths you will provide are like this: C:\\ (First separator found is \)
- If you are in Unix/APPLE the paths you will provide are like this: /home/folder (First separator found is /)

jgandres: Anyway as you said this method should be platform specific, so I think this code makes it easier and more robust.
```
std::string FileSystemAccess::getPathSeparator(const char *path)
{
#if defined (__linux__) || defined (__ANDROID__) || defined  (__APPLE__)
return "/";
#elif defined(_WIN32) || defined(_WIN64)
return "\\";
#elif 
// Default case    
return "\\/";
#endif
}
```
What do you think?

src/filesystem.cpp
jgandres: You are right this need to be adjusted, thanks so much for the comments ;)

src/filesystem.cpp
mattw-mega: yes that looks fine to me, nice one!  thanks

src/filesystem.cpp
jgandres: @mattw-mega could you have a look to the latest commits, I think that solves the problem with multi-by characters:

include/mega/filesystem.h
```diff
@@ -266,7 +293,10 @@ struct MEGA_API FileSystemAccess : public EventTrigger
     void local2name(string*) const;
 
     // convert local path to MEGA format (UTF-8) with unescaping
-    void name2local(string*) const;
+    void name2local(string*, std::string *dstPath = nullptr) const;
```
mattw-mega, line 296: Can we make the 2nd parameter be `const std:;string*` please?  Initially I thought we were changing it to optionally output the result to the 2nd string instead of editing in place, like path2local etc.  thanks!

include/megaapi_impl.h
```diff
@@ -2429,7 +2429,7 @@ class MegaApiImpl : public MegaApp
         static int typeComparator(Node *i, Node *j);
         static bool userComparatorDefaultASC (User *i, User *j);
 
-        char* escapeFsIncompatible(const char *filename);
+        char* escapeFsIncompatible(const char *filename, const char *dstPath);
         char* unescapeFsIncompatible(const char* name);
```
mattw-mega, line 2433: I see these two are asymmetric now, we need to know the fileysystem for encoding but not decoding.   We could add some explanation for that in the megaapi.h comments, and also it makes me wonder what if the original filename included a % and had a couple of hex-compatible characters - we would not alter those for encode, but decode would collapse them to a single character.    Probably we should escape % itself in escapeFsIncompatible if is followed by two hex characters, so that unescape always gives you back the original?   thanks

src/filesystem.cpp
```diff
@@ -69,18 +180,24 @@ void FileSystemAccess::escapefsincompatible(string* name) const
...
+        if (utf8seqsize == 1 && !islocalfscompatible(c, fileSystemType))
         {
+            std::string inc = name->substr(i, 1);
             sprintf(buf, "%%%02x", c);
             name->replace(i, 1, buf);
+            LOG_debug << "Incompatible character for filesystem type "
+                      << fstypetostring(fileSystemType)
+                      << ", replace '" << inc << "' by '" << buf << "'\n";
         }
+        i += utf8seqsize;
```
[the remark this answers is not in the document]
jgandres: Fixed in a0c21e3

include/mega/filesystem.h
jgandres: Fixed in a0c21e3

jgandres:
trigger compilation

src/filesystem.cpp
```diff
@@ -48,15 +48,147 @@ bool FileSystemAccess::islchex(char c) const
...
+        case FS_WIN:
+        default:
+            // NTFS restricted characters => 0x0000-0x001F 0x007F " * / : < > ? \ |
+            // If filesystem couldn't be detected we'll use a restrictive charset to avoid issues.
+            return !strchr("\\/:?\"<>|*", c);
+    }
+}
+
+string FileSystemAccess::getValidPath(const string *path) const
+{
```
sergiohs84, line 169: Consider to improve performance of this method, since it will be called multiple times

src/filesystem.cpp
```diff
@@ -48,15 +48,147 @@ bool FileSystemAccess::islchex(char c) const
...
+        }
+    }
+    return validPath;
 }
 
 // replace characters that are not allowed in local fs names with a %xx escape sequence
-void FileSystemAccess::escapefsincompatible(string* name) const
+void FileSystemAccess::escapefsincompatible(string* name, const string *dstPath) const
 {
+    string validPath = getValidPath(dstPath);
```
sergiohs84, line 191: Move it below the `if`s, since they may `return`.

src/filesystem.cpp
```diff
@@ -69,18 +201,24 @@ void FileSystemAccess::escapefsincompatible(string* name) const
...
-    unsigned char c;
-
-    // replace all occurrences of a badchar with %xx
-    for (size_t i = name->size(); i--; )
-    {
-        c = (unsigned char)(*name)[i];
-
-        if (!islocalfscompatible(c))
+    int fileSystemType = getlocalfstype(&validPath);
+    size_t utf8seqsize = 0, i = 0;
```
sergiohs84, line 205:
```suggestion
    size_t utf8seqsize = 0;
    size_t i = 0;
```

src/filesystem.cpp
```diff
@@ -69,18 +201,24 @@ void FileSystemAccess::escapefsincompatible(string* name) const
...
+    unsigned char c = '0';
+    while (i < name->size())
+    {
+        c = static_cast<unsigned char>((*name)[i]);
+        utf8seqsize = Utils::utf8SequenceSize(c);
+        assert (utf8seqsize);
+        if (utf8seqsize == 1 && !islocalfscompatible(c, fileSystemType))
         {
+            std::string inc = name->substr(i, 1);
```
sergiohs84, line 214: I think `inc` should be declared as a `char` (and with a name like `incompatibleChar` or alike)

sergiohs84: You can simply replace the `inc` in the log by `name->substr(i, 1);`

jgandres: I've forgotten the reason behind std::string inc. name is modified so I need to store the current value and I can't  replace in the log by name->substr(i, 1);

jgandres:
trigger compilation

jgandres:
trigger compilation

jgandres:
trigger compilation

PR #1940, opened 2020-04-01
#15134. Download transfer blocked after fail another transfer with API_EBLOCKED

jgandres:


src/transfer.cpp
```diff
@@ -463,7 +463,7 @@ void Transfer::failed(error e, DBTableTransactionCommitter& committer, dstime ti
          * the actionpacket will eventually remove the target and the sync-engine will force to
          * disable the synchronization of the folder. For non-sync-transfers, remove the file directly.
          */
-        if (e == API_EARGS)
+        if (e == API_EARGS || e == API_EBLOCKED)
         {
              File *f = (*it++);
              if (f->syncxfer)
```
sergiohs84, line 469:
```suggestion
             if (f->syncxfer && e == API_EARGS)
```
Otherwise, you will defer sync transfers and they will get stuck in the same way (read the comment above the `if`).

jgandres:
trigger compilation

PR #1953, opened 2020-04-16
performance: no need to update the LocalNode tree sync status during …

mattw-mega:
…MegaClient destruction.

Saves minutes during MEGAsync shutdown with a million node sync.


mattw-mega:
trigger compilation

PR #1959, opened 2020-04-20
#15363. Add escaping for JSON double-quoted strings

sergiohs84:
...and unescape them for `usabd` command

### DIFF
--- a/src/command.cpp
+++ b/src/command.cpp
@@ -77,11 +77,26 @@ void Command::arg(const char* name, const char* value, int quotes)
     addcomma();
     json.append("\"");
     json.append(name);
-    json.append(quotes ? "\":\"" : "\":");
-    json.append(value);
     if (quotes)
     {
-        json.append("\"");
+        json.append("\":\"");
+        size_t valueLen = strlen(value);
+        for (size_t i = 0; i < valueLen; i++)
+        {
+            if (value[i] == '\"'
+                    || value[i] == '\\'
+                    || value[i] == '/')
+            {
+                json.push_back('\\');
+            }
+            json.push_back(value[i]);
+        }
+        json.push_back('\"');
+    }
+    else
+    {
+        json.append("\":");
+        json.append(value);
     }
 }
 
--- a/src/commands.cpp
+++ b/src/commands.cpp
@@ -7648,6 +7648,7 @@ void CommandGetRegisteredContacts::processResult(MegaApp& app, JSON& json)
                 case MAKENAMEID3('e', 'u', 'd'):
                 {
                     json.storeobject(&entryUserDetail);
+                    JSON::unescape(&entryUserDetail);
                     break;
                 }
                 case MAKENAMEID2('i', 'd'):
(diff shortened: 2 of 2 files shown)


src/command.cpp
```diff
@@ -77,11 +77,26 @@ void Command::arg(const char* name, const char* value, int quotes)
...
     if (quotes)
     {
-        json.append("\"");
+        json.append("\":\"");
+        size_t valueLen = strlen(value);
+        for (size_t i = 0; i < valueLen; i++)
+        {
+            if (value[i] == '\"'
+                    || value[i] == '\\'
+                    || value[i] == '/')
```
mattw-mega, line 88: According to the JSON spec (https://www.ietf.org/rfc/rfc4627.txt, Strings section), we don't need to escape /, but we do need to escape any control codes in the string.

src/command.cpp
```diff
@@ -77,11 +77,26 @@ void Command::arg(const char* name, const char* value, int quotes)
...
     json.append("\"");
     json.append(name);
-    json.append(quotes ? "\":\"" : "\":");
-    json.append(value);
     if (quotes)
     {
-        json.append("\"");
+        json.append("\":\"");
+        size_t valueLen = strlen(value);
+        for (size_t i = 0; i < valueLen; i++)
```
mattw-mega, line 84: Do we need to take UTF-8 into account here, and increment by character size in bytes?  eg might we alter a character that takes two or more bytes to represent but isn't one of those below?

src/command.cpp
sergiohs84: Yep, according to the spec those are the chars that should be escaped for Strings:
```
RFC 4627                          JSON                         July 2006

         string = quotation-mark *char quotation-mark

         char = unescaped /
                escape (
                    %x22 /          ; "    quotation mark  U+0022
                    %x5C /          ; \    reverse solidus U+005C
                    %x2F /          ; /    solidus         U+002F
                    %x62 /          ; b    backspace       U+0008
                    %x66 /          ; f    form feed       U+000C
                    %x6E /          ; n    line feed       U+000A
                    %x72 /          ; r    carriage return U+000D
                    %x74 /          ; t    tab             U+0009
                    %x75 4HEXDIG )  ; uXXXX                U+XXXX
```
I only added the printable chars with the specific case we aim to solve in mind. But a proper escaping should include all the above, indeed.

src/command.cpp
sergiohs84: This is a good point. The specs say:
```The representation of strings is similar to conventions used in the C family of programming languages. A string begins and ends with quotation marks. All Unicode characters may be placed within the quotation marks except for the characters that must be escaped: quotation mark, reverse solidus, and the control characters (U+0000 through U+001F).```
Hence, it seems like UTF-8 chars don't need (and should not) be escaped, unless they are the `"`, the `\` or control characters.

sergiohs84:
Since it's been agreed the JSON parser in the SDK should not be changed, but API to enforce B64 for any string, this PR is closed.

PR #1962, opened 2020-04-21
#15291. Disallow to remove contacts from business accounts

alber2510:


src/user.cpp
```diff
@@ -95,7 +95,15 @@ bool User::serialize(string* d)
     d->append(email.c_str(), l);
 
     d->append((char*)&attrVersion, 1);
-    d->append("\0\0\0\0\0\0", 7);
+
+    char bizMode = 0;
+    if (mBizMode != BIZ_MODE_UNKNOWN) // convert number to ascii
```
sergiohs84, line 100: In case a user has `mBizMode == BIZ_MODE_UNKNOWN`, the value in cache will be `0`. Upon `unserialize()` the value user will be a `BIZ_MODE_SUBUSER`.

sergiohs84: I think there's no need of an specific management of this value as a char, but directly assign the value. I've tested and it works. I've push a new commit for that.

PR #1967, opened 2020-04-23
#15355. More fingerprint accuracy when deciding to copy rather than upload a file

dwmega:
The result of this change-set is to perform a more detailed comparison of two files that have matching fingerprints. This is necessary as file content is sparsely hashed and so two differing files can sometimes yield the same fingerprint.

When a remote node is found with a filename and fingerprint matching a local file, we compare the two by computing the metamac of the local file using the remote node's key. If the metamac matches, we know the two are identical and as such, can avoid uploading the local file and create a copy in the cloud.

If we are unable to compute the metamac or if the metamac ~matches~ does not match, we consider the files distinct. This may result in some files being uploaded needlessly but it will ensure file integrity.

An appropriate unit test is included that exercises these changes.

Note that file corruption can still occur as the metamac comparison is only performed when the local and remote nodes have the same filename. That is, if we upload a file A that fingerprints to X and proceed to sync a different file B that also fingerprints to X, B will not be uploaded and a copy of A will be made in the cloud.

### DIFF
--- a/examples/megacli.cpp
+++ b/examples/megacli.cpp
@@ -3003,6 +3003,7 @@ autocomplete::ACN autocompleteSyntax()
     p->Add(exec_showattributes, sequence(text("showattributes"), remoteFSPath(client, &cwd)));
 
     p->Add(exec_setmaxconnections, sequence(text("setmaxconnections"), either(text("put"), text("get")), opt(wholenumber(4))));
+    p->Add(exec_metamac, sequence(text("metamac"), localFSPath(), remoteFSPath(client, &cwd)));
 
     return autocompleteTemplate = std::move(p);
 }
@@ -7963,5 +7964,68 @@ void DemoAppFolder::nodes_updated(Node **n, int count)
 
     cout << "The folder link contains ";
     nodestats(c[1], "");
+}
+
+void exec_metamac(autocomplete::ACState& s)
+{
+    Node *node = nodebypath(s.words[2].s.c_str());
+    if (!node || node->type != FILENODE)
+    {
+        cerr << s.words[2].s
+             << (node ? ": No such file or directory"
+                      : ": Not a file")
+             << endl;
+        return;
+    }
+
+    auto ifaccess = client->fsaccess->newfileaccess();
+    {
+        std::string local_path;
+
+        client->fsaccess->path2local(&s.words[1].s, &local_path);
+
+        if (!ifaccess->fopen(&local_path, 1, 0))
+        {
+            cerr << "Failed to open: " << s.words[1].s << endl;
+            return;
+        }
+    }
+
+    SymmCipher cipher;
+    int64_t remote_iv;
+    int64_t remote_mm;
+
+    {
+        std::string remote_key = node->nodekey();
+        const char *iva = &remote_key[SymmCipher::KEYLENGTH];
+
+        cipher.setkey((byte*)&remote_key[0], node->type);
+        remote_iv = MemAccess::get<int64_t>(iva);
+        remote_mm = MemAccess::get<int64_t>(iva + sizeof(int64_t));
+    }
+
+    auto result = generate_metamac(cipher, *ifaccess, remote_iv);
+    if (!result.first)
+    {
+        cerr << "Failed to generate metamac for: "
+             << s.words[1].s
+             << endl;
+    }
+    else
+    {
+        const std::ios::fmtflags flags = cout.flags();
+
+        cout << s.words[2].s
+             << ": "
+             << std::hex
+             << (uint64_t)remote_mm
+             << "\n"
+             << s.words[1].s
+             << ": "
+             << (uint64_t)result.second
+             << endl;
+
+        cout.flags(flags);
+    }
 }
 
--- a/examples/megacli.cpp
+++ b/examples/megacli.cpp
@@ -7978,13 +7978,12 @@ void exec_metamac(autocomplete::ACState& s)
         return;
     }
 
-    auto ifaccess = client->fsaccess->newfileaccess();
+    auto ifAccess = client->fsaccess->newfileaccess();
     {
-        std::string local_path;
+        std::string localPath;
+        client->fsaccess->path2local(&s.words[1].s, &localPath);
 
-        client->fsaccess->path2local(&s.words[1].s, &local_path);
-
-        if (!ifaccess->fopen(&local_path, 1, 0))
+        if (!ifAccess->fopen(&localPath, 1, 0))
         {
             cerr << "Failed to open: " << s.words[1].s << endl;
             return;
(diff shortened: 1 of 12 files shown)


include/mega/filesystem.h
```diff
@@ -172,6 +172,18 @@ struct MEGA_API InputStreamAccess
...
+class MEGA_API FileInputStream : public InputStreamAccess
+{
+    FileAccess *fileAccess;
+    m_off_t offset;
+
+public:
+    FileInputStream(FileAccess *fileAccess);
+    virtual m_off_t size();
+    virtual bool read(byte *buffer, unsigned size);
+    virtual ~FileInputStream();
```
mattw-mega, line 184: Since we are moving this class, we could update the interface to use the 'override' keyword rather than `virtual`, it was written pre-c++11.   Since the destructor is empty I think we can remove that too?  thanks

src/megaclient.cpp
```diff
@@ -14006,6 +14006,67 @@ Node* MegaClient::nodebyfingerprint(FileFingerprint* fingerprint)
...
+    fsaccess->local2name(&local_name);
+    
+    // Only compare metamac if the node doesn't already exist.
+    node_vector::const_iterator remote_node =
+      std::find_if(remote_nodes->begin(),
+                   remote_nodes->end(),
+                   [&](const Node *remote_node) -> bool
+                   {
+                       return local_node->type == remote_node->type
+                              && local_name == remote_node->displayname();
```
sergiohs84, line 14027: Why do we want to check the name? If the fingerprint matches, it looks like the same data, so that's one of the cases where we want to check the metamac. Perhaps I'm missing something, but I actually see the metamac comparison very useful for those files that are updated just a few bytes, so same size+fingerprint+name applies and we skip the update. We want, in that case, to check the metamac. Please, correct me if I'm wrong.

examples/megacli.cpp
```diff
@@ -7956,3 +7957,66 @@ void DemoAppFolder::nodes_updated(Node **n, int count)
...
+             << (node ? ": No such file or directory"
+                      : ": Not a file")
+             << endl;
+        return;
+    }
+
+    auto ifaccess = client->fsaccess->newfileaccess();
+    {
+        std::string local_path;
+
```
sergiohs84, line 7975: Empty line

examples/megacli.cpp
```diff
@@ -7956,3 +7957,66 @@ void DemoAppFolder::nodes_updated(Node **n, int count)
...
+        cerr << "Failed to generate metamac for: "
+             << s.words[1].s
+             << endl;
+    }
+    else
+    {
+        const std::ios::fmtflags flags = cout.flags();
+
+        cout << s.words[2].s
+             << ": "
```
sergiohs84, line 8010:
```suggestion
             << " (remote): "
```

examples/megacli.cpp
```diff
@@ -7956,3 +7957,66 @@ void DemoAppFolder::nodes_updated(Node **n, int count)
...
+             << std::hex
+             << (uint64_t)remote_mm
+             << "\n"
+             << s.words[1].s
+             << ": "
```
sergiohs84, line 8015:
```suggestion
             << " (local): "
```

src/megaclient.cpp
```diff
@@ -14006,6 +14006,67 @@ Node* MegaClient::nodebyfingerprint(FileFingerprint* fingerprint)
     return mFingerprints.nodebyfingerprint(fingerprint);
 }
 
+Node* MegaClient::nodebyfingerprint(LocalNode* local_node)
```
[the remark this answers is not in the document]
dwmega: Sure thing. Thank you for clarifying the convention.

src/megaclient.cpp
```diff
@@ -14006,6 +14006,67 @@ Node* MegaClient::nodebyfingerprint(FileFingerprint* fingerprint)
...
+{
+    const node_vector *remote_nodes =
```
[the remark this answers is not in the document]
dwmega: Nice catch. Will fix.

src/megaclient.cpp
dwmega: I agree with you.

The original commit searched for a remote node by calling nodebyfingerprint(...) rather than nodesbyfingerprint(...). If the remote node was of the same type as the local node, it would compare them by computing the local node's metamac using the remote node's key. If they were a match, that node would be returned. If they were not a match or if the metamac could not be computed, nullptr would be returned.

I altered the code (force pushed) so that we considered all remote nodes matching a given fingerprint. If we found a remote node with the same filename and node type, we would skip the metamac comparison. This change was made to be true to the original discussion regarding this issue.

I will add a link to the discussion here shortly.

I personally feel that the original implementation was more sound. That is, look for any node with a matching fingerprint then compare metamacs.

What are your thoughts?

src/megaclient.cpp
```diff
@@ -14006,6 +14006,67 @@ Node* MegaClient::nodebyfingerprint(FileFingerprint* fingerprint)
...
+                              && local_name == remote_node->displayname();
+                   });
+
+    if (remote_node != remote_nodes->end())
+        return *remote_node;
+
+    remote_node = remote_nodes->begin();
+
+    // Metamac only has meaning for files.
+    if ((*remote_node)->type != FILENODE)
```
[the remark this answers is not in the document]
dwmega: The find_if(...) is searching for remote node with the same node type and file name. If such a node is found, it returns that node. This is per the original discussion of the issue.

We set the "remote_node" iterator to "some random candidate." After discussion with Matt, it was decided that "the first element of the vector" was random enough. The local node will be compared against this candidate.

The main reason I am checking node type here, is for safety. I was concerned that someone could call this function passing in a LocalNode describing a directory. That or a remote node describing a directory would somehow match. If this is not a concern, I could replace the type check with an assertion?

src/utils.cpp
```diff
@@ -2278,4 +2298,50 @@ bool operator==(const SyncConfig& lhs, const SyncConfig& rhs)
     return lhs.tie() == rhs.tie();
 }
 
+std::pair<bool, int64_t> generate_metamac(SymmCipher &cipher, FileAccess &ifaccess, const int64_t iv)
```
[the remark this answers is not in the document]
dwmega: Will fix.

src/megaclient.cpp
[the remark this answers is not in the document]
sergiohs84: Actually, that's a good reason, David. I'm also concerned about calling this function with a folder's LocalNode or alike, so let's keep your check as a defense :)
Thanks

src/megaclient.cpp
sergiohs84: :thinking: I think that skipping the filename check makes sense. Imagine I have an incremental backup system and the deltas are synced up to MEGA. It's likely that a large file is only updated at some bytes, but keeping the same name and size (and CRC). Hence, the fingerprint and filename would be the same, but not the data.
Given that we are trying to patch cases like the aforementioned, I would go for the fingerprint check (plus type, but not filename). Perhaps it's not a bad idea, however, to check other matches-by-fingerprint in case the first remote node has a different type.

dwmega: I will change the code such that:
1. We search only for a remote node with a matching fingerprint and node type.
2. If we are unable to find such a node, we return nullptr.
3. If we find such a node, we perform the metamac comparison.
4. If the metamac does not much or if it could not be computed, return nullptr.
5. Otherwise, return the matched remote node.

PR #1978, opened 2020-05-05
Fix: Don't rely on undefined behavior.

dwmega:
Sure, we reserved storage space but that doesn't actually mean the
storage is allocated.


dwmega:
trigger compilation

PR #1979, opened 2020-05-05
15571: do not fireOnRequestUpdate for fetch nodes when the pendingcs is not a fetch nodes

polmr:
This fixes issues with progress being reset before finalizing fetchnodes

### DIFF
--- a/src/megaapi_impl.cpp
+++ b/src/megaapi_impl.cpp
@@ -13834,6 +13834,10 @@ void MegaApiImpl::request_error(error e)
 
 void MegaApiImpl::request_response_progress(m_off_t currentProgress, m_off_t totalProgress)
 {
+    if (!client->isFetchingNodesPendingCS())
+    {
+        return;
+    }
     for (std::map<int,MegaRequestPrivate*>::iterator it = requestMap.begin(); it != requestMap.end(); it++)
     {
         MegaRequestPrivate *request = it->second;
--- a/src/request.cpp
+++ b/src/request.cpp
@@ -26,9 +26,19 @@
 
 namespace mega {
 
+bool Request::includesFetchingNodes() const
+{
+    return mIncludesFetchingNodes;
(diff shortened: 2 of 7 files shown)


src/megaapi_impl.cpp
```diff
@@ -13834,6 +13834,10 @@ void MegaApiImpl::request_error(error e)
 
 void MegaApiImpl::request_response_progress(m_off_t currentProgress, m_off_t totalProgress)
 {
+    if (!client->isFetchingNodesPendingCS())
```
sergiohs84, line 13837: Maybe we can simply use here `MegaClient::fetchingnodes` directly??

src/request.cpp
```diff
@@ -26,9 +26,19 @@
 
 namespace mega {
 
+bool Request::includesFetchingNodes() const
+{
+    return mIncludesFetchingNodes;
```
sergiohs84, line 31: I think we could avoid this flag by using a dynamic cast. Something like:
```suggestion
    return cmds.size() == 1 && dynamic_cast<CommandFetchNodes>(cmds.back());
```

src/megaapi_impl.cpp
polmr: > since the flag might be true while the response to other commands is processed
unfortunately, it is true indeed

polmr: > MegaClient::exec() has access to the list of commands whose HTTP request's response is being processed.
I might be wrong, but I understand that for that we'd need to access RequestsDispatcher::inflightreq. That's why I saved the flag in pendingcs. Controling the flag before or inside `request_response_progess` won't make much difference I believe. (besides megacli no longer receiving it for other comands)

src/request.cpp
polmr: nice one. updaing accordingly

PR #1982, opened 2020-05-06
#15580. Add new method to allow set a DND with a timestamps for global chats

jgandres:


src/megaapi_impl.cpp
```diff
@@ -32153,6 +32159,17 @@ void MegaPushNotificationSettingsPrivate::setChatDnd(MegaHandle chatid, int64_t
     mChatDND[chatid] = timestamp;
 }
 
+void MegaPushNotificationSettingsPrivate::setChatsDnd(int64_t timestamp)
+{
+    assert(timestamp > 0);
+    if (!isGlobalEnabled())
+    {
+        LOG_warn << "setChatsDnd(): global chats notifications are currently disabled."
+                    " Setting a new time period for DND mode";
```
sergiohs84, line 32168:
```suggestion
                    " Setting a new time period for chats DND mode";
```

include/megaapi.h
```diff
@@ -4163,6 +4163,12 @@ class MegaPushNotificationSettings
      */
     virtual bool isGlobalDndEnabled() const;
 
+    /**
+     * @brief Returns whether Do-Not-Disturb mode for chats is enabled or not
```
jgandres, line 4167:
```suggestion
     * @brief Returns whether Do-Not-Disturb mode for chats is enabled or not
     
```

src/megaapi_impl.cpp
```diff
@@ -31991,11 +31991,21 @@ bool MegaPushNotificationSettingsPrivate::isGlobalDndEnabled() const
     return (mGlobalDND == 0 || mGlobalDND > m_time(NULL));
 }
 
+bool MegaPushNotificationSettingsPrivate::isGlobalChatsDndEnabled() const
```
sergiohs84, line 31994: Isn't this new method the same than `!isChatsEnabled()`?? 
I agree the name `isGlobalChatsDndEnabled()` is more homogeneous, but then... do we need to keep the `isChatsEnabled()`?? multiple interfaces to do the same thing with different code is something to avoid, usually.

include/megaapi.h
```diff
@@ -4173,8 +4176,8 @@ class MegaPushNotificationSettings
     /**
      * @brief Returns the timestamp until the DND mode is enabled
      *
-     * This method returns a valid value only if MegaPushNotificationSettings::isGlobalEnabled
-     * returns false and MegaPushNotificationSettings::isGlobalDndEnabled returns true.
+     * This method returns a valid value only if MegaPushNotificationSettings::isGlobalDndEnabled
+     * returns true and MegaPushNotificationSettings::isGlobalDndEnabled returns true.
```
sergiohs84, line 4180: Please, read the full comment and ammend

include/megaapi.h
```diff
@@ -4243,8 +4249,8 @@ class MegaPushNotificationSettings
     /**
      * @brief Returns the timestamp until the Do-Not-Disturb mode for a chat
      *
-     * This method returns a valid value only if MegaPushNotificationSettings::isChatEnabled
-     * returns false and MegaPushNotificationSettings::isChatDndEnabled returns true.
+     * This method returns a valid value only if MegaPushNotificationSettings::isChatDndEnabled
+     * returns true and MegaPushNotificationSettings::isChatDndEnabled returns true.
```
sergiohs84, line 4253: Please, read the comment in full and amend

include/megaapi.h
```diff
@@ -4279,14 +4285,17 @@ class MegaPushNotificationSettings
...
      */
     virtual bool isChatsEnabled() const;
 
     /**
      * @brief Returns the timestamp until the chats DND mode is enabled
      *
-     * This method returns a valid value only if MegaPushNotificationSettings::isChatsEnabled
-     * returns false and MegaPushNotificationSettings::isGlobalChatsDndEnabled returns true.
+     * This method returns a valid value only if MegaPushNotificationSettings::isGlobalChatsDndEnabled
+     * returns true and MegaPushNotificationSettings::isGlobalChatsDndEnabled returns true.
```
sergiohs84, line 4298: Please, read the full comment and amend

include/megaapi_impl.h
```diff
@@ -882,6 +884,7 @@ class MegaPushNotificationSettingsPrivate : public MegaPushNotificationSettings
 
     void enableChat(MegaHandle chatid, bool enable) override;
     void setChatDnd(MegaHandle chatid, int64_t timestamp) override;
+    void setChatsDnd(int64_t timestamp) override;
```
sergiohs84, line 887: Shouldn't it be `setGlobalChatsDnd()` to be in line with the getter and checker (`isGlo...()`) methods?

src/megaapi_impl.cpp
```diff
@@ -31928,7 +31928,7 @@ string MegaPushNotificationSettingsPrivate::generateJson() const
         json.append(",");
     }
 
-    if (mGlobalChatsDND > -1)
+    if (isGlobalChatsDndEnabled())
```
sergiohs84, line 31931: Not sure if it's done in purpose, but previously it used to test whether the global DND for chats was unset. Now, it checks it's effective (note the DND's ts can already be in the past and this method will return false, where the previous method will test successfully (true).
Maybe it's better this way, since the DND would be removed. But take into account it changes the existing logic, and you may want to keep it untouched. If you are unsure, @jgandres , please, check with Alber. Thanks

jgandres: We have been talking about this topic, and we don't see any case we need to preserve an expired DND, If you know any case please tell me in order to make the adjustment.

sergiohs84: I don't think so either. So... all fine ;)

PR #1985, opened 2020-05-07
#15582: do remove transfer and files on tranfer failed.

polmr:
To prevent "Async operation failed" triggering a deletion a fail that didn't cause a deletion and an ulterior crash.
This reverts to the unconditional deletion of the failed transfer before  https://github.com/meganz/sdk/pull/1844: https://github.com/meganz/sdk/pull/1844/files#diff-7dfecf951cd004ec60662f276863edcdL529 
(now the only case when that transfer doesn't get to deletion is if it as a file being upload to a foreign account, which was the original intent in 1844)

### DIFF
--- a/src/transfer.cpp
+++ b/src/transfer.cpp
@@ -517,7 +517,7 @@ void Transfer::failed(error e, DBTableTransactionCommitter& committer, dstime ti
     }
     else
     {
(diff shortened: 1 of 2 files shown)


sergiohs84:
Closing this PR, since the the code that resulted in this bug has been reverted.

PR #1988, opened 2020-05-08
#15003. Handle custom ETD errors

alber2510:
All functionality isn't finished but but allow connect API version v2 and receive ETD erros.
Remains to be done the handle of ETD errors at upper layer


include/mega/command.h
```diff
@@ -92,9 +92,19 @@ class MEGA_API Command
     Command();
     virtual ~Command() = default;
 
+    bool static checkError(int64_t &e, JSON &json, ErrorDetails &errorDetails);
```
sergiohs84, line 95:
```suggestion
    static bool checkError(int64_t &e, JSON &json, ErrorDetails &errorDetails);
```

sergiohs84, line 95: What does this method return? ie:
```cpp
// return true when...```

File: src/command.cpp [hunk 2]
```diff
@@ -48,6 +48,57 @@ const char* Command::getstring() const
 }
 
 // add opcode
+bool Command::checkError(int64_t& e, JSON& json, ErrorDetails& errorDetails)
```
sergiohs84, line 51: The comment above is for the next method

src/command.cpp
```diff
@@ -48,6 +48,57 @@ const char* Command::getstring() const
...
+                    case 'u':
+                        errorDetails.u = json.getint();
+                        break;
+                    case 'l':
+                        errorDetails.l  = json.getint();
+                        break;
+                    case EOO:
+                        json.leaveobject();
+                        return true;
+                        break;
```
sergiohs84, line 89: This break is a leftover, I guess

src/command.cpp
sergiohs84, line 51: Maybe `e` --> `errorCode` will be more meaningful

include/mega/command.h
```diff
@@ -92,9 +92,19 @@ class MEGA_API Command
...
+
     MEGA_DEFAULT_COPY_MOVE(Command)
 };
 
+class ErrorDetails
+{
+public:
+    bool exits = false;
+    int64_t u = 0;
```
sergiohs84, line 104: What is `u`?

include/mega/command.h
```diff
@@ -92,9 +92,19 @@ class MEGA_API Command
...
+    int64_t l = 0;
```
sergiohs84, line 105: What is `l`?

include/mega/command.h
```diff
@@ -92,9 +92,19 @@ class MEGA_API Command
     Command();
     virtual ~Command() = default;
 
+    bool static checkError(int64_t &e, JSON &json, ErrorDetails &errorDetails);
+
     MEGA_DEFAULT_COPY_MOVE(Command)
 };
 
+class ErrorDetails
```
sergiohs84, line 100: Maybe we could include the error code inside this class, which could represent the `Error` :thinking: 
Just thinking loud ^

src/command.cpp
alber2510: Fixed

alber2510: I've used `int64_t` because some commands return a integer value instead of an error. Then I think it's better return a generic type and after that convert to type

include/mega/command.h
alber2510: In method `bool Command::checkError(int64_t& e, JSON& json, ErrorDetails& errorDetails
Parameter `e` is not always an error. If we want to reduce the number of parameter it's possible to add in `ErrorDetails` but maybe we have to rename the class

include/mega/command.h
alber2510: Solved

include/mega/command.h
alber2510: Solved

src/command.cpp
alber2510: Solved

include/mega/command.h
alber2510: Solved

src/commands.cpp
```diff
@@ -979,11 +971,11 @@ CommandSetAttr::CommandSetAttr(MegaClient* client, Node* n, SymmCipher* cipher,
...
 void CommandSetAttr::procresult()
 {
-    if (client->json.isnumeric())
+    Error errorDetails;
+    if (checkError(errorDetails, client->json))
     {
-        error e = (error)client->json.getint();
 #ifdef ENABLE_SYNC
-        if(!e && syncop)
+        if(errorDetails == API_OK && syncop)
```
sergiohs84, line 978: Isn't it the same?

src/commands.cpp
```diff
@@ -1301,12 +1289,12 @@ void CommandPutNodes::procresult()
             }
         }
 #endif
-        client->app->putnodes_result((!e && empty) ? API_ENOENT : e, type, nn);
+        client->app->putnodes_result((!errorDetails && empty) ? API_ENOENT : errorDetails.getError(), type, nn);
```
sergiohs84, line 1292: Here you can simply use a casting to `error`, and avoid the extra getter ;)

src/commands.cpp
```diff
@@ -5715,15 +5737,28 @@ void CommandQueryRecoveryLink::procresult()
...
+    int64_t type;
+    Error errorDetails;
+
+    if (checkError(errorDetails, client->json) && errorDetails < 0)
     {
-        return client->app->queryrecoverylink_result((error)type);
+        return client->app->queryrecoverylink_result(errorDetails);
+    }
+
+    type = static_cast<uint64_t>(errorDetails);
```
sergiohs84, line 5753: Why does the data type of `type` changes from `int` to `int64_t`??
And why the casting to `uint64_t`??
I think we can simply keep the `int`

src/commands.cpp
```diff
@@ -7358,24 +7407,32 @@ CommandMediaCodecs::CommandMediaCodecs(MegaClient* c, Callback cb)
 }
 
 void CommandMediaCodecs::procresult()
-{
-    int version = 0;
-    if (client->json.isnumeric())
+{    
+    Error errorDetails;
+    int64_t result;
```
sergiohs84, line 7412: I think we can simply keep the `int version`. Is there any specific reason for changing the data type?

examples/megacli.cpp
```diff
@@ -7905,11 +7963,15 @@ void DemoAppFolder::login_result(error e)
     }
 }
 
-void DemoAppFolder::fetchnodes_result(error e)
+void DemoAppFolder::fetchnodes_result(const Error& e)
 {
     if (e)
     {
-        cout << "File/folder retrieval failed (" << errorstring(e) << ")" << endl;
```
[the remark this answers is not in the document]
alber2510: Yes, I missing it

examples/megacli.cpp
```diff
@@ -6964,17 +7007,25 @@ void DemoApp::exportnode_result(handle h, handle ph)
 }
 
 // the requested link could not be opened
-void DemoApp::openfilelink_result(error e)
+void DemoApp::openfilelink_result(const Error& e)
 {
-    if (e)
+    if (e != API_OK)
```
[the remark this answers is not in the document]
alber2510: It's not necessary, for me it's more fast to read `e != API_OK`. Probably I changed it without notice

include/megaapi.h
```diff
@@ -5325,6 +5325,15 @@ class MegaError
          */
         MegaError* copy();
 
+        /**
+         * @brief Add extra information for an error
+         *
+         * @param userStatus Extra information error about user
+         * @param linkStatus Extra information error about link
+         */
+        void setExtraErrorInfo(long long userStatus, long long linkStatus);
```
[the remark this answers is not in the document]
alber2510: Probably we have to do, in future probably we have to add more extra info error

src/megaapi_impl.cpp
```diff
@@ -11823,7 +11823,13 @@ dstime MegaApiImpl::pread_failure(error e, int retry, void* param, dstime timeLe
...
         }
         DBTableTransactionCommitter committer(client->tctable);
-        fireOnTransferFinish(transfer, MegaError(e), committer);
+        MegaError err = MegaError(e);
+        if (e == API_ETOOMANY && e.hasExtraInfo())
+        {
+            err.setExtraErrorInfo(e.getUserStatus(), e.getLinkStatus());
+        }
+
+        fireOnTransferFinish(transfer, err, committer);
```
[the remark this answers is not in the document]
alber2510: This would  be apply when we createa MegaErrorPrivate?

src/commands.cpp
alber2510: Yes, it's good

src/commands.cpp
alber2510: None specific, only define the same type that is returned by `json.getint()`

src/commands.cpp
```diff
@@ -62,20 +62,18 @@ HttpReqCommandPutFA::HttpReqCommandPutFA(MegaClient* client, handle cth, fatype
 
 void HttpReqCommandPutFA::procresult()
 {
-    error e;
     client->looprequested = true;
 
-    if (client->json.isnumeric())
+    Error errorDetails;
```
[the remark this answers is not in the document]
alber2510: It's good for me

src/commands.cpp
alber2510: Yes, it's the same

src/commands.cpp
alber2510: Because it can be negative, if it will be negative, it would be considerate as an error. But it's the same

include/mega/types.h
```diff
@@ -209,6 +209,43 @@ typedef enum ErrorCodes
...
+    void setError(error err)
+    {
+        mError = err;
+        mExtraErrorInfo = false;
+    }
+
+    void setErrorWithExtraInfo(error err)
+    {
+        mError = err;
+        mExtraErrorInfo = true;
```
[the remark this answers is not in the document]
alber2510: If `err` appear in json, extra info should be provided. We avoid to assign value twice to `mExtraErrorInfo = true;`. But this is not very important

include/mega/command.h
alber2510: This has been changed

include/mega/types.h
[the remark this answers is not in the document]
sergiohs84: Documentation says:
> 2) If yes, is the extra information available? (It should be assumed that there may be cases it is not provided, just in case).
It's not clear whether the error without extra info will be received with the old fashion or the new format, but just in case I think it's more correct to assume there's extra info only when there's extra info, since the new format might become the standard in the future for reporting any error for any format ;)

src/command.cpp
```diff
@@ -47,6 +47,53 @@ const char* Command::getstring() const
...
 
+//return true when the response is an error, false otherwise (in that case it doesn't consume JSON chars)
+bool Command::checkError(Error& errorDetails, JSON& json)
+{
+    error e;
+    if (json.isNumericError(e))
+    {
+        errorDetails.setErrorCode(e);
+        return true;
+    }
```
mattw-mega, line 58: I see isNumericError was added rather than using the already isnumeric() and getint().   The difference seems to be whether numbers above 0 would be included.   The old code for commands I've checked would just store the positive number in the Error and continue with that, I think we could do the same here without adding a new function in the JSON class that has overlapping functionality with existing functions - unless there are commands where we do return just a single positive number on success? (but that would be an error prone design)  thanks

src/json.cpp
```diff
@@ -584,6 +584,37 @@ bool JSON::extractstringvalue(const string &json, const string &name, string *va
     return true;
 }
 
+bool JSON::isNumericError(error &e)
+{
+    const char* ptr = pos;
+    if (*ptr == ',')
+    {
+        ptr++;
+    }
```
mattw-mega, line 593: as above, this function is almost the same as isnumeric() and getint(), probably we can just use those in the same way the commands did prevously rather than adding overlapping functionality?

src/megaapi.cpp
```diff
@@ -1221,47 +1226,49 @@ long long MegaTransfer::getNotificationNumber() const
...
-    this->value = value;
+
 }
 
-MegaError::MegaError(const MegaError &megaError)
+MegaError* MegaError::copy() const
 {
-	errorCode = megaError.getErrorCode();
-    value = megaError.getValue();
+    return nullptr;
```
mattw-mega, line 1241: Since it as allowed to have a MegaError that is not actually a MegaErrorPrivate perhaps we should return a copy here?

mattw-mega: from MegaTransfer::getLastError

src/megaapi.cpp
```diff
@@ -1221,47 +1226,49 @@ long long MegaTransfer::getNotificationNumber() const
...
 }
 
-MegaError::~MegaError()
-{
+int MegaError::getErrorCode() const 
+{ 
+    return 0;
```
mattw-mega, line 1246: probably we should return errorcode here, for  MegaTransfer::getLastError

src/megaapi.cpp
```diff
@@ -1370,17 +1377,17 @@ const char* MegaError::getErrorString(int errorCode, ErrorContexts context)
 
 const char* MegaError::toString() const 
 { 
-	return getErrorString(); 
+    return nullptr;
```
mattw-mega, line 1380: probably we should cover the to string cases for  MetaTransfer::getLastError also

src/megaapi_impl.cpp
```diff
@@ -2773,9 +2778,12 @@ void MegaTransferPrivate::setLastBytes(char *lastBytes)
     this->lastBytes = lastBytes;
 }
 
-void MegaTransferPrivate::setLastError(MegaError e)
+void MegaTransferPrivate::setLastError(const MegaError *e)
 {
-    this->lastError = e;
+    if (lastError)
+        delete lastError;
```
mattw-mega, line 2784: using unique_ptr would avoid the need for this code

src/megaapi_impl.cpp
```diff
@@ -2773,9 +2778,12 @@ void MegaTransferPrivate::setLastBytes(char *lastBytes)
...
+
+    lastError = e ? static_cast<MegaErrorPrivate*>(e->copy()) : nullptr;
```
mattw-mega, line 2786: static_cast is ok for going up an inheritance heirarchy tree, but for going deeper into the tree we should use dynamic_cast, in case the object passed in is not of that type - eg. just a plain MegaError, not the private version.

src/megaapi_impl.cpp
```diff
@@ -5176,7 +5185,7 @@ MegaApiImpl::~MegaApiImpl()
     delete httpio;
 #endif
 
-    fireOnRequestFinish(request, MegaError(API_OK));
+    fireOnRequestFinish(request, new MegaErrorPrivate(API_OK));
```
mattw-mega, line 5188: as noted above, the parameter here could be mega::make_unique<MegaErrorPrivate>(API_OK) to indicate ownership and have the compiler enforce it.

src/megaapi_impl.cpp
```diff
@@ -9290,27 +9299,30 @@ void MegaApiImpl::fireOnStreamingStart(MegaTransferPrivate *transfer)
...
 }
 
-void MegaApiImpl::fireOnStreamingTemporaryError(MegaTransferPrivate *transfer, MegaError e)
+void MegaApiImpl::fireOnStreamingTemporaryError(MegaTransferPrivate *transfer, MegaErrorPrivate *e)
 {
     for(set<MegaTransferListener *>::iterator it = httpServerListeners.begin(); it != httpServerListeners.end() ; it++)
-        (*it)->onTransferTemporaryError(api, transfer, &e);
+        (*it)->onTransferTemporaryError(api, transfer, e);
+
+    delete e;
```
mattw-mega, line 9307: as noted above, the parameter here can be `std::unique_ptr<MegaErrorPrivate> e`
and then ownership is tracked, human readers and compilers know what is going on, and we don't need the manual call to `delete`    (the onTransferTemporaryError would use `e.get()` )

src/command.cpp
alber2510: Yes, I think so, there are several commands that return a single positive number on success:
CommandMediaCodecs, CommandCreditCardQuerySubscriptions, CommandGetPaymentMethods (this one, can return several positive number)
In `checkError` we have tried to handle only negative plus 0 numbers. We thought that it is not very correct semantically, return a positive number in a function `checkError`

include/mega/types.h
```diff
@@ -209,6 +209,33 @@ typedef enum ErrorCodes
...
+    bool hasExtraInfo() const { return mExtraErrorInfo; }
+    int64_t getUserStatus() const { return mUserStatus; }
+    int64_t getLinkStatus() const { return mLinkStatus; }
+    operator error() const { return mError; }
+
+private:
+    error mError = API_EINTERNAL;
+    bool  mExtraErrorInfo = false;
+    int64_t mUserStatus = 0; // user status
+    int64_t mLinkStatus = 0; // link status
```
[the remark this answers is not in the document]
sergiohs84: Agree about the enumeration.
As for the need or the `mExtraErrorInfo`, we can easily avoid it by initializing the status variables to -1

include/megaapi_impl.h
```diff
@@ -127,6 +127,32 @@ class MegaDbAccess : public SqliteDbAccess
...
+    long long getUserStatus() const override;
+    long long getLinkStatus() const override;
+    const char* getErrorString() const override;
+    const char* toString() const override;
+    const char* __str__() const override;
+    const char* __toString() const override;
+
+private:
+    long long mValue = 0;
+    bool mExtraInfo = false;
```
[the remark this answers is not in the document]
sergiohs84: Agree

include/megaapi_impl.h
```diff
@@ -747,7 +774,7 @@ class MegaTransferPrivate : public MegaTransfer, public Cacheable
 
         MegaTransferListener *listener;
         Transfer *transfer;
-        MegaError lastError;
+        MegaErrorPrivate *lastError = nullptr;
```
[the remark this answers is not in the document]
sergiohs84: Agree

include/megaapi_impl.h
```diff
@@ -2474,8 +2501,8 @@ class MegaApiImpl : public MegaApp
         void httpServerRemoveListener(MegaTransferListener *listener);
 
         void fireOnStreamingStart(MegaTransferPrivate *transfer);
-        void fireOnStreamingTemporaryError(MegaTransferPrivate *transfer, MegaError e);
-        void fireOnStreamingFinish(MegaTransferPrivate *transfer, MegaError e);
+        void fireOnStreamingTemporaryError(MegaTransferPrivate *transfer, MegaErrorPrivate* e);
+        void fireOnStreamingFinish(MegaTransferPrivate *transfer, MegaErrorPrivate *e);
```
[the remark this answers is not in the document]
sergiohs84: Agree

src/json.cpp
sergiohs84: There are 2 main differences between `isnumeric` and `inNumericError` is that the later won't return true for positive numbers (errors are always negative or 0, but there are commands that return a positive value on success), and also it returns the `error` by reference by consuming the chars from the JSON in case of negative number (where `isnumeric` + `getint` would have consumed the chars from the JSON without being an error).
I think the difference worth the addition of the `JSON::isNumericError()` (or simply embed the code in the new `checkError()`)

src/megaapi.cpp
sergiohs84: Agree

src/megaapi.cpp
sergiohs84: Agree

src/megaapi.cpp
sergiohs84: Agree, well spotted!!

src/megaapi_impl.cpp
sergiohs84: Agree

src/megaapi_impl.cpp
sergiohs84: Agree

src/megaapi_impl.cpp
sergiohs84: Agree

src/megaapi_impl.cpp
sergiohs84: Agree

include/mega/types.h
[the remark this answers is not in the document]
alber2510: Solved

include/megaapi_impl.h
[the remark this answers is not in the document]
alber2510: Solved

src/megaapi.cpp
alber2510: Solved

src/megaapi.cpp
alber2510: Solved

src/megaapi_impl.cpp
alber2510: Solved

src/megaapi_impl.cpp
alber2510: Solved

src/megaapi.cpp
alber2510: Solved

include/megaapi_impl.h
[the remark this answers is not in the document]
alber2510: Solved

src/megaapi_impl.cpp
alber2510: It's solved at Matt PR #2045

src/megaapi_impl.cpp
alber2510: It's solved at Matt PR #2045

src/command.cpp
mattw-mega: I see, thanks for the info - of course if those commands already do that then we have to support it.   By the way, reading CommandGetPaymentMethods, I see that 0 returned is actually data to be processed rather than an API_OK value which makes the code quite awkward but it looks correct.  I wonder if an option for checkError() to not consume 0 would help with those cases.  Anyway, I understand the reasoning, thanks for that.  cheers

PR #1993, opened 2020-05-12
#15584. Allow to search only among inshares / outshares / links

jgandres:


jgandres:
trigger compilation

jgandres:
trigger compilation

jgandres:
trigger compilation

src/node.cpp
```diff
@@ -1096,11 +1102,13 @@ void Node::setpubliclink(handle ph, m_time_t cts, m_time_t ets, bool takendown)
     }
     else            // update
     {
+        client->mPublicLinks.erase(plink->ph);
```
sergiohs84, line 1105: I think you don't need to `erase`. The new assignment a few lines below will do its work anyway. Instead, you could set a:
```suggestion
        assert(client->mPublicLinks.find(plink->ph) != client->mPublicLinks.end());
```
And for the block above to create the link:
```
        assert(client->mPublicLinks.find(ph) == client->mPublicLinks.end());
```

jgandres: I added the erase because in else we update the current plink->ph with ph received as param.
**plink->ph = ph;** so if ph has changed we need to remove current map entry and add new one.

However I've testing and I'm not sure ph can change so in this case this line **plink->ph = ph;** is not needed.

src/megaapi_impl.cpp
```diff
@@ -10621,6 +10628,68 @@ bool MegaApiImpl::processMegaTree(MegaNode* n, MegaTreeProcessor* processor, boo
...
+           vector<Node *>& vNodes  = searchProcessor.getResults();
+           result.insert(result.end(), vNodes.begin(), vNodes.end());
+        }
+    }
+    else
+    {
+        // Search in public links
+        ::mega::unique_ptr<MegaNodeList> publicLinks(getPublicLinks(MegaApi::ORDER_NONE));
+
+        for (int i = 0; i < publicLinks->size() && !(cancelToken && cancelToken->isCancelled()); i++)
```
sergiohs84, line 10679: Shouldn't we directly iterate through `mPublicLinks` instead of create a `MegaNodeList` for the shake of retrieving the nodehandles relative to the links?

include/mega/megaclient.h
```diff
@@ -1256,6 +1256,9 @@ class MEGA_API MegaClient
     // maps local fsid to corresponding LocalNode*
     handlelocalnode_map fsidnode;
 
+    // maps public handle to node handle
+    std::map<handle, handle> mPublicLinks;
```
sergiohs84, line 1260: I think the right mapping is `nodehandle` to `publichandle`, and not viceversa.

src/node.cpp
sergiohs84: I see. I assumed at some point in my head that the key of the map would be the nodehandle. I think we should adjust the map. It also helps here, since the item in the map only needs to update the value, rather than remove the item to be inserted again.

jgandres:
trigger compilation

src/megaapi_impl.cpp
```diff
@@ -10674,11 +10671,9 @@ MegaNodeList* MegaApiImpl::searchInAllShares(const char *searchString, MegaCance
     else
     {
         // Search in public links
-        ::mega::unique_ptr<MegaNodeList> publicLinks(getPublicLinks(MegaApi::ORDER_NONE));
-
-        for (int i = 0; i < publicLinks->size() && !(cancelToken && cancelToken->isCancelled()); i++)
+        for (const auto& item : client->mPublicLinks)
```
sergiohs84, line 10674: So we removed the feature to cancel the ongoing search? You need to keep that condition!

jgandres: Bad copy paste so sorry.

jgandres:
Tasks 2291, 2285 2283, failed for a non related test that is already reported.

PR #2003, opened 2020-05-19
#15752. Performance optimizations for working with large numbers of transfers

mattw-mega:
Cancelling transfers in particular, and shutdown were very slow with > 100k transfers
This PR addresses some N^2 behaviour
 - When deleting all nodes, no need for every node to remove itself from the Fingerprints map, just clear() it.  (optimize purge nodes flag)
 - When destroying all transfers with freeq(), no need for the transfers to remove themselves from the list, just clear() it.  (optimize delete flag)
 - When removing transfers from the std::deque due to transfer completion, the removal requires moving all the remaining items back one place.
   If we are removing many items (such as "cancel all" with 100k transfers queued, doing all that shuffling per transfer is ridiculously slow.
   A deque wrapper class is introduced that intercepts erase() and marks the entry erased instead (with its Transfer* set to null as the transfer has now been deleted).
   All other function calls on the deque are also intercepted, and first actually erase those elements (performing the shuffling once only) and then perform that function as usual.
   There was some complication with the 'priority' value which is still needed to be able to find other elements to erase, so LazyEraseTransferPtr was introduced to keep that value even when the entry is marked erased.
 - More efficient getChildren function that just uses sort() instead of manually building up a sorted vector with lower_bound.
 - Avoid sorting getChildren anyway in MegaFolderDownloadController::downloadFolderNode, as that takes significant time for very large folders.

### DIFF
--- a/include/mega/types.h
+++ b/include/mega/types.h
@@ -314,9 +314,14 @@ typedef map<FileFingerprint*, Transfer*, FileFingerprintCmp> transfer_map;
 template <class T, class E>
 class deque_with_lazy_bulk_erase
 {
+    // This is a wrapper class for deque.  Erasing an element from the middle of a deque is not cheap since all the subsequent elements need to be shuffled back.
+    // This wrapper intercepts the erase() calls for single items, and instead marks each one as 'erased'.  
+    // The supplied template class E contains the normal deque entry T, plus a flag or similar to mark an entry erased.
+    // Any other operation on the deque performs all the gathered erases in a single std::remove_if for efficiency.
+    // This makes an enormous difference when cancelling 100k transfers in MEGAsync's transfers window for example.
     deque<E> mDeque;
+    bool mErasing = false;
 
-    bool erasing = false;
 public:
 
     typedef typename deque<E>::iterator iterator;
@@ -325,26 +330,26 @@ public:
     {
         assert(i != mDeque.end());
         i->erase();
-        erasing = true;
+        mErasing = true;
     }
 
     void applyErase()
     {
-        if (erasing)
+        if (mErasing)
         {
             auto newEnd = std::remove_if(mDeque.begin(), mDeque.end(), [](const E& e) { return e.isErased(); } );
             mDeque.erase(newEnd, mDeque.end());
-            erasing = false;
+            mErasing = false;
         }
     }
 
-    size_t size() { applyErase(); return mDeque.size(); }
+    size_t size()                                        { applyErase(); return mDeque.size(); }
     iterator begin(bool canHandleErasedElements = false) { if (!canHandleErasedElements) applyErase(); return mDeque.begin(); }
-    iterator end(bool canHandleErasedElements = false) { if (!canHandleErasedElements) applyErase(); return mDeque.end(); }
-    void push_front(T t) { applyErase(); mDeque.push_front(E(t)); }
-    void push_back(T t) { applyErase(); mDeque.push_back(E(t)); }
-    void insert(iterator i, T t) { applyErase(); mDeque.insert(i, E(t)); }
-    T& operator[](size_t n) { applyErase(); return mDeque[n]; }
+    iterator end(bool canHandleErasedElements = false)   { if (!canHandleErasedElements) applyErase(); return mDeque.end(); }
+    void push_front(T t)                                 { applyErase(); mDeque.push_front(E(t)); }
+    void push_back(T t)                                  { applyErase(); mDeque.push_back(E(t)); }
+    void insert(iterator i, T t)                         { applyErase(); mDeque.insert(i, E(t)); }
+    T& operator[](size_t n)                              { applyErase(); return mDeque[n]; }
 
 };
 
--- a/include/mega/types.h
+++ b/include/mega/types.h
@@ -311,7 +311,42 @@ typedef list<HttpReqCommandPutFA*> putfa_list;
 // map a FileFingerprint to the transfer for that FileFingerprint
 typedef map<FileFingerprint*, Transfer*, FileFingerprintCmp> transfer_map;
 
-typedef deque<Transfer*> transfer_list;
+template <class T, class E>
+class deque_with_lazy_bulk_erase
+{
+    deque<E> mDeque;
+
+    bool erasing = false;
+public:
+
+    typedef typename deque<E>::iterator iterator;
+
+    void erase(iterator i)
+    {
+        assert(i != mDeque.end());
+        i->erase();
+        erasing = true;
+    }
+
+    void applyErase()
+    {
+        if (erasing)
+        {
+            auto newEnd = std::remove_if(mDeque.begin(), mDeque.end(), [](const E& e) { return e.isErased(); } );
+            mDeque.erase(newEnd, mDeque.end());
+            erasing = false;
+        }
+    }
+
+    size_t size() { applyErase(); return mDeque.size(); }
+    iterator begin(bool canHandleErasedElements = false) { if (!canHandleErasedElements) applyErase(); return mDeque.begin(); }
+    iterator end(bool canHandleErasedElements = false) { if (!canHandleErasedElements) applyErase(); return mDeque.end(); }
+    void push_front(T t) { applyErase(); mDeque.push_front(E(t)); }
+    void push_back(T t) { applyErase(); mDeque.push_back(E(t)); }
+    void insert(iterator i, T t) { applyErase(); mDeque.insert(i, E(t)); }
+    T& operator[](size_t n) { applyErase(); return mDeque[n]; }
+
+};
 
 // map a request tag with pending dbids of transfers and files
 typedef map<int, vector<uint32_t> > pendingdbid_map;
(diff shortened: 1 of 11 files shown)


include/mega/types.h
```diff
@@ -311,7 +311,42 @@ typedef list<HttpReqCommandPutFA*> putfa_list;
 // map a FileFingerprint to the transfer for that FileFingerprint
 typedef map<FileFingerprint*, Transfer*, FileFingerprintCmp> transfer_map;
 
-typedef deque<Transfer*> transfer_list;
+template <class T, class E>
+class deque_with_lazy_bulk_erase
```
sergiohs84, line 315: `types.h` is included in about other 25 header files. I think we should try to help the compiler a bit and move the definitions to the `utils.cpp` as other objects declared here (when they don't have a better target source file).

src/megaapi_impl.cpp
```diff
@@ -24907,7 +24889,7 @@ void MegaFolderDownloadController::downloadFolderNode(MegaNode *node, string *pa
     }
     else
     {
-        children = megaApi->getChildren(node);
+        children = megaApi->getChildren(node, MegaApi::ORDER_NONE);  // no order is much faster for a very large folder (or nested folders with large subfolders)
```
sergiohs84, line 24892: Good catch!!
Maybe we should set `ORDER_NONE` as default value? I can see other usages where the default order is used, but I hesitate it's in purpose, like in:
- `MegaApiImpl::getBackupFolders()`
- `MegaApiImpl::authorizeMegaNodePrivate()`
- `MegaBackupController::removeexceeding()`
All of them can benefit of the `ORDER_NONE`, since the order is irrelevant for them (but maybe not for apps calling `MegaApi::getChildren()`

src/megaapi_impl.cpp
```diff
@@ -16938,42 +16938,24 @@ MegaNodeList *MegaApiImpl::getChildren(MegaNode* p, int order)
...
-    {
-        result = new MegaNodeListPrivate(childrenNodes.data(), int(childrenNodes.size()));
-    }
-    else
-    {
-        result = new MegaNodeListPrivate();
-    }
-    sdkMutex.unlock();
-    return result;
+    return new MegaNodeListPrivate(childrenNodes.data(), int(childrenNodes.size()));;
```
sergiohs84, line 16958: We could take this chance to apply the same improvements to the other analog methods:
- `MegaApiImpl::getFileFolderChildren()`
- `MegaApiImpl::getIndex()`
The other public methods that allows to specify an order for the results already use `ORDER_NONE` by default (ie. `search()` ) :)

src/node.cpp
```diff
@@ -114,7 +114,10 @@ Node::~Node()
     client->preadabort(this);
 
     // remove node's fingerprint from hash
-    client->mFingerprints.remove(this);
+    if (!client->mOptimizePurgeNodes)
+    {
+        client->mFingerprints.remove(this);
+    }
```
sergiohs84, line 120: If we don't remove the entry in `mFingerprints`, won't we be in risk of trying a remote copy for a fresh upload of a file with the same fingerprint (see `MegaClient::nodebyfingerprint()`). If so, the node has been deleted, so would need to check for side effects.

sergiohs84: :thinking: I've just checked, the `mOptimizePurgeNodes` is set only at `purgenodesusersabortsc()` and the `mFingerprints` is cleared right before deleting nodes, so it should be safe :smiley: 
However, I think a `mNodeCounters.clear()` is missing at the `purgenodesusersabortsc()` for consistency?

include/mega/types.h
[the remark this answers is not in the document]
mattw-mega: Yes I tried to keep them small for that reason (and in general).  But the main constraint is that it's a template class, so those functions have to be in the header.   I agree that types.h is growing and perhaps has too much in it now - shall we split out a separate header, perhaps helperClasses.h and put these inline classes in it?  thanks

include/mega/types.h
```diff
@@ -311,7 +311,42 @@ typedef list<HttpReqCommandPutFA*> putfa_list;
...
+{
+    deque<E> mDeque;
+
+    bool erasing = false;
```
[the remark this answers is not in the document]
mattw-mega: updated, thanks

src/megaapi_impl.cpp
mattw-mega: That sounds like a good improvement, I would like to keep it in mind and do it in a separate PR though, on develop, since this PR is heading for a release branch, as it will require some time and testing.

mattw-mega: That sounds like a good improvement, though I'm not sure if it's right for this PR which is just about speeding up the large transfer case.   I think we should keep it in mind for a future PR, and especially if we do see some performance issues in those other cases.

src/megaapi_impl.cpp
mattw-mega: I agree but I'd like to wait and see if the change from lower_bound to sort() has any unintended issues - in theory I think the old lower_bound technique would produce something similar to stable_sort, but with equal ranges reversed.    I think we should keep that in mind for future improvements once we are certain this one is good.    And getIndex is an interesting case, I'm not sure about that one since, just before it locks the mutex or just after is releases it, the child list could have changed, and then the caller may get back an incorrect or out of date index - if they use that in a subsequent call to get that item, they will usually get the right one but occasionally get something else.

src/node.cpp
mattw-mega: yes good point, it's good practice to keep all the affected items together like that.   That one is cleared later on during the init() call, but I agree it's better to keep these together so I've added that call, thanks.

include/mega/types.h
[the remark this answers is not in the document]
sergiohs84: Oh, you are correct! I didn't realize about the template class restriction, sorry. Nevertheless, it's a good idea to split out this file like that. Let me add a comment to the ticket to split out the `megaapi_impl.cpp`, so we do both at once rather than in this PR ;)

src/megaapi_impl.cpp
sergiohs84: The `getIndex()` looked weird/risky to me also, I don't see why the app would need to ask the SDK about the index of an already returned sorted list of children. Apparently, it's used only by Windows Phone 8, which is currently obsolete and is replaced by the Universal Windows Platform. Nevertheless, out of curiosity, I've checked the code and it's used right after addition of a new node. The app wants to know the index, in the list of children, to insert it in the UI in the right position.
It's subject to races, as you pointed out. I think we should tag the method as deprecated.

src/megaapi_impl.cpp
sergiohs84: I would add it to this PR. In the end, the PR aims to solve performance issues and you found out that sorting was a bottleneck on large sets. So... why not to improve all over the code now that it has been analyzed? Specially when it's just 3 lines :)

src/transfer.cpp
```diff
@@ -1889,31 +1896,31 @@ error TransferList::pause(Transfer *transfer, bool enable, DBTableTransactionCom
     return API_EFAILED;
 }
 
-transfer_list::iterator TransferList::begin(direction_t direction)
+auto TransferList::begin(direction_t direction) -> transfer_list::iterator
```
sergiohs84, line 1899: Out of curiosity, what's the motivation of this change?

mattw-mega: transfer_list is now a type defined in TransferList (as a specialization of the lazy bulk erase deque), whereas before it was  global.   In order to name it in the old way, it would have to be TransferList::transfer_list::iterator which is getting a bit long and repetitive.  Using the new syntax with the type trailing after -> means it is evaluated in the class' context rather than global context, so we can leave off the TransferList::  which is one of the aims of that syntax enhancement in C++11 (besides also bringing it in line with lambda syntax).   thanks

sergiohs84: Thank you for the explanation, Matt.

src/megaapi_impl.cpp
mattw-mega: Ok, I've gone through them and I agree they are suitable in call cases - changes made.  In fact I removed the default so it's clear what we are doing each time we call it (it's a private interface, and only a few cases used the default anyway)

src/megaapi_impl.cpp
mattw-mega: Ok, I've made the changes here - getFileFolderChildren is quite a lot smaller!   Since I was making changes there anyway, I've arranged it in such a way that the new allocations will behave nicely if an out-of-memory error is thrown, passing in unique_ptr to the list contstructors, which meant a few small changes in the list class, just using unique_ptr instead of raw. 
About getIndex, that's good news that it's not used anymore.  Let's just remove it then, and avoid it being used accidentally.   If there are any complaints we can explain the issue, and if necessary provide an alternative (maybe they pass us the list they know, and we can provide the index against that list).    I think for ourselves, it's better to remove the bad code than still have it there where it's a maintenance issue.  thanks!

PR #2009, opened 2020-05-21
#13885. Fix for crypto on threads in the Transfer destructor

mattw-mega:
When a Transfer is destroyed, we should process and write any data already downloaded so as to not waste the user's bandwidth.
Also those downloaded chunks must wait for decryption to finish if it is in progress, and decrypt any chunks downloaded but not yet decrypted.

Also the AsyncQueue is adjusted to have a flag as to whether operations can be discarded on MegaClient destruction (which is one case for transfer deletion)
Decryption of downloads are not discardable, but encrypt for uploads are.
If the MegaClient is destroyed, we let the decryptions run, but discard the encryptions.

The threading issue is addressed because we no longer try to merge the mac map of a still-decrypting chunk.

### DIFF
--- a/tests/integration/SdkTest_test.cpp
+++ b/tests/integration/SdkTest_test.cpp
@@ -3736,7 +3736,6 @@ TEST_F(SdkTest, SdkTestCloudraidTransfers)
             {
                 t.reset();
                 lastOnTranferFinishedCount = onTranferFinishedCount;
-                cout << "new download" << endl;
                 deleteFile(filename.c_str());
                 onTransferUpdate_progress = 0;
                 onTransferUpdate_filesize = 0;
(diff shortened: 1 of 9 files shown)


tests/integration/SdkTest_test.cpp
```diff
@@ -3716,16 +3723,32 @@ TEST_F(SdkTest, SdkTestCloudraidTransfers)
...
+                t.resume();
+
                 lastprogress = onTransferUpdate_progress;
             }
-            WaitMillisec(100);
+            else if (onTranferFinishedCount > lastOnTranferFinishedCount)
+            {
+                t.reset();
+                lastOnTranferFinishedCount = onTranferFinishedCount;
+                cout << "new download" << endl;
```
sergiohs84, line 3739: Please, avoid to commit this temporary debugging messages. In the standard output there should be exclusively tests results. Thanks
```suggestion
```

sergiohs84:
The fix looks good and consistent. Thanks for that, Matt. I'll wait for @dwmega feedback.
I've moved the target branch of the PR to the release, BTW. If you don't mind, let me proceed with the merge, so I update the tickets and changelists (also notify iOS to prepare the new build).
Thanks

dwmega:
Looks good. Great work, @mattw-mega!

PR #2019, opened 2020-05-27
#15858. Fix sdk compilation for android

jgandres:


include/mega/filesystem.h
```diff
@@ -22,7 +22,11 @@
 #ifndef MEGA_FILESYSTEM_H
 #define MEGA_FILESYSTEM_H 1
 
-#if defined (__linux__) || defined (__ANDROID__)
+#if defined (__linux__) && !defined (__ANDROID__)
+#include </usr/include/linux/magic.h>
```
sergiohs84, line 26: I think the `/usr/include` is redundant here, and it could be counterproductive, since the default include path does not necessarily need to be at `/usr/include`
```suggestion
#include <linux/magic.h>
```

include/mega/filesystem.h
```diff
@@ -22,7 +22,11 @@
...
+#endif
+
+#if defined (__linux__)
```
sergiohs84, line 29: Maybe it's good to remark, somewhere, that `__linux__` always include `__ANDROID__`??
```suggestion
#if defined (__linux__) || defined (__ANDROID__) // __ANDROID__ is always included in __linux__
```

PR #2021, opened 2020-05-27
#15860: Resume action packets after unblocking (release PR)

polmr:
- unblock sc receiving after whyamiblocked returns unblock
- add some logging lines

### DIFF
--- a/include/mega/megaclient.h
+++ b/include/mega/megaclient.h
@@ -317,10 +317,10 @@ public:
     // check the reason of being blocked
     void whyamiblocked();
 
-    // block. This stops quering for action packets
+    // stops querying for action packets due to the account being blocked (-16 on sc channel)
     void block();
 
-    // called when unblocked. Resumes querying for action packets
+    // resumes querying for action packets when the account is unblocked (if it was previously stopped due to being blocked)
     void unblock();
 
     // dump current session
--- a/include/mega/megaclient.h
+++ b/include/mega/megaclient.h
@@ -317,6 +317,12 @@ public:
     // check the reason of being blocked
     void whyamiblocked();
 
+    // block. This stops quering for action packets
+    void block();
+
+    // called when unblocked. Resumes querying for action packets
+    void unblock();
+
     // dump current session
     int dumpsession(byte*, size_t);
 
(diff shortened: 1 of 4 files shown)


src/megaclient.cpp
```diff
@@ -2082,6 +2083,20 @@ void MegaClient::exec()
         }
         syncactivity = false;
 
+        if (stopsc || scpaused || !statecurrent || !syncsup)
+        {
+            LOG_verbose << " Megaclient exec is pending resolutions."
+                        << " scpaused=" << scpaused
+                        << " stopsc=" << stopsc
```
sergiohs84, line 2090:
```suggestion
                        << " stopsc=" << stopsc
                        << " mScStoppedDueToBlock=" << mScStoppedDueToBlock
```

src/sync.cpp
```diff
@@ -1301,7 +1301,7 @@ LocalNode* Sync::checkpath(LocalNode* l, string* localpath, string* localname, d
         isroot = l == localroot.get() && !newname.size();
     }
 
-    LOG_verbose << "Scanning: " << path;
+    LOG_verbose << "Scanning: " << path << " in=" << initializing << " full=" << fullscan << " l=" << l;
```
sergiohs84, line 1304: `if (localname == nullptr) then l == nullptr`. Is it safe to pass `l` to the logger in that case?

include/mega/megaclient.h
```diff
@@ -317,6 +317,12 @@ class MEGA_API MegaClient
     // check the reason of being blocked
     void whyamiblocked();
 
+    // block. This stops quering for action packets
```
sergiohs84, line 320:
```suggestion
    // stops querying for action packets due to the account being blocked (-16 on sc channel)
```

include/mega/megaclient.h
```diff
@@ -317,6 +317,12 @@ class MEGA_API MegaClient
...
+    void block();
+
+    // called when unblocked. Resumes querying for action packets
```
sergiohs84, line 323:
```suggestion
    // resumes querying for action packets when the account is unblocked (if it was previously stopped due to being blocked)
```

src/sync.cpp
polmr: yes, it'll print `NULL` or something similar

PR #2027, opened 2020-05-29
Delete Android bindings for already removed `getIndex()` methods

javiergm1983:
These methods are already removed from the SDK and the `develop` branch doesn't compile for Android.


javiergm1983:
trigger compilation

PR #2029, opened 2020-05-29
#15893: do not (temporarily) fail uploads when in transfer overquota

polmr:


src/transfer.cpp
```diff
@@ -387,6 +387,7 @@ void Transfer::failed(error e, DBTableTransactionCommitter& committer, dstime ti
 
     if (e == API_EOVERQUOTA)
     {
+        assert(type != PUT || !timeleft); //only expected overstorage overquota for uploads
```
mattw-mega, line 390: I think we can check both PUT and GET cases here - some code relies on testing only timeleft, so these two have to match:    `assert((type == PUT && !timeleft) || (type == GET && timeleft)); `

PR #2031, opened 2020-06-02
Add libwebsockets to the scripts to build 3rd party dependencies

mattw-mega:
Libwebsockets is needed if using Karere for chat messages.


contrib/cmake/3rdparty_deps.txt
```diff
@@ -68,6 +68,9 @@ libwebsockets
 # - If using the latest Pdfium, use at least VS2017 and skip the branch checkout above, and substitute the pdfium-masterbranch-CMakeLists.txt in vcpkg/ports/pdfium and make this one small patch (other changes may be needed if the master branch has changed):
 #      in pdfium\core\fxcrt\fx_memory_wrappers.h(26)   comment out the static_assert (uint8_t counts as an arithmentic type)
 
-
 pdfium
 pdfium-freetype
+
+# If building for use with MEGAChat messaging then libwebsocket is needed (but voice/video calling needs WebRTC which changes the dependencies a lot, openssl in particular)
+libwebsockets
```
polmr, line 75: we might want to leave this out, and have a separate deps file in MEGAchat project? Otherwise we'll end up building this for MEGAsync when that's not required. Ultimately, the idea is to have a deps file per app/os. So maybe we can keep this line and start having one without it in MEGAsync project ...

PR #2032, opened 2020-06-02
#15903. Over Disk Quota Paywall

sergiohs84:


include/megaapi.h
```diff
@@ -6029,6 +6052,11 @@ class MegaGlobalListener
          *     After calling it, this callback will be called again with the corresponding
          *     state if there is really a change.
          *
+         *     - MegaApi::STORAGE_STATE_PAYWALL = 4
+         *     The account has been full for a long time. Now most of actions are disallowed.
+         *     It's needed to call MegaApi::getUserData to check the deadline/warnings
```
polmr, line 6057:
```suggestion
         *     You will need to call MegaApi::getUserData before retrieving the deadline/warnings
```

polmr: Plus , we might want to have some words in `getUserData` noting that will also update values such as those

src/commands.cpp
```diff
@@ -452,7 +470,12 @@ void CommandPutFileBackgroundURL::procresult()
 
     if (client->json.isnumeric())
     {
-        error e = (error)client->json.getint();
+        error e = static_cast<error>(client->json.getint());
+        if (e == API_EPAYWALL)
+        {
+            client->activateoverquota(0, true);
+        }
+
```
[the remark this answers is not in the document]
sergiohs84: Good point! I think the best would be to merge this code with the custom-ETD-error branch, which already copes with the errors in all commands by calling `checkError()`. It is almost ready to merge, so I can update this branch and avoid the repetead block by moving it to the new `checkError()`.
As for the specs, maybe you checked an outdated version? It specifies a long list of commands, not only the `p`: 

> New command-level error code EPAYWALL (name open for debate), likely to be errorcode -29. This must be supported on at least the API commands: m, p, d, g, u, ufa, pfa, pp, ul, l, s, s2, a, dr, dv, rv, ur, ur2, upc, upca, ui, se, sec, usabd, smss, smsv, smsr.

src/megaclient.cpp
```diff
@@ -927,20 +928,24 @@ void MegaClient::activateoverquota(dstime timeleft)
...
     {
         LOG_warn << "Storage overquota";
-        for (transfer_map::iterator it = transfers[PUT].begin(); it != transfers[PUT].end(); it++)
+        int end = (isPaywall) ? PUT : GET;  // in Paywall state, none DLs/UPs can progress
+        for (int d = GET; d <= end; d += PUT - GET)
         {
-            Transfer *t = it->second;
-            t->bt.backoff(NEVER);
-            if (t->slot)
+            for (transfer_map::iterator it = transfers[d].begin(); it != transfers[d].end(); it++)
```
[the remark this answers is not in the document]
sergiohs84: Yep, definitely. I'll fix it. Thanks

src/megaclient.cpp
```diff
@@ -3217,22 +3223,26 @@ bool MegaClient::abortbackoff(bool includexfers)
     if (includexfers)
     {
         overquotauntil = 0;
-        int end = (ststatus != STORAGE_RED) ? PUT : GET;
-        for (int d = GET; d <= end; d += PUT - GET)
+        if (ststatus != STORAGE_PAYWALL)    // in ODQ Paywall, ULs/DLs are not allowed
         {
-            for (transfer_map::iterator it = transfers[d].begin(); it != transfers[d].end(); it++)
+            // in ODQ Red, only ULs are disallowed
+            int end = (ststatus != STORAGE_RED) ? PUT : GET;
```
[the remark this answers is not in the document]
sergiohs84: Yes, it makes perfect sense. Let's retry only if state is other than Paywall, no in Paywall.

src/commands.cpp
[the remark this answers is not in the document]
mattw-mega: Sounds good - I must have been look at an outdated spec then, thanks

include/megaapi.h
mdsahinco: Ok

sergiohs84:
trigger compilation

joel-meng-nz:
trigger compilation

src/commands.cpp
```diff
@@ -7929,19 +8020,19 @@ CommandGetRegisteredContacts::CommandGetRegisteredContacts(MegaClient* client, c
...
 void CommandGetRegisteredContacts::procresult()
-{
-    processResult(*client->app, client->json);
-}
-
-void CommandGetRegisteredContacts::processResult(MegaApp& app, JSON& json)
 {
     Error e;
-    if (checkError(e, json))
+    if (checkError(e, client->json))
```
sergiohs84, line 8025: I've just moved the error's handling to the non-static method. The static `processResult()` used by unit-tests only aims to test the JSON response, not the error cases.

src/command.cpp
```diff
@@ -81,16 +84,24 @@ bool Command::checkError(Error& errorDetails, JSON& json)
                        errorDetails.setLinkStatus(json.getint());
                         break;
                     case EOO:
-                        return true;
+                        exit = true;
```
polmr, line 87: this now can entail returning false (if !errorDetected). could that be a problem?

src/megaapi_impl.cpp
```diff
@@ -31314,6 +31368,35 @@ void MegaHandleListPrivate::addMegaHandle(MegaHandle h)
...
+
+MegaIntegerList* MegaIntegerListPrivate::copy() const
+{
+    return new MegaIntegerListPrivate(mIntegers);
+}
+
+int64_t MegaIntegerListPrivate::get(int i) const
+{
+    if (i >= static_cast<int>(mIntegers.size()))
+        return -1;
```
polmr, line 31390:
```suggestion
        {
                return -1;
        }
```

src/command.cpp
sergiohs84: I don't think so. It's made in purpose: if no errocode is read from the JSON (whether `[<error_code>]` or `{"err":<error_code>, ...}`), then there is no error detected at this stage.

src/megaapi_impl.cpp
sergiohs84: Already agreed that `if`s and alike, when followed by single-line, can be left without curly brackets. I also prefer to add them, but not mandatory ;)

PR #2033, opened 2020-06-03
With thread_local in use, the text appearing in MEGAsync on windows X…

mattw-mega:
…P was italic with strikethrough.

### DIFF
--- a/include/mega/logging.h
+++ b/include/mega/logging.h
@@ -201,10 +201,11 @@ class SimpleLogger
     static OutputStreams getOutput(enum LogLevel ll);
 #else
 
-#ifdef WIN32
-    //static thread_local 
+#ifdef WIN64
+    static thread_local std::array<char, LOGGER_CHUNKS_SIZE> mBuffer;
+#elif WIN32
     // Keep this as a normal class member on WIN32 until we abandon XP (Qt fonts appear with strikethrough on XP otherwise)
-    std::array<char, LOGGER_CHUNKS_SIZE> mBuffer;
+    /*static thread_local*/ std::array<char, LOGGER_CHUNKS_SIZE> mBuffer;
 #else
     static __thread std::array<char, LOGGER_CHUNKS_SIZE> mBuffer;
 #endif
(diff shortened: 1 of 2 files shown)


include/mega/logging.h
```diff
@@ -202,7 +202,9 @@ class SimpleLogger
 #else
 
 #ifdef WIN32
-    static thread_local std::array<char, LOGGER_CHUNKS_SIZE> mBuffer;
+    //static thread_local 
+    // Keep this as a normal class member on WIN32 until we abandon XP (Qt fonts appear with strikethrough on XP otherwise)
+    std::array<char, LOGGER_CHUNKS_SIZE> mBuffer;
```
polmr, line 207:
```suggestion
    #ifdef _WIN64
        static thread_local std::array<char, LOGGER_CHUNKS_SIZE> mBuffer;
    #else
        // Keep this as a normal class member on WIN32 until we abandon XP (Qt fonts appear with strikethrough on XP otherwise)
        std::array<char, LOGGER_CHUNKS_SIZE> mBuffer;
    #endif
```

src/logging.cpp
```diff
@@ -45,7 +45,7 @@ long long SimpleLogger::maxPayloadLogSize  = 10240;
 #ifdef ENABLE_LOG_PERFORMANCE
 
 #ifdef WIN32
-thread_local std::array<char, LOGGER_CHUNKS_SIZE> SimpleLogger::mBuffer; 
+//thread_local std::array<char, LOGGER_CHUNKS_SIZE> SimpleLogger::mBuffer;
```
polmr, line 48:
```suggestion
#ifdef _WIN64
    thread_local std::array<char, LOGGER_CHUNKS_SIZE> SimpleLogger::mBuffer; // Keep this as a normal class member on WIN32 until we abandon XP (Qt fonts appear with strikethrough on XP otherwise)
#endif
```

include/mega/logging.h
mattw-mega: good idea, thanks, will do

PR #2034, opened 2020-06-04
#15930: Update libs for Android example

aabellagm:
Updated libs and build.sh and some needed flags.
Changed android platform from 14 to 16 or 21, depending on the arch.
Changed how OpenSSL is built now, according to instructions for newer versions.
Added the current 4 archs: x86 x86_64 arm-v7a and armv64-v8a.
Archs to be built could be selected using BUILD_ARCH environment variable. If no defined, all will be built.


aabellagm:
trigger compilation android

aabellagm:
trigger compilation android

aabellagm:
trigger compilation android

aabellagm:
trigger compilation android

aabellagm:
trigger compilation android

PR #2035, opened 2020-06-04
#15957: do not do transfer io while blocked, nor dispatch any transfer

polmr:


src/megaclient.cpp
```diff
@@ -10515,6 +10530,10 @@ void MegaClient::unblock()
         stopsc = false; // will resume querying for action packets
     }
     mScStoppedDueToBlock = false;
+
+    LOG_verbose << "Unblocking MegaClient";
+
+    mBlocked = false;
```
mattw-mega, line 10536: Adding the mBlocked member seems like a good step to me.   I find the mScStoppedDueToBlock interaction with stopsc a bit hard to follow, but I think with mBlocked we would not need that one anymore?   Just test (stopsc || mBlocked) where we used to test just for stopsc - only a couple of places.    I think that would be much better because stopsc was originally meant to stop processing sc messages entirely, and it does set the backoff timer to NEVER which probably means it wouldn't resume even with the flag reset.

polmr: > it wouldn't resume even with the flag reset.
it did in our testing :-). But to be truthful, rechecking the code ... I wonder why
> with mBlocked we would not need that one anymore? [mScStoppedDueToBlock]
mmm potentially stopsc could have been set before the block takes place. And in that case we wouldn't want to unset it. Therefore the flag for `mScStoppedDueToBlock`. Checking for both flags is an alternative, we should then no longer set stopsc when received API_EBLOCKED via `sc`

polmr: Suggestions addressed.

PR #2036, opened 2020-06-05
#15835. Fix/avoid detecting renamed file old name as present based on unchanged shortname during filesystem notification overload

mattw-mega:


mattw-mega:
trigger compilation

src/win32/fs.cpp
```diff
@@ -1408,102 +1424,211 @@ void WinDirNotify::process(DWORD dwBytes)
...
-WinDirNotify::WinDirNotify(string* localbasepath, string* ignore) : DirNotify(localbasepath, ignore)
+std::mutex WinDirNotify::smNotifyMutex;
+std::atomic<unsigned> WinDirNotify::smNotifierCount{0};
+HANDLE WinDirNotify::smEventHandle = NULL;
+std::deque<std::function<void()>> WinDirNotify::smQueue;
+std::unique_ptr<std::thread> WinDirNotify::smNotifierThread;
+
+void WinDirNotify::notifierThreadFunction()
 {
+    LOG_debug << "Filesytem notify thread started";
```
sergiohs84, line 1495:
```suggestion
    LOG_debug << "Filesystem notify thread started";
```

src/win32/fs.cpp
```diff
@@ -1408,102 +1424,211 @@ void WinDirNotify::process(DWORD dwBytes)
...
+                smQueue.pop_front();
+            }
+        }
+        if (f) 
+        {
+            f();
+            recheck = true;
+        }
+    }
+    LOG_debug << "Filesytem notify thread stopped";
```
sergiohs84, line 1522:
```suggestion
    LOG_debug << "Filesystem notify thread stopped";
```

src/win32/fs.cpp
```diff
@@ -1408,102 +1424,211 @@ void WinDirNotify::process(DWORD dwBytes)
...
+        }
+        SetEvent(smEventHandle);
     }
     else
     {
-        failed = GetLastError();
-        failreason = "CreateFileW was unable to open the folder";
-        LOG_err << "Unable to initialize filesystem notifications. Error: " << failed;
+        int err = GetLastError();
+        setFailed(err, "CreateFileW was unable to open the folder");
```
sergiohs84, line 1585: It could be useful to add the `longname` to the reason here: _CreateFileW was unable to open the folder \<longname\>_ Actually, maybe better to be added to the log message below :)

tests/integration/Sync_test.cpp
```diff
@@ -1128,39 +1156,51 @@ struct StandardClient : public MegaApp
...
         auto si = syncSet.find(syncid);
         if (si == syncSet.end())
         {
-            cout << "syncid " << syncid << " not found " << endl;
+            cout << clientname << " syncid " << syncid << " not found " << endl;
             return false;
         }
 
-        // compare model aganst nodes representing remote state
+        //// compare model aganst nodes representing remote state
```
sergiohs84, line 1177:
```suggestion
        // compare model against nodes representing remote state
```

tests/integration/Sync_test.cpp
```diff
@@ -2024,6 +2085,67 @@ GTEST_TEST(Sync, BasicSync_AddLocalFolder)
     ASSERT_TRUE(clientA2.confirmModel_mainthread(model.findnode("f"), 2));
 }
 
+
+// todo: add this test once the sync can keep up with file system notifications - at the moment
+// it's too slow becuase we wait for the cloud before processing the next layer of files+folders.
```
sergiohs84, line 2090:
```suggestion
// it's too slow because we wait for the cloud before processing the next layer of files+folders.
```

src/win32/fs.cpp
```diff
@@ -1408,102 +1424,211 @@ void WinDirNotify::process(DWORD dwBytes)
...
                             | FILE_NOTIFY_CHANGE_DIR_NAME
                             | FILE_NOTIFY_CHANGE_LAST_WRITE
                             | FILE_NOTIFY_CHANGE_SIZE
                             | FILE_NOTIFY_CHANGE_CREATION,
-                              &dwBytes, &overlapped, completion))
+                              &dwBytes, &overlapped, completion);
+
+    if (!readRet && GetLastError() == ERROR_INVALID_PARAMETER)
+    {
+        // ReadDirectoryChangesW fails with ERROR_INVALID_PARAMETER when the buffer length is greater than 64 KB and the application is monitoring a directory over the network. This is due to a packet size limitation with the underlying file sharing protocols.
```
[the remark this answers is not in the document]
mattw-mega: We used to use two pre-allocated buffers that were exactly 65534 bytes long - that comment is text from the WIN32 documentation indicating why we don't use more.  Initially I was looking at increasing the size of the buffer to make sure we had enough space, but as it turned out no amount was enough (even testing with 4MB), because we always needed more on the next batch.  So for a while I did have some code in there to reduce the size back down to 64KB if we got that error.    But now with the threading solution, we only use up to 4KB in my testing (on a very fast machine).   I think retaining the 65534 size is sensible for now, it's already half the size we used to use since we are only allocating one of those buffers now, and usually we will just copy out 4KB at then reuse that buffer.   If it gets larger than that then we'll swap the buffer for processing, and allocate 64KB again for the next batch.     I've moved the comment and clarified it, as it's useful to know why we chose 65534, I think.  thanks

src/win32/fs.cpp
mattw-mega: Agreed, but it's not super simple to do so currently- we need to convert the string from wchar to char first, and also the path will have been logged a few lines above with the sync creation logging.  When we introduce the LocalPath class, there will be a simple easy to use function to make that wchar to char conversion, let's add that logging then rather than spend time on it now and change it later.   thanks!

src/win32/fs.cpp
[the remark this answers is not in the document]
mattw-mega: Actually, good point, we don't need the second call anymore since ERROR_INVALID_PARAMETER  won't occur since we are calling with just under 64KB.   I've removed that block too, thanks!

include/mega/types.h
```diff
@@ -440,14 +441,67 @@ typedef enum { TRANSFERSTATE_NONE = 0, TRANSFERSTATE_QUEUED, TRANSFERSTATE_ACTIV
...
+        if (!mNotifications.empty())
+        {
+            t = std::move(mNotifications.front());
+            mNotifications.pop_front(); 
+            return true; 
+        }
+        return false;
+    }
+
+    void unpopFront(const T& t)
```
dwmega: Why unpopFront instead of pushFront?

mattw-mega: For this usage, the item put on the front should only be the item that was just taken off, and also calling it pushFront would make it seem equal to pushBack meaning you could use it in either direction, but pushing back and popping front is the envisaged use here, I thought this naming would make it easier to see which was the intended direction.  thanks!

src/win32/fs.cpp
```diff
@@ -1435,6 +1435,9 @@ void WinDirNotify::readchanges()
 #ifndef WINDOWS_PHONE
     if (notifybuf.size() != 65534)
     {
+        // Use 65534 for the buffer size becaues (from doco):
```
sergiohs84, line 1438:
```suggestion
        // Use 65534 for the buffer size because (from documentation): 
```

PR #2045, opened 2020-06-15
Using unique_ptr helps catch errors where we might end up not dealloc…

mattw-mega:
…ating or double deallocating

In fact converting these cases revealed one case of each
- leak using copy() in MegaApiImpl::fireOnRequestFinish
- eventual double delete due to passing non-owned pointer to ownership assuming function in MegaBackupController::onTransferTemporaryError

### DIFF
--- a/include/megaapi_impl.h
+++ b/include/megaapi_impl.h
@@ -2501,8 +2501,8 @@ class MegaApiImpl : public MegaApp
         void httpServerRemoveListener(MegaTransferListener *listener);
 
         void fireOnStreamingStart(MegaTransferPrivate *transfer);
-        void fireOnStreamingTemporaryError(MegaTransferPrivate *transfer, MegaErrorPrivate* e);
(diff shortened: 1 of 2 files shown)


mattw-mega:
trigger compilation

include/megaapi_impl.h
```diff
@@ -2596,9 +2596,9 @@ class MegaApiImpl : public MegaApp
         void getCountryCallingCodes(MegaRequestListener *listener = NULL);
 
         void fireOnTransferStart(MegaTransferPrivate *transfer);
-        void fireOnTransferFinish(MegaTransferPrivate *transfer, MegaError *e, DBTableTransactionCommitter& committer);
+        void fireOnTransferFinish(MegaTransferPrivate *transfer, unique_ptr<MegaError> e, DBTableTransactionCommitter& committer);
```
sergiohs84, line 2599: I think we should pass the `MegaErrorPrivate` here, instead of `MegaError`.
Same for `fireOnTransferTemporaryError()`, `fireOnBackupFinish()` and `fireOnBackupTemporaryError()`

PR #2049, opened 2020-06-16
#15942. Implement sync commands: register, remove and update

alber2510:


src/commands.cpp
```diff
@@ -8182,4 +8182,130 @@ void CommandFolderLinkInfo::procresult()
...
+
+void CommandSyncPut::procresult()
+{
+    Error e;
+    if (checkError(e, client->json))
+    {
+        return client->app->syncput_result(e, UNDEF);
+    }
+
+    handle h = client->json.gethandle(MegaClient::NODEHANDLE); // [sync id]
```
polmr, line 8260: picky suggestion: rename h to syncID and remove the comment

include/megaapi_impl.h
```diff
@@ -2983,6 +2983,10 @@ class MegaApiImpl : public MegaApp
         std::mutex mSyncable_fa_mutex;
 #endif
 
+        void syncput_result(error, handle) override;
+        void syncputheartbeat_result(error) override;
+        void syncremove_result(error, handle) override;
```
polmr, line 2988: We could use some comments. here, or probably in megapp.h: one line indicating what does this do.
Plus, having parameters with names (surronded by /* */) would also help: 
e.g:
```
        // sync registered / updated 
        void syncput_result(error, handle /*syncId*/) override;
```

include/mega/command.h
```diff
@@ -1205,6 +1205,44 @@ class MEGA_API CommandFolderLinkInfo: public Command
...
+    // Invalid values:
+    // - type: SyncType::INVALID
+    // - nodeHandle: UNDEF
+    // - localFolder: nullptr
+    // - deviceId: nullptr
+    // - SyncName: nullptr
+    // - state: -1
+    // - subState: -1
+    // - extraData: nullptr
+    CommandSyncPut(MegaClient* client, handle syncId, SyncType type, handle nodeHandle, const char* localFolder, const char* deviceId, const char* syncName, int state, int subState, const char* extraData);
```
polmr, line 1227: Thinking-out-loud: Perhaps it'll be worth it to separate the 2 use cases: register vs update in 2 different commands, although the both entail "sp".
We might also like to have a separated App callback: `syncupdate_result`.
Later on, we might need MegaApiImpl to fire diffferent callbacks for `syncput_result` vs `syncupdate_result`, in case apps require to manage/show heartbeated syncs

include/mega/command.h
```diff
@@ -1205,6 +1205,44 @@ class MEGA_API CommandFolderLinkInfo: public Command
     CommandFolderLinkInfo(MegaClient*, handle);
 };
 
+class MEGA_API CommandSyncPut : public Command
```
polmr, line 1208: Dismissable picky proposal: throughout all this PR, change `Sync` by `HeartbeatedSync`, or something like that to differentiate this `Sync` concept from the actual Sync engine stuff. Sync related app callbacks are also sync*** which might become confusing overtime. In any case: specs talk about Syncs whereas it could be other stuff like Camera Uploads, I guess it's a little late for a full rebranding, but we still might want to consider it within the SDK ;-)

src/commands.cpp
alber2510: Yes, it's ok for me. I'll modify it

include/megaapi_impl.h
alber2510: Yes, it's ok for me. I'll modify it

include/mega/command.h
alber2510: I'm not sure how it's going to be implemented at app side, if you consider that have two commands be worth. I'll implement it.
I was thinking that many of this stuff is going to keep internally

include/mega/command.h
alber2510: Yes, I'll apply a refactoring

include/mega/command.h
sergiohs84: I kind of agree with both of you. Nevertheless, having two separate callbacks for registering and for updating could be more flexible in the future, so I think we should go that way.

include/mega/command.h
sergiohs84: Regarding refactoring these new Commands (an also param names of the heartbeats, like `syncId` or `syncName`), I agree that not always it's a `Sync`, but sometimes a CU or likely anything else in the future.
A different naming should be assigned, and I vote for renaming `sync` to `backup` wherever it's appropriate. In the end, it is for the backup centre: clients register backups, send backup-heartbeats, etc. So... *backup* everywhere :)

src/commands.cpp
```diff
@@ -8182,4 +8182,142 @@ void CommandFolderLinkInfo::procresult()
...
+void CommandBackupPutHeartBeat::procresult()
+{
+    Error e;
+    if (checkError(e, client->json))
+    {
+        return client->app->backupputheartbeat_result(e);
+    }
+
+    client->json.storeobject();
+    return client->app->backupputheartbeat_result(API_EINTERNAL);
```
sergiohs84, line 8300: This method is void, it should return nothing:
```suggestion cpp
    client->app->backupputheartbeat_result(API_EINTERNAL);
```

include/mega/megaapp.h
```diff
@@ -389,6 +389,11 @@ struct MEGA_API MegaApp
 
     virtual void getmiscflags_result(error) { }
 
+    virtual void backupput_result(error, handle /*backup id*/) { }
```
sergiohs84, line 392: Should we start using `Error` in these new callbacks, rather than `error`? Maybe a reference o a unique_ptr, if dynamic allocation is needed. The advantage would be to get the additional information of the error at upper layers when the API provides so.

alber2510: Solved it

PR #2056, opened 2020-06-18
#15943. Sync/Backup - New user attribute to save in TLV device id - device name

alber2510:


src/megaapi_impl.cpp
```diff
@@ -14875,6 +14902,21 @@ void MegaApiImpl::getua_result(TLVstore *tlv, attr_t type)
...
+                if (!buf)
+                {
+                    e = API_ENOENT;
+                    break;
+                }
+                else
+                {
+                    request->setName(Base64::atob(buf).c_str());
+                }
+
```
sergiohs84, line 14917: The `if` above has a break, so no need for an `else` block here :)
```suggestion cpp
                request->setName(Base64::atob(buf).c_str());
```

PR #2059, opened 2020-06-19
HotFix: Invalidate attribute when we receivie API_EEXPIRED

alber2510:


alber2510:
Apps have built their versions and it's no a critical error. We add in `develop`

PR #2071, opened 2020-06-25
#15003. Handle custom etd response at request level

alber2510:


sergiohs84:
Please, @mattw-mega , would you mind to check this PR too? I think it's a possible workaround that is flexible enough for future changes in the format/syntax of the `"err"` JSON.

src/megaclient.cpp
```diff
@@ -1770,7 +1770,18 @@ void MegaClient::exec()
                             else
                             {
                                 // request failed
-                                error e = (error)atoi(pendingcs->in.c_str());
+                                error e = API_EINTERNAL;
+                                std::string err;
+                                if (strncmp(pendingcs->in.c_str(), "{\"err\":", 7) == 0 && pendingcs->in.find("}") != std::string::npos)
+                                {
+                                    err = pendingcs->in.substr(0, pendingcs->in.find("}") + 1);
```
mattw-mega, line 1777: I think you could use Command::checkError here - the difference is probably just whether the initial { would have been consumed yet or not.   In the speculative instant completion branch, I needed to change it to process from the { anyway.   Or at least use JSON::storeobject, which will account for nested objects inside the error object, in case we add those in future.   Searching for the closing } won't deal with those nested objects - and storeobject() would also extract the negative number if it is just the number, so no need for a separate if/else for that.    Presumably checkError would do the same, if used, though that one doesn't extract the string - probably better to just use storeobject() on matching  "-" or "{err:".  thanks

src/megaclient.cpp
```diff
@@ -1853,7 +1864,7 @@ void MegaClient::exec()
                                 pendingcs = NULL;
                                 csretrying = false;
 
-                                reqs.servererror(API_ESSL, this);
+                                reqs.servererror("-23", this);
```
mattw-mega, line 1867: we can use std::to_string(API_ESSL) here, to avoid hard-coding magic constants and make it clear which error we are producing.   Performance is not important for this case, if that was the concern.  thanks

sergiohs84, line 1867:
```suggestion
                                reqs.servererror(std::to_string(API_ESSL), this);
```

src/megaclient.cpp
alber2510: @mattw-mega I've applied a modification using `storeObject`, as you comment.  I use `{"err:"` for matching insead of `-` because there is a code to check `if (!e) then I suppose that number returned can be `0`
Advise me when you finish the speculative instant completion branch and I check if I can use `Command::checkError` 
Thanks for review :)
If you observer any other thing comment me to solve it

PR #2074, opened 2020-06-29
#16157. Hotfix: PR #2070 changes performance fix for filesystem type lookup for character escaping

mattw-mega:


src/megaclient.cpp
```diff
@@ -11028,6 +11037,7 @@ void MegaClient::disabletransferresumption(const char *loggedoutid)
     {
         return;
     }
+    purgeOrphanTransfers(true);
```
sergiohs84, line 11040: After this first call to `closetc()` (now both `purgeOrphanTransfers()` and `closetc()`), the `tctable` should be closed already. I don't understand why we open the transfer's cache again and repeat the same operations. Seems like the result of a merge at some point, in my opinion (despite it was already there in the code before this change, so feel free to dismiss this comment).

sergiohs84:
I think my comment is not required to be addressed along with the hotfix, but we should proceed with the patch for the release.
Please, @mattw-mega, if you feel the suggestion about `disabletransferresumption()` makes sense, let's prepare a small PR to adjust it. Thanks

src/megaclient.cpp
mattw-mega: Yes it does look a bit odd.  It's been that way since 2016 though.   I think perhaps the close/open/close sequence is intended to overwrite the file with an empty table?  I guess the 'disable' also implies that existing cached transfers are deleted, and enabling won't bring them back.    I imagine most apps either have this turned on permanently, or turned off permanently, so there is probably quite a lot of simplification we could do here if that is the case - I wonder if we could make this entirely SDK managed similar to sync resumption, rather than having the app call to enable/disable.

sergiohs84: Looking  at the original commit that introduced the double close (8e037b694bff37c03a49563242848e27c656a647), I think the original purpose was to allow the purge of a cache related to a folder-link, in example, or to an already logged out instance of MegaApi. No need to change it, IMO.

As for the transfer-resumption, we still have Android without transfer-resumption enabled (even if you don't believe it, :facepalm: ), though they have a ticket (#13569). But I'd say transfers should always be resumed, indeed. Let's move the topic to an open discussion with apps teams, because there might be hidden side effects :)

PR #2084, opened 2020-07-01
#16229: protect transfer in up/down controller onTransferStart

polmr:


polmr:
This attemps to completely fix the bug, rendering the if(transfer) controls no longer required: https://github.com/meganz/sdk/pull/2085

PR #2085, opened 2020-07-01
#16229: Proper folder transfers cancellation

polmr:
- allow to cancel a folder transfer (via TYPE_CANCEL_TRANSFER)
- MegaFolderDown/UploadController::cancel now cancel also pending transfers
- delayed setting transfer = nullptr, to fully digest all subtransfers callbacks
- change order in abortPendingActions: first remove queued transfers and then in progress

### DIFF
--- a/include/megaapi_impl.h
+++ b/include/megaapi_impl.h
@@ -216,6 +216,7 @@ protected:
     int recursive;
     int tag;
     int pendingTransfers;
+    bool cancelled = false;
     std::set<MegaTransferPrivate*> subTransfers;
     int mIncompleteTransfers = { 0 };
     MegaErrorPrivate mLastError = { API_OK };
@@ -394,6 +395,9 @@ public:
 protected:
     void downloadFolderNode(MegaNode *node, string *path);
     void checkCompletion();
+
+private:
+    bool cancelled = false;
 
 public:
     void onTransferStart(MegaApi *, MegaTransfer *t) override;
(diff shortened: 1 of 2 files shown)


src/megaapi_impl.cpp
```diff
@@ -20543,9 +20558,11 @@ void MegaApiImpl::sendPendingRequests()
             if (!megaTransfer->isStreamingTransfer())
             {
                 Transfer *transfer = megaTransfer->getTransfer();
-                if (!transfer)
+                if (!transfer) //folder transfer
```
polmr, line 20561: Note for reviewer, there might be a better way to identify a folder transfer

sergiohs84: What about `MegaTransfer::isFolderTransfer()`??

include/megaapi_impl.h
```diff
@@ -395,6 +396,9 @@ class MegaFolderDownloadController : public MegaTransferListener, public MegaRec
     void downloadFolderNode(MegaNode *node, string *path);
     void checkCompletion();
 
+private:
+    bool cancelled = false;
```
sergiohs84, line 400: It is already declared at the parent class :) No need here

src/megaapi_impl.cpp
```diff
@@ -7444,6 +7444,17 @@ void MegaApiImpl::abortPendingActions(error preverror)
     // -- Transfers in progress --
     {
         DBTableTransactionCommitter committer(client->tctable);
+
```
sergiohs84, line 7447: Note for reviewers: we move the cleanup of the queued transfers before the cleanup of ongoing transfers because, otherwise, finishing the the ongoing folder transfer will destroy the object, and it will delete queued transfers looking for them in the queue. However, if the queue is empty, we save some cycles.

PR #2087, opened 2020-07-02
#16239. Do cancel all transfers

polmr:
- Store folder transfers in a separate map
- When queuing the TYPE_CANCEL_TRANSFERS, store in the last queued transfer.
- When resolving TYPE_CANCEL_TRANSFERS:
1. cancel pending transfers up to the last one queued by the time CANCEL_TRANSFERS req was created
2. cancel client->transfers
3. cancel folder-transfers
- Renew MegaFolderUp/DownloadController::cancel: it was not removing initiated subtransfers from cache!


src/megaapi_impl.cpp
```diff
@@ -25184,14 +25304,73 @@ void MegaFolderDownloadController::cancel()
...
-        subTransfer->setState(MegaTransfer::STATE_COMPLETED);
-        DBTableTransactionCommitter committer(client->tctable);
-        megaApi->fireOnTransferFinish(subTransfer, make_unique<MegaErrorPrivate>(API_EINCOMPLETE), committer);
+
+        Transfer *transfer = subTransfer->getTransfer();
+        if (!transfer)
+        {
+            LOG_warn << "Subtransfer without attached Transfer for folder transfer: " << subTransfer->getFileName();
+
+            subTransfer->setState(MegaTransfer::STATE_COMPLETED); //TODO: copmleted?
```
sergiohs84, line 25319: Not sure about the implications of the state of the transfer. Isn't it exclusively for the app? Maybe it's more appropriate to report it as `STATE_CANCELLED`.
(same here:
https://github.com/meganz/sdk/pull/2087/files#diff-c3de3d92d9d9472cc3d715cb978d2940R23952
https://github.com/meganz/sdk/pull/2087/files#diff-c3de3d92d9d9472cc3d715cb978d2940R23905
https://github.com/meganz/sdk/pull/2087/files#diff-c3de3d92d9d9472cc3d715cb978d2940R25366)

src/megaapi_impl.cpp
```diff
@@ -25184,14 +25304,73 @@ void MegaFolderDownloadController::cancel()
...
+        while (iterator != files.end())
+        {
+            File *file = *iterator;
+            iterator++;
+            if (file->tag == subTransfer->getTag())
+            {
+                found = true;
+                if (!file->syncxfer)
+                {
+                    client->stopxfer(file, nullptr); //No commiter, we are here within sendPendingRequest
```
sergiohs84, line 25351: Not sure if the comment is correct here. Why should the committer be invalid here? Shouldn't we use the committer for the cancellation of all transfers declared at the beginning of this method?
(same here: https://github.com/meganz/sdk/pull/2087/files#diff-c3de3d92d9d9472cc3d715cb978d2940R23937)

src/megaapi_impl.cpp
```diff
@@ -20649,6 +20670,38 @@ void MegaApiImpl::sendPendingRequests()
...
             if (!flag)
             {
+                long long cancelledPending = 0;
+                // 1. cancel queued transfers, not yet started (and not added to cache), up to the last one queued when cancelTransfers was queued
+                auto lastQueuedTransfer = request->getTransferTag();
+                while (MegaTransferPrivate *transfer = transferQueue.pop())
+                {
+                    if (transfer->getPlaceInQueue() > lastQueuedTransfer)
+                    {
+                        break;
```
mattw-mega, line 20680: I think we will have popped off one too many transfers off the queue here - should we use push_front to put it back?

src/megaapi_impl.cpp
```diff
@@ -20649,6 +20670,38 @@ void MegaApiImpl::sendPendingRequests()
...
+                    }
+                    if (!transfer->isSyncTransfer() && transfer->getType() == direction)
+                    {
+                        fireOnTransferStart(transfer);
+                        transfer->setState(MegaTransfer::STATE_CANCELLED);
+                        fireOnTransferFinish(transfer, make_unique<MegaErrorPrivate>(API_EINCOMPLETE), committer);
```
mattw-mega, line 20686: I wonder if it would be a better user experience if we were to introduce a new error code EUSERCANCELLED which would reassure people that the transfer cancelled due to their own action, and not some other cause?  If course we would need to show that in the GUI too

src/megaapi_impl.cpp
```diff
@@ -23829,14 +23890,73 @@ void MegaFolderUploadController::cancel()
     //remove subtransfers from pending transferQueue
     megaApi->cancelPendingTransfersByFolderTag(tag);
 
+    //remove ongoing subtransfers
+    long long cancelledSubTransfers = 0;
+    DBTableTransactionCommitter committer(client->tctable);
```
mattw-mega, line 23895: Ah yes, of course this should be outside the loop :-)

src/megaapi_impl.cpp
```diff
@@ -23829,14 +23890,73 @@ void MegaFolderUploadController::cancel()
...
-        subTransfer->setState(MegaTransfer::STATE_COMPLETED);
-        DBTableTransactionCommitter committer(client->tctable);
-        megaApi->fireOnTransferFinish(subTransfer, make_unique<MegaErrorPrivate>(API_EINCOMPLETE), committer);
+
+        Transfer *transfer = subTransfer->getTransfer();
+        if (!transfer)
+        {
+            LOG_warn << "Subtransfer without attached Transfer for folder transfer: " << subTransfer->getFileName();
+
+            subTransfer->setState(MegaTransfer::STATE_COMPLETED); //TODO: copmleted?
```
mattw-mega, line 23905: I see there is a STATE_CANCELLED, that would be more appropriate

src/megaapi_impl.cpp
```diff
@@ -23829,14 +23890,73 @@ void MegaFolderUploadController::cancel()
...
+        }
+
+#ifdef _WIN32
+        if (transfer->type == GET)
+        {
+            transfer->localfilename.append("", 1);
+            WIN32_FILE_ATTRIBUTE_DATA fad;
+            if (GetFileAttributesExW((LPCWSTR)transfer->localfilename.data(), GetFileExInfoStandard, &fad))
+                SetFileAttributesW((LPCWSTR)transfer->localfilename.data(), fad.dwFileAttributes & ~FILE_ATTRIBUTE_HIDDEN);
+            transfer->localfilename.resize(transfer->localfilename.size()-1);
```
mattw-mega, line 23918: It looks like this is the same code as in MegaFileGet::updatelocalname() - we could call that once we find the File* inside the transfer instead?

src/megaapi_impl.cpp
```diff
@@ -23829,14 +23890,73 @@ void MegaFolderUploadController::cancel()
...
+        subTransfer->setLastError(&megaError);
+
+        bool found = false;
+        file_list files = transfer->files;
+        file_list::iterator iterator = files.begin();
+        while (iterator != files.end())
+        {
+            File *file = *iterator;
+            iterator++;
+            if (file->tag == subTransfer->getTag())
```
mattw-mega, line 23932: I think every file download that is part of the folder download gets the same tag?   So we won't necessarily match the exact transfer file for this MegaTransfer - perhaps we should remove all that have the matching tag, without breaking the loop?

src/megaapi_impl.cpp
mattw-mega: The committer parameters are really just to remind us that one should be on the stack before we make the call - I see we have created one outside the loop anyway,  there's no harm in passing &committer here, that would be the normal way to use it, so it all looks normal.   There's no harm in having more than one on the stack either.  I would have made the parameter a reference rather than pointer if we didn't have the case of these being called from destructors (with committers higher on the stack already)  thanks

src/megaapi_impl.cpp
polmr: you are right, further more: we are poping sync transfers & those with different direction too! fixing it, by having a method that pops up exactly what we want in TransferQueue. thanks for spotting that

src/megaapi_impl.cpp
polmr: I believe this might go beyond the scope of this PR, don't you think? :relaxed:

src/megaapi_impl.cpp
polmr: I took this code from TYPE_CANCEL_TRANSFER . I thought the idea was to unhide the temporary download file ... Not sure if this relates to File* ... :thinking:

polmr: On second though, both cases could be wrong, and this should only be done if there were no more File* associated with that Transfer?

src/megaapi_impl.cpp
polmr: I too borrowed the code from TYPE_CANCEL_TRANSFER. 
> I think every file download that is part of the folder download gets the same tag? 
they get the same getFolderTransferTag, not the same tag, don't they?

I'm concerned that there could be more File* matching the MegaTransfer::tag. that would be happening in TYPE_CANCEL_TRANSFER too :thinking: . and breaking the loop could be incorrect ...

src/megaapi_impl.cpp
polmr: mmm the problem with having a commiter is that it turns into an assert failing:
stopxfer -> removeTransferFile -> filecachedel -> checkCommiter:

```
void DbTable::checkCommitter(DBTableTransactionCommitter* committer)
{
    assert(!committer || committer == mTransactionCommitter);
}``` 
that is why i passed nullptr (since it is not required)

Comment (polmr) [reply src/megaapi_impl.cpp]: I've added a new commit to have a workaround. resusing the table commiter if one was present

File: src/megaapi_impl.cpp (hunk 8 above)
Comment (polmr) [reply src/megaapi_impl.cpp:23959]: Second read: no concerns: only one File will match MegaTransfer::tag. File::tag is unique AFAIK

File: src/megaapi_impl.cpp (hunk 4 above)
Comment (mattw-mega) [reply src/megaapi_impl.cpp]: Yes - just a thought for the future - actually since we set state cancelled, maybe the GUI can use that to indicate to the user that the transfer stopped due to their own action, rather than being something they need to worry about - perhaps we don't need the error code.

File: src/megaapi_impl.cpp (hunk 8 above)
Comment (mattw-mega) [reply src/megaapi_impl.cpp:23959]: Ah ok, good to know, thanks for checking that :-)

File: src/megaapi_impl.cpp (hunk 2 above)
Comment (mattw-mega) [reply src/megaapi_impl.cpp]: Ah I see, you've reminded me - the assert is there to alert us when we ended up with two committers on the stack (not that it's really a problem, but just to keep things tidy).  And it's a case where we've been called from a destructor (~MegaTransferPrivate(), which calls cancel() on the upload/download controller), which is why the first committer on the stack hasn't been passed in here.

File: src/megaapi_impl.cpp (hunk 7 above)
Comment (mattw-mega) [reply src/megaapi_impl.cpp:23945]: I just tested the normal case, and the temp file for download is hidden on the first progress() call, and unhidden once all the data has been downloaded, in Transfer::complete().   I think this is probably just to keep the directory looking tidy if many files are being downloaded at once.   And for a cancelled download, we should be about to delete that temporary file anyway, so I think this is fine for now.   thanks

File: src/megaapi_impl.cpp (hunk 4 above)
Comment (polmr) [reply src/megaapi_impl.cpp]: Yes, as a matter of fact, MEGAsync does not show CANCELLED transfers AFAIK

**Comment by sergiohs84** [comment]
Careful, the last run from Jenkins has crashed at test `RecursiveUploadWithLogout`

File: src/megaapi_impl.cpp [hunk 9]
```diff
@@ -11914,6 +11914,29 @@ void MegaApiImpl::file_complete(File *f)
     }
 }
 
+void MegaApiImpl::transfer_complete(Transfer *t)
+{
+    if (t)
```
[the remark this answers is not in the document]
polmr: check removed

PR #2088, opened 2020-07-03
#16244. Merged LocalPath, and used it to fix sync cases for drives such as F:\

mattw-mega:
Squash-merged the feature/local-path-class branch, keeping enough of it to add LocalPath to the code base, and use it for fixing this sync case.
On windows, syncing of drive paths such as F:\ which have to end in \, had trouble because the path addition ended up with paths like F:\\folder which are illegal.
Now LocalPath takes care of concatenating those paths, and getting those cases right.
Additionally, it simplifies the code considerably by factoring out fiddly string manipulation, and worries about which platform you're on.
Additionally, it abstracts away the windows case where we are storing two-byte characters in std::string, so now we have a type tracking whether the string contains those.
Previously it was very easy to mis-use a string, passing a two-byte character string to a function expecting UTF-8 or vice versa.
But the main problem fixed is that now you can set up a sync from F:\ to eg. /F in the cloud.

### DIFF
--- a/include/mega/filesystem.h
+++ b/include/mega/filesystem.h
@@ -155,8 +155,8 @@ public:
     void truncate(size_t bytePos) { localpath.resize(bytePos); }
     size_t lastpartlocal(const FileSystemAccess& fsaccess) const;
     void append(const LocalPath& additionalPath);
-    void separatorAppend(const LocalPath& additionalPath, bool separatorAlways, const std::string& localseparator);
-    void separatorPrepend(const LocalPath& additionalPath, const std::string& localseparator);
+    void appendWithSeparator(const LocalPath& additionalPath, bool separatorAlways, const std::string& localseparator);
+    void prependWithSeparator(const LocalPath& additionalPath, const std::string& localseparator);
     void trimNonDriveTrailingSeparator(const FileSystemAccess& fsaccess);
     bool findNextSeparator(size_t& separatorBytePos, const FileSystemAccess& fsaccess) const;
     bool findPrevSeparator(size_t& separatorBytePos, const FileSystemAccess& fsaccess) const;
--- a/include/mega/filesystem.h
+++ b/include/mega/filesystem.h
@@ -117,6 +117,80 @@ struct MEGA_API AsyncIOContext
     FileAccess *fa;
 };
 
+
+
+// LocalPath represents a path in the local filesystem, and wraps up common operations in a convenient fashion.
+// On mac/linux, local paths are in utf8 but in windows local paths are utf16, that is wrapped up here.
+
+struct MEGA_API FileSystemAccess;
+class MEGA_API LocalPath;
+
+class ScopedLengthRestore {
+    LocalPath& path;
+    size_t length;
+public:
+    // On destruction, puts the LocalPath length back to what it was on construction of this class
+    ScopedLengthRestore(LocalPath&);
+    ~ScopedLengthRestore();
+};
+
+class MEGA_API LocalPath
+{
+    std::string localpath;
+
+    friend class ScopedLengthRestore;
+    size_t getLength() { return localpath.size(); }
+    void setLength(size_t length) { localpath.resize(length); }
+
+public:
+
+    LocalPath() {}
+    explicit LocalPath(string&& s) : localpath(std::move(s)) {}
+
+    std::string* editStringDirect();
+    const std::string* editStringDirect() const;
+    bool empty() const;
+    void clear() { localpath.clear(); }
+    void erase() { localpath.erase(); }
+    void truncate(size_t bytePos) { localpath.resize(bytePos); }
+    size_t lastpartlocal(const FileSystemAccess& fsaccess) const;
+    void append(const LocalPath& additionalPath);
+    void separatorAppend(const LocalPath& additionalPath, bool separatorAlways, const string& localseparator);
+    void separatorPrepend(const LocalPath& additionalPath, const string& localseparator);
+    void trimNonDriveTrailingSeparator(const FileSystemAccess& fsaccess);
+    bool findNextSeparator(size_t& separatorBytePos, const FileSystemAccess& fsaccess) const;
+    bool findPrevSeparator(size_t& separatorBytePos, const FileSystemAccess& fsaccess) const;
+    size_t getLeafnameByteIndex(const FileSystemAccess& fsaccess) const;
+    bool backEqual(size_t bytePos, const string& compareTo) const;
+    bool backEqual(size_t bytePos, const LocalPath& compareTo) const;
+    LocalPath subpathFrom(size_t bytePos) const;
+    string substrTo(size_t bytePos) const;
+
+    bool isContainingPathOf(const LocalPath& path, const FileSystemAccess& fsaccess);
+
+    
+    string toPath(const FileSystemAccess& fsaccess) const;
+    string toName(const FileSystemAccess& fsaccess, FileSystemType fsType = FS_UNKNOWN) const;
+    static LocalPath fromPath(const string& path, const FileSystemAccess& fsaccess);
+    static LocalPath fromName(string path, const FileSystemAccess& fsaccess, FileSystemType fsType);
+    static LocalPath fromLocalname(string localname);
+    static LocalPath tmpNameLocal(const FileSystemAccess& fsaccess);
+
+    bool operator==(const LocalPath& p) const { return localpath == p.localpath; }
+    bool operator!=(const LocalPath& p) const { return localpath != p.localpath; }
+    bool operator<(const LocalPath& p) const { return localpath < p.localpath; }
+};
+
+inline LocalPath operator+(LocalPath& a, LocalPath& b)
+{
+    LocalPath result = a;
+    result.append(b);
+    return result;
+}
+
+// map a request tag with pending paths of temporary files
+typedef map<int, vector<LocalPath> > pendingfiles_map;
+
 struct MEGA_API DirAccess;
 
 // generic host file/directory access interface
@@ -145,7 +219,7 @@ struct MEGA_API FileAccess
     int errorcode = 0;
 
     // for files "opened" in nonblocking mode, the current local filename
-    string nonblocking_localname;
+    LocalPath nonblocking_localname;
 
     // waiter to notify on filesystem events
     Waiter *waiter;
@@ -153,17 +227,17 @@ struct MEGA_API FileAccess
     // blocking mode: open for reading, writing or reading and writing.
     // This one really does open the file, and openf(), closef() will have no effect
     // If iteratingDir is supplied, this fopen() call must be for the directory entry being iterated by dopen()/dnext()
-    virtual bool fopen(string*, bool read, bool write, DirAccess* iteratingDir = nullptr) = 0;
+    virtual bool fopen(LocalPath&, bool read, bool write, DirAccess* iteratingDir = nullptr) = 0;
 
     // nonblocking open: Only prepares for opening.  Actually stats the file/folder, getting mtime, size, type.
     // Call openf() afterwards to actually open it if required.  For folders, returns false with type==FOLDERNODE.
-    bool fopen(string*);
+    bool fopen(LocalPath&);
 
     // check if a local path is a folder
-    bool isfolder(string*);
+    bool isfolder(LocalPath&);
 
     // update localname (only has an effect if operating in by-name mode)
-    virtual void updatelocalname(string*) = 0;
+    virtual void updatelocalname(LocalPath&) = 0;
 
     // absolute position read, with NUL padding
     bool fread(string *, unsigned, unsigned, m_off_t);
@@ -187,13 +261,13 @@ struct MEGA_API FileAccess
 
     virtual bool asyncavailable() { return false; }
 
-    AsyncIOContext *asyncfopen(string *);
+    AsyncIOContext *asyncfopen(LocalPath&);
 
     // non-locking ops: open/close temporary hFile
     bool asyncopenf();
     void asyncclosef();
 
-    AsyncIOContext *asyncfopen(string *, bool, bool, m_off_t = 0);
+    AsyncIOContext *asyncfopen(LocalPath&, bool, bool, m_off_t = 0);
     AsyncIOContext* asyncfread(string *, unsigned, unsigned, m_off_t);
     AsyncIOContext* asyncfwrite(const byte *, unsigned, m_off_t);
 
@@ -237,10 +311,10 @@ public:
 struct MEGA_API DirAccess
 {
     // open for scanning
-    virtual bool dopen(string*, FileAccess*, bool) = 0;
+    virtual bool dopen(LocalPath*, FileAccess*, bool) = 0;
 
     // get next record
-    virtual bool dnext(string*, string*, bool = true, nodetype_t* = NULL) = 0;
+    virtual bool dnext(LocalPath&, LocalPath&, bool = true, nodetype_t* = NULL) = 0;
 
     virtual ~DirAccess() { }
 };
@@ -248,7 +322,7 @@ struct MEGA_API DirAccess
 struct Notification
 {
     dstime timestamp;
-    string path;
+    LocalPath path;
     LocalNode* localnode;
 };
 
@@ -298,12 +372,12 @@ public:
     int  getFailed(string& reason);
 
     // base path
-    string localbasepath;
+    LocalPath localbasepath;
 
     virtual void addnotify(LocalNode*, string*) { }
     virtual void delnotify(LocalNode*) { }
 
-    void notify(notifyqueue, LocalNode *, const char*, size_t, bool = false);
+    void notify(notifyqueue, LocalNode *, LocalPath&&, bool = false);
 
     // filesystem fingerprint
     virtual fsfp_t fsfingerprint() const;
@@ -312,12 +386,12 @@ public:
     // This should return false for any FAT filesystem.
     virtual bool fsstableids() const;
 
-    // ignore this
-    string ignore;
+    // ignore this (debris folder)
+    LocalPath ignore;
 
     Sync *sync;
 
-    DirNotify(string*, string*);
+    DirNotify(const LocalPath&, const LocalPath&);
     virtual ~DirNotify() {}
 };
 
@@ -345,7 +419,7 @@ struct MEGA_API FileSystemAccess : public EventTrigger
 
     // instantiate DirNotify object (default to periodic scanning handler if no
     // notification configured) with given root path
-    virtual DirNotify* newdirnotify(string*, string*, Waiter*);
+    virtual DirNotify* newdirnotify(LocalPath&, LocalPath&, Waiter*);
 
     // check if character is lowercase hex ASCII
     bool islchex(char) const;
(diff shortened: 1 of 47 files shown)


dwmega:
trigger compilation

src/filesystem.cpp
```diff
@@ -856,4 +853,207 @@ bool FileInputStream::read(byte *buffer, unsigned size)
     return false;
 }
 
+const std::string* LocalPath::editStringDirect() const
+{
+    // this function for compatibiltiy while converting to use LocalPath class.  TODO: phase out this function
```
sergiohs84, line 858: There's 79 usages of this method. It seems not really phased out. In case you consider it's important to stop using it, please, create a ticket, @mattw-mega

src/filesystem.cpp
```diff
@@ -856,4 +853,207 @@ bool FileInputStream::read(byte *buffer, unsigned size)
...
+
+bool LocalPath::backEqual(size_t bytePos, const LocalPath& compareTo) const
+{
+    auto n = compareTo.localpath.size();
+    return bytePos + n == localpath.size() && memcmp(compareTo.localpath.data(), localpath.data() + bytePos, n);
+}
+
+LocalPath LocalPath::subpathFrom(size_t bytePos) const
+{
+    return LocalPath::fromLocalname(localpath.substr(bytePos));
```
sergiohs84, line 988: Should we check here that `bytePos <= localpath.size()`? It depends on the usage of this function, but if the condition fails, `string::substr()` will throw out-of-range exception.

src/filesystem.cpp
mattw-mega: Yes, I should have commented this one better - we're still using editStringDirect for now, in order to still access the std::string that is now wrapped up in this class, but has exactly the same content.  In order to hopefully make review easier, the std::string content is still the same now as it was before wrapping it up, so in review we can compare before and after and see that the contained string is still the same.   Later we can change it to std::wstring for windows which will match its use much, much better.   Ticket created: 16393.   But for now, as this ticket is already changing a huge number of lines (sorry about that), I've tried to limit the amount of change, and pass the old string to functions that didn't need to be converted to at this point, and so this function is still used quite a lot.    But I think the PR achieves the main aim, which is use a c++ type to represent the local paths, so it's much easier to track whether a std::string is containing utf8 or (for windows) utf16.   And hopefully minimizing the amount of lines that need to be reviewed to achieve that.   thanks!

include/mega/filesystem.h
```diff
@@ -379,43 +453,43 @@ struct MEGA_API FileSystemAccess : public EventTrigger
...
-    virtual bool copylocal(string*, string*, m_time_t) = 0;
+    virtual bool copylocal(LocalPath&, LocalPath&, m_time_t) = 0;
 
     // delete file
-    virtual bool unlinklocal(string*) = 0;
+    virtual bool unlinklocal(LocalPath&) = 0;
 
     // delete empty directory
-    virtual bool rmdirlocal(string*) = 0;
+    virtual bool rmdirlocal(LocalPath&) = 0;
```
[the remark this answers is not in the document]
mattw-mega: Yes, I totally agree and wanted to do it that way.  However, on windows we are prevented from doing that for now because they do actually edit the string.  Because it's storing a std::string with utf16 characters, the terminating null needs to be two bytes as well, but for std::string the terminating null is only one byte.  So the functions for windows append another null, and then remove it again before returning.    So, the parameter can't be const (or it would be a lie) for now.    I think the way to get there is to incorproate this this PR for now which gives us the big benefit of having the type introduced, and without changing too much for now - I thought making that change now would make the PR too large and difficult to review in one go.   But the next step can be to convert the internal string to std::wstring for windows, and fix up those windows functions using it, so that the parameter is not changed, and can be passed const.   Updating those windows specific functions will be a large number of lines changed too, which I've avoided for now - hopefully in review the change between old and new should show we are using the same string in the same way, just by accessing it inside the LocalPath wrapper class.    Sergio also asked about that in the prior review comment, and I've opened a ticket (16393) for that future improvement.

src/filesystem.cpp
```diff
@@ -856,4 +853,207 @@ bool FileInputStream::read(byte *buffer, unsigned size)
...
+
+string LocalPath::toName(const FileSystemAccess& fsaccess, FileSystemType fsType) const
+{
+    string name = localpath;
+    fsaccess.local2name(&name, fsType);
+    return name;
+}
+
+
+LocalPath LocalPath::fromPath(const string& path, const FileSystemAccess& fsaccess)
```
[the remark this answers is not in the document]
mattw-mega: sorry but I'm not seeing the difference here - is it a whitespace issue that github is not showing me?  thanks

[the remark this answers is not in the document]
dwmega: Pablo suggests the removal of an extra preceding newline.

include/mega/filesystem.h
```diff
@@ -117,6 +117,80 @@ struct MEGA_API AsyncIOContext
...
+    size_t getLeafnameByteIndex(const FileSystemAccess& fsaccess) const;
+    bool backEqual(size_t bytePos, const string& compareTo) const;
+    bool backEqual(size_t bytePos, const LocalPath& compareTo) const;
+    LocalPath subpathFrom(size_t bytePos) const;
+    string substrTo(size_t bytePos) const;
+
+    bool isContainingPathOf(const LocalPath& path, const FileSystemAccess& fsaccess);
+
+    
+    string toPath(const FileSystemAccess& fsaccess) const;
```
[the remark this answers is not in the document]
mattw-mega: Comments added, good idea, thanks

include/mega/filesystem.h
```diff
@@ -117,6 +117,80 @@ struct MEGA_API AsyncIOContext
...
+    std::string localpath;
+
+    friend class ScopedLengthRestore;
+    size_t getLength() { return localpath.size(); }
+    void setLength(size_t length) { localpath.resize(length); }
+
+public:
+
+    LocalPath() {}
+    explicit LocalPath(string&& s) : localpath(std::move(s)) {}
```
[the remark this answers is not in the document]
mattw-mega: I agree fully qualified names in headers reduce issues like that.   In this case we are inside the `mega` namespace, and we have specified that we are `using std::string;` in this namespace in types.h so just `string` is precise and will generate an error if there is any ambiguity, and I had the impression from other headers that was preferred - I'm happy with either method.  I've made those std::string in the header for this class per this request :-)  cheers

include/mega/filesystem.h
```diff
@@ -362,8 +436,8 @@ struct MEGA_API FileSystemAccess : public EventTrigger
     void unescapefsincompatible(string*,FileSystemType) const;
```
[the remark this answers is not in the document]
mattw-mega: Yes, that is exactly the sort of issue that LocalPath is meant to help with.  In this case there are a couple of things - getlocalfstype was merged from a separate branch while LocalPath was still in development on another branch, so it needs use to be very aware of what the string content we are passing it is.   And also, LocalPath depends on FileSystemAccess, and if we add a dependency in the other direction the system becomes a bit more complex, and it would be more lines changed and needing review, so I thought on balance I would not change it at this point.   Perhaps in future we should change that, but I thought that the volume of changes would be the biggest problem with this PR.   If that part is working ok then it's less risky to convert it at this point too.

src/node.cpp
```diff
@@ -1382,7 +1368,7 @@ void LocalNode::init(Sync* csync, nodetype_t ctype, LocalNode* cparent, string*
     // enable folder notification
     if (type == FOLDERNODE)
     {
-        sync->dirnotify->addnotify(this, cfullpath);
```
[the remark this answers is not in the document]
mattw-mega: Yes this is another case where the PR would be ok with fewer lines changed - in fact, for windows we ignore that parameter anyway, and for the other platforms it's utf8 in either case.    But we certainly could convert addnotify() in future, it would make good sense.

src/filesystem.cpp
```diff
@@ -856,4 +853,207 @@ bool FileInputStream::read(byte *buffer, unsigned size)
...
+}
+
+size_t LocalPath::getLeafnameByteIndex(const FileSystemAccess& fsaccess) const
+{
+    // todo: take utf8 two byte characters into account
+    size_t p = localpath.size();
+    p -= fsaccess.localseparator.size() % 2; // just in case on windows
+    while (p && (p -= fsaccess.localseparator.size()))
+    {
+        if (!memcmp(localpath.data() + p, fsaccess.localseparator.data(), fsaccess.localseparator.size()))
```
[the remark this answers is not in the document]
mattw-mega: yes, good spotting.  This is code that was moved from LocalNode::setnameparent , and I've changed it as little as possible otherwise.   Luckily there are only two cases - localseparator is 2 bytes on windows, and 1 bytes on linux.  I think it's ok to leave it as is on that basis, and because less change is less risky, what do you think?   thanks

src/filesystem.cpp
mattw-mega: I was thinking that once it was clear where all the usages of functions like this that need to know indexes into our wrapped string are, we would figure out a better public interface for LocalPath that would mean indexes are not needed, so the interface is independent of the internal representation.   In the meantime, this function should only be called in ways that the code was called previously - which is to say, with a bytePos from getLeafnameByteIndex() or lastpartlocal().    Also those two should be consolidated, but looking at them now they are slightly different...  and without wanting to complicate the PR or risk introducing subtle issues, it might be better to keep this as is in this (already very large) PR, which is performing the same string calculations as before, and make that adjustment in at future PR dedicated to it?  I'll make a note in that new PR for adjusting LocalPath.  thanks!

src/filesystem.cpp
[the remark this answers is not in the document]
mattw-mega: Ah! Thanks - in that case, it's addressed along with the comments added to document the conversions to & from std::string.  thanks :-)

src/filesystem.cpp
sergiohs84: Thanks, I appreciate you care about the amount of lines in the diff. The PR is, with no doubt, a step in the right direction, it's great you take the time to improve the readability and maintenance of the code :+1:

src/filesystem.cpp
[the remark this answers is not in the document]
sergiohs84: Agree. Let's keep it as it is. Thanks for double-checking :)

include/mega/filesystem.h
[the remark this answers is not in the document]
polmr: thanks Matt. it was a peaky unrequired change ( sorry ;-) ) only intended if we move this code elsewhere outside `mega` namespace

include/mega/filesystem.h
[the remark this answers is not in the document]
polmr: Sure, let's not overcomplicate this PR. I thought I mention that, because it sounded to me there could be a bug here (coming from elswhere, not this PR). I've contacted Javier Gonzalez, he's taking a look on that.

src/filesystem.cpp
```diff
@@ -856,4 +853,207 @@ bool FileInputStream::read(byte *buffer, unsigned size)
...
+}
+
+LocalPath LocalPath::tmpNameLocal(const FileSystemAccess& fsaccess)
+{
+    LocalPath lp;
+    fsaccess.tmpnamelocal(lp);
+    return lp;
+}
+
+bool LocalPath::isContainingPathOf(const LocalPath& path, const FileSystemAccess& fsaccess)
```
polmr, line 1038: we might want to consider adding unit test for this and others.

src/megaapi_impl.cpp
```diff
@@ -8374,34 +8354,30 @@ int MegaApiImpl::syncPathState(string* path)
     // once we do manage to lock, return to normal operation.
     syncPathStateLockTimeout = false;
 
+    LocalPath localpath = LocalPath::fromLocalname(*path);
```
polmr, line 8357: this could be avoided if !client->syncs.size():
[untested]:

```suggestion
    if (!client->syncs.size())
    {
        return state;
     }
    LocalPath localpath = LocalPath::fromLocalname(*path);
```

src/megaapi_impl.cpp
```diff
@@ -23862,36 +23805,27 @@ void MegaFolderUploadController::cancel()
...
-        while (da->dnext(&localPath, &localname, client->followsymlinks, &dirEntryType))
+        while (da->dnext(localPath, localname, client->followsymlinks, &dirEntryType))
         {
-            if (t)
-            {
-                localPath.append(client->fsaccess->localseparator);
-            }
-
-            localPath.append(localname);
+            ScopedLengthRestore restoreLen(localPath);
```
polmr, line 23821: :heart_eyes:

include/mega/filesystem.h
```diff
@@ -117,6 +117,94 @@ struct MEGA_API AsyncIOContext
...
+    std::string* editStringDirect();
+    const std::string* editStringDirect() const;
+    bool empty() const;
+    void clear() { localpath.clear(); }
+    void erase(size_t pos = 0, size_t count = std::string::npos) { localpath.erase(pos, count); }
+    void truncate(size_t bytePos) { localpath.resize(bytePos); }
+    size_t lastpartlocal(const FileSystemAccess& fsaccess) const;
+    void append(const LocalPath& additionalPath);
+    void separatorAppend(const LocalPath& additionalPath, bool separatorAlways, const std::string& localseparator);
+    void separatorPrepend(const LocalPath& additionalPath, const std::string& localseparator);
```
polmr, line 159: Suggested renaming:
```suggestion
    void appendWithSeparator(const LocalPath& additionalPath, bool separatorAlways, const std::string& localseparator);
    void prependWithSeparator(const LocalPath& additionalPath, const std::string& localseparator);
```

src/posix/fs.cpp
```diff
@@ -1203,24 +1215,23 @@ void PosixFileSystemAccess::emptydirlocal(string* name, dev_t basedev)
                  || *d->d_name != '.'
                  || (d->d_name[1] && (d->d_name[1] != '.' || d->d_name[2])))
                 {
-                    t = name->size();
-                    name->append("/");
-                    name->append(d->d_name);
+                    ScopedLengthRestore restore(name);
+                    PosixFileSystemAccess pfsa;
```
polmr, line 1219: creating and destroying this might have some performance impact: we could consider having a singleton instance used here and in `::dnext` too

polmr: (thinking specially in inotify_init1 call)

src/sync.cpp
```diff
@@ -85,57 +85,46 @@ using FingerprintLocalNodeMap = std::multimap<const LightFileFingerprint*, Local
...
+        ScopedLengthRestore restoreLength(localpath);
+        localpath.separatorAppend(localname, false, fsaccess.localseparator);
 
         // check if this record is to be ignored
-        if (app.sync_syncable(&sync, name.c_str(), &localpath))
+        if (app.sync_syncable(&sync, localname.toName(fsaccess).c_str(), localpath))
         {
             // skip the sync's debris folder
-            if (isPathSyncable(localpath, localdebris, localseparator))
+            if (!localdebris.isContainingPathOf(localpath, fsaccess))
```
polmr, line 123: :+1: !

src/sync.cpp
```diff
@@ -167,21 +156,22 @@ bool combinedFingerprint(LightFileFingerprint& ffp, const localnode_map& nodeMap
...
+bool combinedFingerprint(LightFileFingerprint& ffp, FileSystemAccess& fsaccess, const set<LocalPath>& paths)
 {
     bool success = false;
-    for (const auto& path : paths)
+    for (auto& path : paths)
     {
         auto fa = fsaccess.newfileaccess(false);
-        if (!fa->fopen(const_cast<string*>(&path), true, false))
+        auto pathArg = path; // todo: sort out const
+        if (!fa->fopen(pathArg, true, false))
```
polmr, line 166:
```suggestion
        if (!fa->fopen(const_cast<LocalPath&>(path), true, false))
```
? :thinking:

tests/integration/Sync_test.cpp
```diff
@@ -951,12 +951,8 @@ struct StandardClient : public MegaApp
...
         }
 
-        string localpath;
-        n->getlocalpath(&localpath, false);
-        ::mega::FileSystemType fileSystemType = client.fsaccess->getFilesystemType(&localpath);
-        client.fsaccess->local2name(&localpath, fileSystemType);
-        string n_localname = n->localname;
-        client.fsaccess->local2name(&n_localname, fileSystemType);
+        auto localpath = n->getLocalPath(false).toName(*client.fsaccess, FS_UNKNOWN);
+        string n_localname = n->localname.toName(*client.fsaccess, FS_UNKNOWN);
```
polmr, line 955: why removing getting file system type from local path? :thinking:

tests/unit/Sync_test.cpp
```diff
@@ -319,7 +332,7 @@ TEST(Sync, computeReverseMatchScore_twoByteSeparator)
     test_computeReversePathMatchScore("//");
 }
 
-TEST(Sync, assignFilesystemIds_whenFilesystemFingerprintsMatchLocalNodes)
+/*TEST(Sync, assignFilesystemIds_whenFilesystemFingerprintsMatchLocalNodes)
```
polmr, line 335: JIC: you can name DISABLE_xxx (or sth like that) so that tests are not executed

src/megaapi_impl.cpp
mattw-mega: good point, this is a case where we want to be very quick, responding to the file manager, especially when there are no syncs - added, thanks

include/mega/filesystem.h
mattw-mega: good idea, adjusted

src/posix/fs.cpp
mattw-mega: Yes good point.  I had checked the one for windows, which is trivial, but not this one.   In this case, we are actually in a member function of one of those, so we can just use its own localseparator :-)   adjusted.

src/sync.cpp
mattw-mega: This is one that will come right after the changes in the subsequent ticket, after which those parameters really can be const.   I think casting away const can be a bit dangerous so I'd rather avoid that, and wait for the proper fix?  thanks

tests/integration/Sync_test.cpp
mattw-mega: Ah yes - the filenames used in these tests are all simple ascii so no character conversion is needed - also, to make that call we would need a utf8 string to pass, which we don't have readily available, only the LocalPath (which for windows is utf16 of course).   So this is actually one of those cases you suspected, where we would not be getting the correct file system type anyway (on windows), due to inquiring about the wrong sort of string.   But since it won't make any difference for this test, it's easier just to pass the unknown value.

tests/unit/Sync_test.cpp
mattw-mega: Thanks - there were quite a few of these tests and they seem to very sensitive to the internals of the classes under test, I couldn't figure out how to adjust them in a short space of time.  I think ideally we would aim for the tests to work at a higher level so we don't have too many maintenance issues with them.

mattw-mega:
trigger compilation

src/posix/fs.cpp
mattw-mega: Ah, it turns out to be a static function, so I've put that back in, but outside the loop.  It's a function to delete a directory tree, so it won't be called particularly often, so one construction should be fine.  thanks

mattw-mega:
android PR build failure is due to it not being able to download sqlite: network unreachable

PR #2093, opened 2020-07-06
#16290. Check tLength before extracting thumbnail.

vtmateos:
Check all related fields of `libraw_data_t imgdata` before extracting RAW thumbnail to avoid possible crashes due to incorrect data values.

### DIFF
--- a/src/gfx/qt.cpp
+++ b/src/gfx/qt.cpp
@@ -808,9 +808,11 @@ QImageReader *GfxProcQT::readbitmapLibraw(int &w, int &h, int &orientation, QStr
               << " " << imgdata.sizes.height
               << " " << imgdata.thumbnail.twidth
               << " " << imgdata.thumbnail.theight
+              << " " << imgdata.thumbnail.tlength
               << " " << imgdata.sizes.flip;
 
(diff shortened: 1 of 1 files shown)


src/gfx/qt.cpp
```diff
@@ -808,9 +808,11 @@ QImageReader *GfxProcQT::readbitmapLibraw(int &w, int &h, int &orientation, QStr
               << " " << imgdata.sizes.height
               << " " << imgdata.thumbnail.twidth
               << " " << imgdata.thumbnail.theight
+              << " " << imgdata.thumbnail.tlength
               << " " << imgdata.sizes.flip;
 
-    if (imgdata.thumbnail.twidth > 0 && imgdata.thumbnail.theight > 0)
+    if (imgdata.thumbnail.twidth > 0 && imgdata.thumbnail.theight > 0
```
polmr, line 814: we might want to add a LOG_warn in the else.

vtmateos: I think maybe it is enough with only printing the `imgdata`, avoiding an extra log message. Thanks for the suggestion.

PR #2095, opened 2020-07-14
#16372. Have localfp set to 0 when resuming sync in macos

polmr:


sergiohs84:
The ticket ID in this PR doesn't correspond to an SDK ticket. Please, create one and assign the target version upon merge into `develop`.

PR #2097, opened 2020-07-15
#16271. Add support for new API commands (tq, bs, us)

jgandres:


src/megaapi_impl.cpp
```diff
@@ -21132,8 +21140,50 @@ void MegaApiImpl::sendPendingRequests()
...
+                }
+            }
+            else if (!strcmp(command, "bs"))
+            {
+                if (bs < -1 || bs > 2)
+                {
+                    e = API_EARGS;
+                    break;
+                }
+                if (!isBusinessAccount())
```
sergiohs84, line 21172: This check might fail if the user attempts to set the `bs` for a different user by passing its email address. I think we should simply remove it.

src/megaapi_impl.cpp
```diff
@@ -21132,8 +21140,50 @@ void MegaApiImpl::sendPendingRequests()
...
+                {
+                    e = API_EARGS;
+                    break;
+                }
+            }
+            else if (!strcmp(command, "us"))
+            {
+                if (us == 1 || (us < 0 || us > 9))
```
sergiohs84, line 21180:
```suggestion
                if (us == 1 || us < 0 || us > 9)
```

src/megaapi_impl.cpp
```diff
@@ -21132,8 +21140,50 @@ void MegaApiImpl::sendPendingRequests()
                 e = API_EARGS;
                 break;
             }
+
+            long long q = request->getTotalBytes();
+            int bs = request->getAccess();
+            int us = request->getNumDetails();
 #ifdef DEBUG
-            client->senddevcommand(command, email);
+            if (strcmp(command, "aodq")
```
sergiohs84, line 21148: Maybe, instead of comparing the strings twice, you could define `bool isOdqSubcmd = strcmp(command, "aodq");` and use it here and below :)

PR #2101, opened 2020-07-16
#16239. Do cancel all transfers. hotfix over 3.7.3

polmr:
- Store folder transfers in a separate map
- When queuing the TYPE_CANCEL_TRANSFERS, store in the last queued transfer.
- When resolving TYPE_CANCEL_TRANSFERS:
1. cancel pending transfers up to the last one queued by the time CANCEL_TRANSFERS req was created
2. cancel client->transfers
3. cancel folder-transfers
- Renew MegaFolderUp/DownloadController::cancel: it was not removing initiated subtransfers from cache!


polmr:
trigger compilation

PR #2112, opened 2020-07-21
SDK-1094. Syncs heartbeat

polmr:
- add HearBeat classes as part of the intermediate layer.
- add listeners to commands, so that we can reuse a yet unsent one, to
avoid hammering in case of recovering of a long term lock for instance
- add code to register heartbeats, update heartbeats, beat every 30
seconds and with all the expected data and unregister heartbeats too, a
- pending: filling up heartbeat data when changes occur.


src/megaapi_impl.cpp
```diff
@@ -5157,6 +5157,9 @@ void MegaApiImpl::init(MegaApi *api, const char *appKey, MegaGfxProcessor* proce
     httpio->unlock();
 #endif
 
+    mHeartBeatMonitor = ::mega::make_unique<MegaHeartBeatMonitor>(client);
+    listeners.insert(mHeartBeatMonitor.get());
```
sergiohs84, line 5161: I think this listener should be unregistered at some point. Furthermore, IMO it would be better to use the standard method to add it: `MegaApi::addListener()`. Perhaps, it could register itself if we pass `this` instead of `client`, and unregister at dtor :thinking:

src/heartbeats.cpp
```diff
@@ -0,0 +1,611 @@
...
+
+    // update values with those of the transfer
+    pending->totalBytes = transfer->getTotalBytes();
+    pending->transferredBytes = transfer->getTransferredBytes();
+
+    // reflect those in the globals
+    total += pending->totalBytes;
+    transferred += pending->transferredBytes;
+
+    setTotalBytes(total);
```
sergiohs84: Once the transfer has started (which will represent the first call to this method), does the `totalBytes` of the `transfer` changes? I would not expect so. Maybe for folder-transfers (which we probably want to discard in the heartbeats, since we have subtransfers for each.

src/heartbeats.cpp
```diff
@@ -0,0 +1,611 @@
...
+}
+
+void HeartBeatSyncInfo::updateTransferInfo(MegaTransfer *transfer)
+{
+    if (mPendingTransfers.find(transfer->getTag()) == mPendingTransfers.end())
+    {
+        mPendingTransfers.insert(std::make_pair(transfer->getTag(), ::mega::make_unique<PendingTransferInfo>()));
+    }
+
+    const auto &pending = mPendingTransfers[transfer->getTag()];
```
sergiohs84, line 86: We should avoid the double lookup in the map. It should be fast, but we know that it might become a bottleneck easily when there are a huge number of transfers already in the queue. Better to have an `if-else`. Something like:
```suggestion cpp
    const auto it = mPendingTransfers.find(transfer->getTag());
    PendingTransferInfo *pending = nullptr;
    if (it == mPendingTransfers.end())
    {
        pending = mPendingTransfers.insert(std::make_pair(transfer->getTag(), ::mega::make_unique<PendingTransferInfo>())).first->second.get();
    }
    else
        pending = it->second.get();
```

sergiohs84: I have it on a stash for this method and the `removePendingTransfer()`, let me know if you agree.

src/heartbeats.cpp
```diff
@@ -0,0 +1,611 @@
...
+{
+}
+
+void MegaHeartBeatMonitor::reset()
+{
+    mPendingBackupPuts.clear();
+    mHeartBeatedSyncs.clear();
+    mPendingSyncUpdates.clear();
+    mLastBeat = 0;
+}
```
sergiohs84: This method seems to be unused in the code (and incomplete, since it does not reset all the class members).
Should we simply remove it?

src/heartbeats.cpp
sergiohs84: Solved

src/heartbeats.cpp
sergiohs84: Solved!

src/heartbeats.cpp
```diff
@@ -0,0 +1,605 @@
...
+
+        string localFolderEncrypted(mClient->cypherTLVTextWithMasterKey("lf", sync->getLocalFolder()) );
+        string deviceIDEncrypted(mClient->cypherTLVTextWithMasterKey("de", mClient->getDeviceid()) );
+        string nameEncrypted(mClient->cypherTLVTextWithMasterKey("na", sync->getName()) );
+
+
+        if (std::find(mPendingBackupPuts.begin(), mPendingBackupPuts.end(), sync->getTag()) == mPendingBackupPuts.end())
+        {
+            pushInPending = true;
+            mClient->reqs.add(new CommandBackupPut(mClient, BackupType::TWO_WAY, sync->getMegaHandle(), localFolderEncrypted.c_str(),
```
sergiohs84, line 389: Here, the backup type should be taken from the config, which already defines the one way sync types.
I'll push a converter method from `SyncConfig::Type` to `BackupType`

src/heartbeats.cpp
```diff
@@ -0,0 +1,605 @@
...
+    string extraData;
+    if (syncID == UNDEF) //register
+    {
+
+        string localFolderEncrypted(mClient->cypherTLVTextWithMasterKey("lf", sync->getLocalFolder()) );
+        string deviceIDEncrypted(mClient->cypherTLVTextWithMasterKey("de", mClient->getDeviceid()) );
+        string nameEncrypted(mClient->cypherTLVTextWithMasterKey("na", sync->getName()) );
+
+
+        if (std::find(mPendingBackupPuts.begin(), mPendingBackupPuts.end(), sync->getTag()) == mPendingBackupPuts.end())
```
sergiohs84, line 386: I think we should not use the `syncTag` as the key for the pending API commands. Not here, and not for the pending updates (in `mPendingSyncUpdates`). A possible workaround is to use the command's tag. Another approach would be to maintain a queue of lambda functions that digest the response of those API commands.

src/heartbeats.cpp
```diff
@@ -0,0 +1,605 @@
...
+            pushInPending = true;
+            mClient->reqs.add(new CommandBackupPut(mClient, BackupType::TWO_WAY, sync->getMegaHandle(), localFolderEncrypted.c_str(),
+                                           deviceIDEncrypted.c_str(), nameEncrypted.c_str(),
+                                           getHBState(sync), getHBSubstatus(sync), getHBExtraData(sync)
+                                           ));
+        }
+        else // ID not received yet, let's queue the update (copying sync data)
+        {
+            LOG_debug << " Queuing sync update, register is on progress for sync: " << sync->getLocalFolder();
+            mPendingSyncUpdates[sync->getTag()].reset(sync->copy()); // we replace any previous pending updates
```
sergiohs84, line 397: Instead of a `MegaSync` instance, we could maintain a copy of the data that requires to be updated, ie. the backup type, the local folder, the name of the backup...
That would make it work with CU indifferently :)

src/heartbeats.cpp
```diff
@@ -0,0 +1,605 @@
...
+
+int MegaBackupMonitor::getHBState(MegaSync *sync)
+{
+    if (sync->isTemporaryDisabled())
+    {
+        return MegaBackupMonitor::State::TEMPORARY_DISABLED;
+    }
+    else if (sync->isActive())
+    {
+        //TODO: consider use case: paused, if transfers are paused!?
```
sergiohs84, line 319: Are you suggesting a new state `PAUSED`?? Perhaps it's useful. Should we ask André about APIs support?

src/heartbeats.cpp
```diff
@@ -0,0 +1,605 @@
...
+        string localFolderEncrypted(mClient->cypherTLVTextWithMasterKey("lf", sync->getLocalFolder()) );
+        string deviceIDEncrypted(mClient->cypherTLVTextWithMasterKey("de", mClient->getDeviceid()) );
+        string nameEncrypted(mClient->cypherTLVTextWithMasterKey("na", sync->getName()) );
+
+
+        if (std::find(mPendingBackupPuts.begin(), mPendingBackupPuts.end(), sync->getTag()) == mPendingBackupPuts.end())
+        {
+            pushInPending = true;
+            mClient->reqs.add(new CommandBackupPut(mClient, BackupType::TWO_WAY, sync->getMegaHandle(), localFolderEncrypted.c_str(),
+                                           deviceIDEncrypted.c_str(), nameEncrypted.c_str(),
```
sergiohs84, line 390: We should avoid to create a new string (this signature of the command's ctor receives const references to string directly). I'll fix it.

src/heartbeats.cpp
sergiohs84: Please, @polmr , check the changes to convert the backup type and, if you agree, resolve the conversation, please.

src/heartbeats.cpp
sergiohs84: Done!

src/megaapi_impl.cpp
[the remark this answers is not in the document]
polmr: I don't think we need to unregister: mHeartBeatMonitor is owned by MegaApiImp, removed upon its destruction. No callbacks will occur, no leaks. Also, `::addListener` blocks the mutex, which is unrequired here (thread not even initialized) :thinking:

src/heartbeats.cpp
polmr: Neither do I, but I'd rather not count on that: in case we ever use some fancy transfer whose size is not fixed (something streamed), I don't expect high consumption here. Your point is very valid, but I'd rather be conservative here ;-)

src/heartbeats.cpp
polmr: `updateSyncInfo` is called indiferently from `updateOrRegisterSync` and from the code that handles these mPendingSyncUpdates.
If we were to solve this elegantly we should create some BackupInfo class with a (const MegaSync &sync, handle Id = UNDEF) constructor, and refactor the code to update/register.
Note to self: updateOrRegisterSync is accepting MegaSync, we would need to move the code that looks for cached backup Id outside ...

src/heartbeats.cpp
polmr: `Paused` was stated in the specs. Although I haven't added it since there is no `pause` referred to a sync. We could infer it from transfers pauses, but I'm not 100% that's what's wanted ...

src/heartbeats.cpp
polmr: Implemented here: https://github.com/meganz/sdk/pull/2140, however, we cannot have a mPendingSyncUpdates generic for all types of backups, since it needs to be a map with sync tag as key, so as to have one only update, and we need to be able to find it in registration callback. The code in that PR is a little bit more prepared for CU backups, but we will need to keep a separate map of `mPendingCUUpdates`, and some code similar to void `updateOrRegisterSync(MegaSync *sync)` but for CUs.

src/heartbeats.cpp
polmr: used a lambda function that holds the tag

include/mega/heartbeats.h
```diff
@@ -118,12 +88,71 @@ class HeartBeatBackupInfo : public CommandListener
...
+    HeartBeatTransferProgressedInfo(handle backupId);
+    MEGA_DISABLE_COPY(HeartBeatTransferProgressedInfo)
+
+    virtual void updateTransferInfo(MegaTransfer *transfer);
+    virtual void removePendingTransfer(MegaTransfer *transfer);
+    virtual void clearFinshedTransfers();
+    virtual void setTotalBytes(long long value);
+    virtual void setTransferredBytes(long long value);
+
+    virtual double progress() const override;
```
sergiohs84, line 109:
```suggestion
    double progress() const override;
```

include/mega/heartbeats.h
```diff
@@ -118,12 +88,71 @@ class HeartBeatBackupInfo : public CommandListener
...
+        PENDING = 3, // The sync engine is working, e.g: scanning local folders
+        INACTIVE = 4, // Sync is not active. A state != ACTIVE should have been sent through '''sp'''
+        UNKNOWN = 5, // Unknown status
+    };
+
+    HeartBeatSyncInfo(int tag, handle backupId);
+    MEGA_DISABLE_COPY(HeartBeatSyncInfo)
+
+    virtual int syncTag() const;
+    virtual void updateStatus(mega::MegaClient *client) override;
```
sergiohs84, line 144:
```suggestion
    void updateStatus(mega::MegaClient *client) override;
```

include/mega/heartbeats.h
```diff
@@ -118,12 +88,71 @@ class HeartBeatBackupInfo : public CommandListener
...
+        SYNCING = 2, // The sync engine is working, transfers are in progress
+        PENDING = 3, // The sync engine is working, e.g: scanning local folders
+        INACTIVE = 4, // Sync is not active. A state != ACTIVE should have been sent through '''sp'''
+        UNKNOWN = 5, // Unknown status
+    };
+
+    HeartBeatSyncInfo(int tag, handle backupId);
+    MEGA_DISABLE_COPY(HeartBeatSyncInfo)
+
+    virtual int syncTag() const;
```
sergiohs84, line 143: Are we planning to specialize this class? Otherwise, I think this method shouldn't be `virtual`

include/mega/heartbeats.h
```diff
@@ -37,77 +37,47 @@ namespace mega
...
-    {
-        UPTODATE = 1, // Up to date: local and remote paths are in sync
-        SYNCING = 2, // The sync engine is working, transfers are in progress
-        PENDING = 3, // The sync engine is working, e.g: scanning local folders
-        INACTIVE = 4, // Sync is not active. A state != ACTIVE should have been sent through '''sp'''
-        UNKNOWN = 5, // Unknown status
-    };
-
-    HeartBeatBackupInfo(int tag, handle aBackupId);
+    HeartBeatBackupInfo(handle backupId);
```
sergiohs84, line 40: I'm missing a virtual dtor here. This class is specialized, so in case of a call `delete` on a pointer of this type, we want dtors of child classes to be called, if any. Better to have them virtual from the beginning, that to forget about it when a specialized class adds something to its dtor., IMHO

src/heartbeats.cpp
```diff
@@ -46,14 +45,14 @@ void HeartBeatBackupInfo::setRunningCommand(Command *runningCommand)
...
-HeartBeatBackupInfo::Status HeartBeatBackupInfo::status() const
+int HeartBeatBackupInfo::status() const
 {
     return mStatus;
 }
 
 double HeartBeatBackupInfo::progress() const
 {
-    return std::max(0., std::min(1., static_cast<double>(mTransferredBytes) / static_cast<double>(mTotalBytes)));
+    return 0.;
```
sergiohs84, line 55: Maybe the class can be abstract <-- methods like this could be pure virtual

include/mega/heartbeats.h
polmr: potential child class may override this one :thinking:

include/mega/heartbeats.h
polmr: same as before: virtual is harmless here, isn't it? :thinking:

include/mega/heartbeats.h
polmr: not planning, but maybe onewaysyncs could potentially lead to a child class.

include/mega/heartbeats.h
polmr: that would prevents implicit definition of the move constructor and the move assignment operator, but ok, I'll add defaults

src/heartbeats.cpp
polmr: good point. Since everything else makes HeartBeatInfo usable, I think it'll be better to have some mProgress and a setter. Just in case we don't want to have an inherited version for CU's or whatever backup may exist in the future

src/heartbeats.cpp
polmr: addresed in later commits

src/heartbeats.cpp
polmr: done

src/heartbeats.cpp
```diff
@@ -0,0 +1,780 @@
...
+{
+    string localFolderEncrypted(mClient->cypherTLVTextWithMasterKey("lf", info.localFolder()) );
+    string deviceIDEncrypted(mClient->cypherTLVTextWithMasterKey("de", mClient->getDeviceid()) );
+    string nameEncrypted(mClient->cypherTLVTextWithMasterKey("na", info.name()) );
+
+    mClient->reqs.add(new CommandBackupPut(mClient, info.backupId(), info.type(), info.megaHandle(),
+                                           localFolderEncrypted.c_str(),
+                                           deviceIDEncrypted.c_str(), nameEncrypted.c_str(),
+                                           info.state(), info.subState(), info.extra().c_str()
+                                           ));
```
polmr, line 533: DO NOT MERGE: only changed values need to be updated. Otherwise, for instance, if a backup is renamed in the backup center, it'll be overridden here

src/commands.cpp
```diff
@@ -8136,12 +8137,15 @@ CommandBackupPut::CommandBackupPut(MegaClient *client, BackupType type, handle n
...
+    if (!extraData.empty())
+        arg("e", extraData.c_str());
 
     tag = client->reqtag;
     mUpdate = false;
 }
 
+// to update an already registered backup
 CommandBackupPut::CommandBackupPut(MegaClient* client, handle backupId, BackupType type, handle nodeHandle, const char* localFolder, const char *deviceId, const char* backupName, int state, int subState, const char* extraData)
 {
```
catalinr-m, line 8150: Not related to recent commits, but I noticed that updating the state is done in the following way:
```
    if (state > 0)
    {
        arg("s", state);
    }
```
while specs say
```
"s":0,    // sync's state (0: enabled, 1: disabled)
```
which looks like it cannot go _back_ to `enabled` state.

polmr:
trigger compilation

src/heartbeats.cpp
sergiohs84: I think this comment is obsolete, isn't it?

src/commands.cpp
polmr: very valid indeed. The same should apply to substate. Nice catch @catalinr-m . I'll change to >=0

src/heartbeats.cpp
polmr: So I think. Name is no longer send through this

include/mega/heartbeats.h
polmr: Removed after some philosophical discussion. Agreed that `virtual` is unnecessary and the advantage of having it to prevent some very rare situation is outweighed by the unlikeliness of that.

sergiohs84 (approved):

sergiohs84 (approved):

PR #2113, opened 2020-07-22
#16424: fix compilation after SCSN changes

polmr:
failing to use << scsn when LOG_PERFORMANCE enabled

### DIFF
--- a/src/megaclient.cpp
+++ b/src/megaclient.cpp
@@ -4789,10 +4789,10 @@ void MegaClient::initsc()
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


sergiohs84:
Please, check the failed compilation for Win64, @polmr .

Also, take into account the issue that this PR solves is present in 3.7.2 and 3.7.3, still open. They might require a hotfix

polmr:
SDK is failing due to changes introduced by LocalPath. (unrelated to this). building `develop` would fail too I'm afraid.
hotfix is already in 3.7.3. 3.7.2 does not require: not intended for MEGAsync

PR #2118, opened 2020-07-22
Fix compilation issues macos

vtmateos:
Fix uses of LocalPath for macOS systems.

### DIFF
--- a/src/sync.cpp
+++ b/src/sync.cpp
@@ -701,13 +701,13 @@ Sync::Sync(MegaClient* cclient, SyncConfig config, const char* cdebris,
     if (macOSmajorVersion() >= 19) //macOS catalina+
     {
         LOG_debug << "macOS 10.15+ filesystem detected. Checking fseventspath.";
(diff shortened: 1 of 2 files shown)


src/sync.cpp
```diff
@@ -701,21 +701,21 @@ Sync::Sync(MegaClient* cclient, SyncConfig config, const char* cdebris,
...
     {
         LOG_debug << "macOS 10.15+ filesystem detected. Checking fseventspath.";
-        string supercrootpath = "/System/Volumes/Data" + crootpath;
+        string supercrootpath = "/System/Volumes/Data" + *crootpath.editStringDirect();
 
         int fd = open(supercrootpath.c_str(), O_RDONLY);
         if (fd == -1)
         {
             LOG_debug << "Unable to open path using fseventspath.";
-            mFsEventsPath = crootpath;
```
mattw-mega, line 710: Looks good, the only thing I would say is that it might be convenient to make mFsEventsPath a LocalPath instead of a string, but this is fine for now too.  Thanks for fixing this up.

PR #2129, opened 2020-07-27
#16413. Fix upload nonresponseiveness due to checking filesystem typeComparator

mattw-mega:
For a large folder upload, minimize checking of which filesystem we are working in.  Just once per folder.
Took advantage of the new LocalPath to simplify whether the path we are checking is utf8 or utf16 (for windows)
getlocalfstype and getFilesystemType work on LocalPath now, which means fewer switches between utf8/utf16
that means we can use LocalPath functions and don't need getValidPath anymore
Upload functions in megaApi_impl take a FS parameter, for cases where we already know.  If it is unknown then the FS is looked up.

### DIFF
--- a/include/mega/filesystem.h
+++ b/include/mega/filesystem.h
@@ -161,6 +161,8 @@ public:
     bool findNextSeparator(size_t& separatorBytePos, const FileSystemAccess& fsaccess) const;
     bool findPrevSeparator(size_t& separatorBytePos, const FileSystemAccess& fsaccess) const;
     bool endsInSeparator(const FileSystemAccess& fsaccess) const;
+
+    // get the index of the leaf name.  A trailing / is considered part of the leaf.
     size_t getLeafnameByteIndex(const FileSystemAccess& fsaccess) const;
     bool backEqual(size_t bytePos, const LocalPath& compareTo) const;
     LocalPath subpathFrom(size_t bytePos) const;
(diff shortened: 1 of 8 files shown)


include/mega/filesystem.h
```diff
@@ -160,11 +160,16 @@ class MEGA_API LocalPath
     void trimNonDriveTrailingSeparator(const FileSystemAccess& fsaccess);
     bool findNextSeparator(size_t& separatorBytePos, const FileSystemAccess& fsaccess) const;
     bool findPrevSeparator(size_t& separatorBytePos, const FileSystemAccess& fsaccess) const;
+    bool endsInSeparator(const FileSystemAccess& fsaccess) const;
+
+    // get the index of the leaf name.  A trailing / is considered part of the leaf.
```
polmr, line 165:
```suggestion
    // get the index of the leaf name.  A trailing separator is considered part of the leaf.
```
Not worth a commit ;-)

include/megaapi_impl.h
```diff
@@ -1843,8 +1843,8 @@ struct MegaFileGet : public MegaFile
     void progress() override;
     void completed(Transfer*, LocalNode*) override;
     void terminated() override;
-	MegaFileGet(MegaClient *client, Node* n, string dstPath);
-    MegaFileGet(MegaClient *client, MegaNode* n, string dstPath);
+	MegaFileGet(MegaClient *client, Node* n, const LocalPath& dstPath, FileSystemType fsType);
```
polmr, line 1846:
```suggestion
    MegaFileGet(MegaClient *client, Node* n, const LocalPath& dstPath, FileSystemType fsType);
```

PR #2133, opened 2020-07-28
small fix for transfer tag, make it always 'long long', never 'int'

sergiohs84:


aabellagm:
trigger compilation android

aabellagm:
trigger compilation android

PR #2140, opened 2020-07-30
SDK-1094. Add encapsulation of backup info to register

polmr:
- move some sync specific static methods from monitor to new subclass
MegaBackupInfoSync

### DIFF
--- a/src/heartbeats.cpp
+++ b/src/heartbeats.cpp
@@ -609,17 +609,17 @@ std::shared_ptr<HeartBeatTransferProgressedInfo> MegaBackupMonitor::getHeartBeat
             Node *n = mClient->nodebyhandle(transfer->getType() == MegaTransfer::TYPE_UPLOAD ? transfer->getParentHandle() : transfer->getNodeHandle());
             while (n)
             {
-                if (n && n->localnode && n->localnode->sync)
+                if (n->localnode && n->localnode->sync)
                 {
                     syncTag = n->localnode->sync->tag;
                     mTransferToSyncMap[transfer->getTag()] = syncTag;
                     break;
                 }
-                LOG_warn << "Heartbeat could not get sync tag direclty from transfer handle. Going up";
                 n = n->parent;
             }
         }
 
+        assert(syncTag && "couldn't get syncTag from sync transfer");
         if (syncTag)
         {
             auto hBPair = mHeartBeatedSyncs.find(syncTag);
--- a/src/heartbeats.cpp
+++ b/src/heartbeats.cpp
@@ -288,6 +288,125 @@ void HeartBeatSyncInfo::updateStatus(MegaClient *client)
 }
 
 
+
+////////////// BackupInfo ////////////////
+
+MegaBackupInfo::MegaBackupInfo(BackupType type, string localFolder, string name, handle megaHandle, int state, int substate, std::string extra, handle backupId)
+    : mType(type), mLocalFolder(localFolder), mName(name), mMegaHandle(megaHandle),
+      mState(state), mSubState(substate), mExtra(extra), mBackupId(backupId)
+{
+
+}
+
+BackupType MegaBackupInfo::type() const
+{
+    return mType;
+}
+
+handle MegaBackupInfo::backupId() const
+{
+    return mBackupId;
+}
+
+string MegaBackupInfo::localFolder() const
+{
+    return mLocalFolder;
+}
+
+string MegaBackupInfo::name() const
+{
+    return mName;
+}
+
+handle MegaBackupInfo::megaHandle() const
+{
+    return mMegaHandle;
+}
+
+int MegaBackupInfo::state() const
+{
+    return mState;
+}
+
+int MegaBackupInfo::subState() const
+{
+    return mSubState;
+}
+
+string MegaBackupInfo::extra() const
+{
+    return mExtra;
+}
+
+void MegaBackupInfo::setBackupId(const handle &backupId)
+{
+    mBackupId = backupId;
+}
+
+MegaBackupInfoSync::MegaBackupInfoSync(MegaClient *client, const MegaSync &sync, handle backupid)
+    : MegaBackupInfo(getSyncType(client, sync), sync.getLocalFolder(), sync.getName(), sync.getMegaHandle()
+                 , getSyncState(sync), getSyncSubstatus(sync), getSyncExtraData(sync), backupid)
+{
+
+
+}
+int MegaBackupInfoSync::getSyncState(const MegaSync &sync)
+{
+    if (sync.isTemporaryDisabled())
+    {
+        return State::TEMPORARY_DISABLED;
+    }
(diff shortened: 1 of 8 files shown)


polmr:
trigger compilation

src/heartbeats.cpp
```diff
@@ -494,52 +574,65 @@ void MegaBackupMonitor::onSyncStateChanged(MegaApi* /*api*/, MegaSync *sync)
...
-    int syncTag = 0;
+        mClient->reqs.add(new CommandBackupRemove(mClient, hBPair->second->backupId()));
 
-    // use map to get the syncTag directly if there was one
-    auto mTSPair = mTransferToSyncMap.find(transfer->getTag());
-    if (mTSPair != mTransferToSyncMap.end())
-    {
-        syncTag = mTSPair->second;
+        mHeartBeatedSyncs.erase(hBPair); //This is speculative: could be moved to backupremove_result
+        // in case we wanted to handle possible failing cases.
```
sergiohs84, line 585: It would be an edge case, but... if we speculative remove it and the API doesn't process the removal, it will create an inconsistent state where the API returns a backup to the backup centre, but no client is heartbeating it.
I think it would be better to not do it speculative, not because of failures (which I think won't happen), but for a proper management of cases where the client is offline, in example, and gets closed before sending out the command.

src/heartbeats.cpp
```diff
@@ -494,52 +574,65 @@ void MegaBackupMonitor::onSyncStateChanged(MegaApi* /*api*/, MegaSync *sync)
...
+{
+#ifdef ENABLE_SYNC
+    if (transfer->isSyncTransfer())
     {
-        Node *n = mClient->nodebyhandle(transfer->getType() == MegaTransfer::TYPE_UPLOAD ? transfer->getParentHandle() : transfer->getNodeHandle());
-        while (n)
+        int syncTag = 0;
+
+        // use map to get the syncTag directly if there was one
+        auto mTSPair = mTransferToSyncMap.find(transfer->getTag());
```
sergiohs84, line 599: `m` prefix is for members of a class. I'll adjust it.

src/heartbeats.cpp
```diff
@@ -494,52 +574,65 @@ void MegaBackupMonitor::onSyncStateChanged(MegaApi* /*api*/, MegaSync *sync)
...
-    {
-        auto hBPair = mHeartBeatedSyncs.find(syncTag);
-        if (hBPair != mHeartBeatedSyncs.end()) //transfer info arrived before having created the HeartBeatSyncInfo object
+        if (!syncTag) //first time
         {
-            return hBPair->second;
+            Node *n = mClient->nodebyhandle(transfer->getType() == MegaTransfer::TYPE_UPLOAD ? transfer->getParentHandle() : transfer->getNodeHandle());
+            while (n)
+            {
+                if (n && n->localnode && n->localnode->sync)
```
sergiohs84, line 610:
```suggestion
                if (n->localnode && n->localnode->sync)
```
`n` is already checked in the `while()`

src/heartbeats.cpp
```diff
@@ -494,52 +574,65 @@ void MegaBackupMonitor::onSyncStateChanged(MegaApi* /*api*/, MegaSync *sync)
...
+                {
+                    syncTag = n->localnode->sync->tag;
+                    mTransferToSyncMap[transfer->getTag()] = syncTag;
+                    break;
+                }
+                LOG_warn << "Heartbeat could not get sync tag direclty from transfer handle. Going up";
```
sergiohs84, line 616: In case of a transfer very deep in the subfolders of the sync, I think the message could be logged too many times (one per level going up to the root of the sync). Maybe we should avoid it?

src/heartbeats.cpp
```diff
@@ -494,52 +574,65 @@ void MegaBackupMonitor::onSyncStateChanged(MegaApi* /*api*/, MegaSync *sync)
...
+            auto hBPair = mHeartBeatedSyncs.find(syncTag);
+            if (hBPair != mHeartBeatedSyncs.end()) //transfer info arrived before having created the HeartBeatSyncInfo object
+            {
+                return hBPair->second;
+            }
+            else
+            {
+                //create new HeartBeatSyncInfo
+                return mHeartBeatedSyncs.insert(std::make_pair(syncTag, std::make_shared<HeartBeatSyncInfo>(syncTag, UNDEF))).first->second;
+            }
```
sergiohs84, line 632: `if (!syncTag)` at this point, it would be an error that we may want to catch. Perhaps an `assert(syncTag)` before the `if (syncTag)`?

src/heartbeats.cpp
```diff
@@ -287,13 +288,134 @@ void HeartBeatSyncInfo::updateStatus(MegaClient *client)
...
+}
+int MegaBackupInfoSync::getSyncState(const MegaSync &sync)
+{
+    if (sync.isTemporaryDisabled())
+    {
+        return State::TEMPORARY_DISABLED;
+    }
+    else if (sync.isActive())
+    {
+        //TODO: consider use case: paused, if transfers are paused!?
```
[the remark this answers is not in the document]
polmr: no, the note indicates that we might want to consider adding a new substate of ACTIVE. Current code will beat as ACTIVE a sync, even if transfers are paused, which might be good enough (or not)

src/heartbeats.cpp
polmr: I see. The spec does mention that the apps should try to "do their best to inform the API if a sync is deleted/permanently removed". I wonder if that example of yours fits that description. There could be some rules in the API that can easily remove syncs after 30 mins of clients that are otherwise alive.. Anyway, having the sync removed from mHeartBeatedSyncs speculatively or only upon successful API return, won't fix the proposed use case. After `onSyncDeleted` callbacks are called, the SDK no longer knows about the sync. Fixing that will entail caching some `to_be_deleted_backups`, or caching some `unresolved_requests`. Doable, but would it be worth it?

src/heartbeats.cpp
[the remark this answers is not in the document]
polmr: added commit to have pause state and have it updated when pause state changes

src/heartbeats.cpp
sergiohs84: Definitely not worthy for an edge case like this one. Let's assume that the inconsistency might happen in an adverse scenario. In the end, the same thing can happen when the user uninstall the app without the change to unregister the sync, so the Backup Centre UI will need to consider those orphaned backups at some point and provide a way to get rid of them.
Thanks

sergiohs84 (approved):
Hi @polmr . I've added a trivial commit to remove unused params from the new callbacks (plus minor refactoring of loop over syncs). Please, have a look, and if it all looks fine, proceed to merge. Thanks

polmr:
Fair enough. I added the parameters in case we want to do something with those in the future. But let's keep it simple for the moment ;-)

PR #2143, opened 2020-07-31
Avoid one extra fopen() per Sync::scan() when not initializing

mattw-mega:
Also fixed a couple of warnings building 32 bit

### DIFF
--- a/include/mega/filter.h
+++ b/include/mega/filter.h
@@ -1,0 +1,190 @@
+/**
+ * @file mega/filter.h
+ * @brief Classes representing file filters.
+ *
+ * (c) 2013-2014 by Mega Limited, Auckland, New Zealand
+ *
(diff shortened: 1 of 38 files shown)


include/mega/filter.h
```diff
@@ -0,0 +1,190 @@
...
+{
+    FT_ALL,
+    FT_DIRECTORIES,
+    FT_FILES,
+    NUM_FILTER_TARGETS
+}; /* FilterTarget */
+
+enum FilterType
+{
+    FT_NAME,
```
sergiohs84, line 50: Should we find a different prefix, so it doesn't clash with the prefix for values in `FilterTarget`?

dwmega: This branch / PR can be removed.

The changes that could be integrated, have been.

feature/ignorefile has changed a fair bit since this PR was created.
These enumerations no longer exist, for example.

PR #2145, opened 2020-07-31
#16593. Update 3rdParty Windows compilation

catalinr-m:


contrib/cmake/vcpkg_extra_triplets/x64-windows-mega.cmake
```diff
@@ -1,4 +1,12 @@
-set(VCPKG_VISUAL_STUDIO_PATH "C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\Professional")
+if(EXISTS "C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\Professional")
+    set(VCPKG_VISUAL_STUDIO_PATH "C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\Professional")
+elseif(EXISTS "C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\Community")
+    message(AUTHOR_WARNING "Microsoft Visual Studio 2017 Community will be used. Professional would be more appropriate.")
+    set(VCPKG_VISUAL_STUDIO_PATH "C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\Community")
+else()
+    message(FATAL_ERROR "Microsoft Visual Studio 2017 could not be found")
+endif()
+
```
catalinr-m, line 9: Should this be applied to `x64-windows-mega-staticdev.cmake` and `x86-windows-mega.cmake` too?

mattw-mega: Good idea, yes please

catalinr-m: done

catalinr-m:
OK, I did not realize that versions could differ. But does this only apply to `opengl`? I see now that `3rdParty\vcpkg\ports_moved\opengl\CONTROL` file contains
```
Source: opengl
Version: 0.0-7
```
...and file `portfile.cmake` there has a different fix for the opengl issue. 

But then how exactly should this be handled?
If `0.0-5` and `0.0-7` should coexist, then maybe current changes can be left for the former, then add a new folder for the latter with the original unchanged files (and maybe name it `latest`, so that any newer version fetched by vcpkg will not be confusingly handled using files from a place named after an older version?) .
Else if only the latest version should be present, then `0.0-5` folder should be renamed, and its files replaced with the latest version.
Either way, I imagine `sdk\contrib\cmake\preferred-ports.txt` is the one needing update. Are there any other places?

p.s. (and maybe not the right place to ask) Is there a point in using `preferred-ports` with specific versions, if the files there are used with whatever vcpkg fetches? Sorry if I'm missing something obvious.

mattw-mega:
Yes good questions, I'll start a discussion

dwmega:
Hey @catalinr-m,

I'd delete the 0.0-5 directory under vcpkg_extra_ports/opengl and create a new directory, 0.0-7.
This directory would contain the contents of ports/opengl from vcpkg.

You'd then update the opengl entry in the preferred-ports.txt file to pin the version to 0.0-7.

As far as I understand our dependencies, we don't need 0.0-5 and 0.0-7 to coexist.

The decision whether we accept any version of some dependency should be made by someone who has a deeper understanding of our build process, particularly with respect to other platforms we target.

The main reason we have vcpkg_extra_ports is to ensure that we always have "safe versions" of our dependencies. That is, we have verified that we can build against these particular versions. This is important for libraries such as ffmpeg whose external APIs change often.

catalinr-m:
Looks like there was no version change after all.
Looking again at `3rdParty\vcpkg\packages\opengl_x64-windows-mega\CONTROL`, it does say
```
Package: opengl
Version: 0.0-5
Architecture: x64-windows-mega
```
so the previous `3rdParty\vcpkg\ports_moved\opengl\CONTROL` was the wrong one to look at.

I guess the current changes can be merged now.

(Should newer versions be used, can probably be done in a separate task.)

PR #2149, opened 2020-08-04
Fix/tcprelay on linux

dwmega:
Various tweaks to get tcprelay building and working on Linux and Windows.


tests/tool/tcprelay/main.cpp
```diff
@@ -641,10 +669,94 @@ MegaCLILogger logger;
...
+                s.erase(j, string::npos);
+                break;
+            }
+        }
+    }
+    return strdup(s.c_str());
+}
+
+char** my_rl_completion(const char */*text*/, int /*start*/, int end)
+{
```
mattw-mega, line 695: Can we add a space between the first */ characters please?   MSVC complains about end-comment outside of a comment, and the line is showing up red here in the github compare screen.  thanks

dwmega: Sure thing.

Since they're not being used, I'll just leave out the names entirely.

PR #2150, opened 2020-08-04
Update ReadeMe.md with the CMake info

ms-mega:


README.md
```diff
@@ -145,7 +145,7 @@ This should build 3rdParty dependencies.
 Once the 3rdParty dependencies are built, you can start building your SDK as follows:
 * cd contrib\cmake
 * mkdir cmake-build-x64 (folder names matching pattern cmake-build-* in contrib/cmake are ignored by git)
-* Download and install CMake from https://cmake.org/download/
+* Download and install CMake from https://cmake.org/download/. Mininimum required version is 3.7.2
```
mattw-mega, line 148: MEGAsync on `develop` requires 3.15, probably we should recommend that one

ms-mega: That's right. MEGASync on develop requires minimum 3.15.
However, SDK - that this ReadMe.md targets - has: 
`cmake_minimum_required(VERSION 3.7.2)`

mattw-mega: Yes, I think I copy-pasted that from some example a long time ago, and it's never been updated, and probably isn't correct.   thanks for the update :-)

PR #2152, opened 2020-08-06
Added ignoreAttributes parameter to FileAccess:fopen(...).

dwmega:
fopen(...) will fail to open files on Windows possessing certain
attributes. These attributes are used by Windows to mark files as being
hidden or as having special significance to the system.

This new flag gives the caller control over whether such files are
rejected and is intended to be used when verifying the presence of a
sync root.


polmr:
trigger compilation

PR #2159, opened 2020-08-10
Filesystem Tweaks

dwmega:
The purpose of this changeset is twofold:
1. Fix a bug where we determine the filesystem of a directory's parent rather than itself.
2. Reorganize the code a little to make things a little cleaner.

The bug fixed by 1. occurs when we try to determine the filesystem type of a path that doesn't end up in a directory separator. Current code truncates the leaf of the path and returns the filesystem type of the resulting path. On Linux, this can result in the wrong filesystem type being returned.

Say our sync root is /mnt/zug. We ask for /mnt/zug's filesystem type but return /mnt's. You can see the problem.

What we do now is check the filesystem type of the path as given. If this fails, we try to determine the filesystem type of path's parent. We'll usually perform a single stat as the path will exist. At worst, we perform two stat calls.

The changes introduced by 2 could be considered aesthetic. To me, it seemed worthwhile moving the POSIX and Windows parts of the code into posix/fs.cpp and win32/fs.cpp respectively. This reduces the use of preprocessor conditionals.

### DIFF
--- a/src/filesystem.cpp
+++ b/src/filesystem.cpp
@@ -70,6 +70,8 @@ const char *FileSystemAccess::fstypetostring(FileSystemType type) const
             return "SDCARDFS";
         case FS_F2FS:
             return "F2FS";
+        case FS_XFS:
+            return "XFS";
         case FS_UNKNOWN:    // fall through
             return "UNKNOWN FS";
     }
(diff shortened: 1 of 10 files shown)


src/filesystem.cpp
```diff
@@ -286,7 +207,7 @@ void FileSystemAccess::escapefsincompatible(string* name, FileSystemType fileSys
             name->replace(i, 1, buf);
             LOG_debug << "Escape incompatible character for filesystem type "
                       << fstypetostring(fileSystemType)
-                      << ", replace '" << std::string(&incompatibleChar, 1) << "' by '" << buf << "'\n";
+                      << ", replace '" << incompatibleChar << "' by '" << buf << "'\n";
```
jgandres, line 210: With this change the macro LOG_debug, will print numerical value of incompatibleChar, so you need to preserve the std::string if you want to print the character.

PR #2170, opened 2020-08-12
#SDK-1178. Avoid duplicated separator (hotfix for develop)

sergiohs84:


sergiohs84:
> this one seems to include the escape/unescape characters change for the release branch, which probably shouldn't be merged to develop?

Actually, the escape/unescape is something debatable to be included in `develop`. In the end, if we honor the Git Flow, it's a hotfix that should become part of `develop` for the shake of correctness and for a clean version management. The ticket to restore the escaping based on filesystem already mentions the commit that should be reverted as part of the task, so I think it's safe to merge that hotfix :)

PR #2173, opened 2020-08-13
SDK-1180: Fix waiter notifications on MacOS

mattw-mega:
The pipe() call was setting the read file descriptor to 3 on many calls
Possibly the first call on each thread
If the 3 file descriptor was closed, the next pipe() call would deliver 3 again
And ultimately one waiter could steal the notifies of another by reading the byte from the pipe
Which would leave that other waiter stalled.

### DIFF
--- a/src/posix/waiter.cpp
+++ b/src/posix/waiter.cpp
@@ -26,11 +26,20 @@ dstime Waiter::ds;
 
 PosixWaiter::PosixWaiter()
 {
-    // pipe to be able to leave the select() call
-    if (pipe(m_pipe) < 0)
+    for (;;)
     {
-        LOG_fatal << "Error creating pipe";
-        throw std::runtime_error("Error creating pipe");
+        // pipe to be able to leave the select() call
+        if (pipe(m_pipe) < 0)
+        {
+            LOG_fatal << "Error creating pipe";
+            throw std::runtime_error("Error creating pipe");
+        }
+
+        // On MacOS, pipe() keeps giving us 3 as the first fd.  Possibly on the first pipe() call per thread?
+        if (m_pipe[0] > 3 && m_pipe[1] > 3) break;
+
+        if (m_pipe[0] > 3) close(m_pipe[0]);
+        if (m_pipe[1] > 3) close(m_pipe[1]);
     }
 
     if (fcntl(m_pipe[0], F_SETFL, O_NONBLOCK) < 0)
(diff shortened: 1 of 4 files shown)


src/posix/waiter.cpp
```diff
@@ -26,11 +26,20 @@ dstime Waiter::ds;
...
-        throw std::runtime_error("Error creating pipe");
+        // pipe to be able to leave the select() call
+        if (pipe(m_pipe) < 0)
+        {
+            LOG_fatal << "Error creating pipe";
+            throw std::runtime_error("Error creating pipe");
+        }
+
+        // On MacOS, pipe() keeps giving us 3 as the first fd.  Possibly on the first pipe() call per thread?
+        if (m_pipe[0] > 3 && m_pipe[1] > 3) break;
```
sergiohs84, line 39: With more than 2 threads in place, we could hit the case where thread1 gets the fd=3 and both thread2 and thread3 gets the fd=4 <-- issue not solved, but depending on the race condition

Aasx-12341234:
https://github.com/meganz/sdk/pull/2173/files

src/posix/waiter.cpp
```diff
@@ -116,13 +125,16 @@ int PosixWaiter::wait()
     uint8_t buf;
     bool external = false;
 
+    while (mAtomicBytesSent > 0)
     {
-        std::lock_guard<std::mutex> g(mMutex);
-        while (read(m_pipe[0], &buf, sizeof buf) > 0)
+        auto ret = read(m_pipe[0], &buf, 1);
```
sergiohs84, line 130: Nice one :wink:

src/posix/waiter.cpp
```diff
@@ -26,11 +26,20 @@ dstime Waiter::ds;
...
+        {
+            LOG_fatal << "Error creating pipe";
+            throw std::runtime_error("Error creating pipe");
+        }
+
+        // On MacOS, pipe() keeps giving us 3 as the first fd.  Possibly on the first pipe() call per thread?
+        if (m_pipe[0] > 3 && m_pipe[1] > 3) break;
+
+        if (m_pipe[0] > 3) close(m_pipe[0]);
+        if (m_pipe[1] > 3) close(m_pipe[1]);
```
sergiohs84, line 42: :thinking: I feel we could leak the `fd=3` and keep it opened forever. If first call to pipe returns 3 and 4, only the 4 will be closed. Is it the purpose? so you don't get the `3` anymore in subsequent calls?
Looks like a nasty bug to return the same `fd` in result of different calls (without closing it before).

mattw-mega: Yes the intention was to leak the fd 3 so that we don't accidentally reuse it - however we have a much better explanation and fix now, so we can abandon this PR (for now at least, unless some other issue arises with the waiter on Mac)  thanks

sergiohs84:
Closed in pro of https://github.com/meganz/sdk/pull/2179

PR #2186, opened 2020-08-17
SDK-1187. Uploads stalling on windows occasionally with many transfers in progress

mattw-mega:


src/posix/net.cpp
```diff
@@ -524,6 +524,7 @@ void CurlHttpIO::addaresevents(Waiter *waiter)
 #if defined(_WIN32)
     for (auto& mapPair : prevAressockets)
     {
+        // todo: has the socket always been closed at this point anyway? (and, any chance the socket number might already be being reused?
```
sergiohs84, line 527: Let's check it before merging the hotfix, please. It looks like something risky.

src/posix/net.cpp
```diff
@@ -2526,17 +2540,37 @@ int CurlHttpIO::socket_callback(CURL *, curl_socket_t s, int what, void *userp,
...
+        if (it == socketmap.end())
+        {
+            LOG_debug << "Adding curl socket " << s << " to " << what;
+#ifdef WIN32
+            auto it_bool = socketmap.emplace(s, SockInfo(httpio->mSocketsWaitEvent));
+#else
+            auto it_bool = socketmap.emplace(s, SockInfo());
+#endif
+            it = it_bool.first;
+        }
```
sergiohs84, line 2567:
```suggestion cpp
        if (it == socketmap.end())
        {
            LOG_debug << "Adding curl socket " << s << " to " << what;
#ifdef WIN32
            it = socketmap.emplace(s, SockInfo(httpio->mSocketsWaitEvent)).first;
#else
            it = socketmap.emplace(s, SockInfo()).first;
#endif
        }
```

src/win32/waiter.cpp
```diff
@@ -92,41 +92,50 @@ void Waiter::bumpds()
...
+        {
+            r |= flags[dwWaitResult - WAIT_OBJECT_0];
+        }
     }
-    if ((dwWaitResult >= WAIT_OBJECT_0) && (dwWaitResult < WAIT_OBJECT_0 + flags.size()))
+    else
     {
-        r |= flags[dwWaitResult - WAIT_OBJECT_0];
+        assert(false); // alert developers that we're hitting the limit
+        r |= NEEDEXEC;
```
sergiohs84, line 129: If we need to wait, but we immediately tell the Waiter that we need execution.. :thinking: it would be risky, IMO. We will do a loop, no event may actually need execution, so we loop for nothing and next call to `wait()` will again `NEEDEXEC`. Additionally, in the case that one event has completed, the line above will not avoid the missing notification when the new event to wait for completes (so we end up in the same situation that the current observed issue). 

Please, note the above is how I understand it. Of course, I might be wrong, don't see the full picture in here... so please, let me know if I'm misunderstanding how it works. Thanks in advance.

On the other hand, now that we are not adding so many "wait objects" thanks to reusing the same one to wait for multiple sockets (cUrl HANDLEs), it should not be a problem anymore (or at least hard to hit the limit). So I kind of agree it's a pragmatic workaround, specially with the `assert(false)` in there :smiley:

src/win32/waiter.cpp
```diff
@@ -135,6 +144,15 @@ int WinWaiter::wait()
 bool WinWaiter::addhandle(HANDLE handle, int flag)
 {
     assert(handles.size() == flags.size() && handles.size() >= index);
+
+#ifdef DEBUG
+    for (unsigned i = index; i--; )
+    {
+        // double check we only add one of each handle
+        assert(handles[i] != handle);
+    }
```
sergiohs84, line 153: Perhaps `foreach()` is more appropriate for these cases. But it probably depends on the each person :wink:

src/posix/net.cpp
```diff
@@ -486,7 +473,23 @@ void CurlHttpIO::addaresevents(Waiter *waiter)
...
+#endif
+                it = it_bool.first;
+            }
+            else
+            {
+                auto it_bool = aressockets.emplace(socks[i], std::move(it->second));
+                prevAressockets.erase(it);
+                it = it_bool.first;
+            }
+            SockInfo& info = it->second;
```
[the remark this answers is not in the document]
mattw-mega: Although `it` started out as an iterator into the `prev` map of sockets, it ends up as an iterator into the current sockets map. so `itPrev` is a bit misleading.   I'll adjust the name to something else

src/win32/waiter.cpp
mattw-mega: Yes, the idea here is that this code path should not be hit, but if we expand handle usage such that we start hitting the limit again, developers will notice first in testing due to the assert(), and at least we won't suffer from mysterious pauses.   If it does actually happen, then while we have > 64 handles, we will be in a busy loop, consuming one core, but again without the mysterious pauses, and the busy core would be something less mysterious for devs to investigate.   Once the number of handles falls below 64 again, the busy loop goes back to normal.

src/win32/waiter.cpp
mattw-mega: It is a vector, but for performance (because it kept showing up in the profiler, it was measured and not just guesswork), we don't keep modifying the size of the vector  (even just changing the size with resize() and push_back()).  The size of the vector is the maximum number of handles we've ever waited on.    With this PR that number will be a lot smaller, and perhaps not such a performance issue, but having already investigated and fixed that I'd like to keep the benefit.   So we just use index to keep track of how many of the vector's entries we have filled in for this upcoming wait.   So if we used `for(auto h: handles)` , we would iterate over too many entries and possibly get the wrong answer - I do prefer that form too, when appropriate.  thanks

sergiohs84: Oh, now I see why to use the `index`. I should have checked how it's set in advance. My apologies, since it makes perfect sense :wink:

src/posix/net.cpp
mattw-mega: Ok, this one is checked and adjusted - since c-ares doesn't inform us about socket creations or closures, but only about the set currently in use, we'll leave sockets other that those untouched.   If any of those sockets do set the event, it's only one extra loop.    While looking into this, I've improved the logging of cURL considerably as you'll see in the subsequent changes - and come across a separate issue that I'll make a ticket for.   thanks

sergiohs84 (approved):

PR #2187, opened 2020-08-17
SDK-1185:Add an API to upload files with top priority and custom filenames

ms-mega:
Add an API to upload files with top priority and custom filenames


Gnzlt (approved):
Works as expected 👍

sergiohs84 (approved):

PR #2193, opened 2020-08-19
Fix/curl upload chunk spurious windows timeout and windows waiter handle limit fix

mattw-mega:
merge of the corresponding hotfix branch to develop


sergiohs84 (approved):
We usually apply hotfixes to develop by merging the original branch of the hotfix (the same one that targets the corresponding release). Sometimes it's not possible due to conflicts, though.
In this case, given you already prepared this branch with both hotfixes, let's keep it as it is and simply merge it ;)

PR #2197, opened 2020-08-20
#SNC-1786. Allow copySyncDataToCache to receive sync error too

polmr:
This is useful to be able to restore sync configurations from previous sessions, not only 
to pass app sync cache to sdk's cache

- also do not override sync error for remote not found when already
failed


polmr:
Closed in favor of https://github.com/meganz/sdk/pull/2209 (recovering syncs from previous sessions will be done by the SDK from now on)

PR #2198, opened 2020-08-21
SDK-1202. Initial targeted syncdown()/syncup() support

mattw-mega:
Add the ability to mark a LocalNode for targeted syncup or syncdown.  
Defer making a syncup/syncdown decision about a node if there are already commands being sent to the server affecting that node (this will be very important for Speculative Instant Completion Removal)
Moved some intermediate layer flag setting back to the sync core where it should have been.

### DIFF
--- a/include/mega/node.h
+++ b/include/mega/node.h
@@ -123,7 +123,10 @@ struct CommandChain
 
     void push_back(Command* c)
     {
-        if (!chain) chain.reset(new std::list<Command*>);
+        if (!chain)
+        {
+            chain.reset(new std::list<Command*>);
+        }
         chain->push_back(c);
     }
 
@@ -377,9 +380,11 @@ struct MEGA_API LocalNode : public File
     // global sync reference
     handle syncid = mega::UNDEF;
 
-    enum { synctree_resolved = 0, 
-           synctree_descendantflagged = 1, 
-           synctree_scanhere = 2 };  // scan here also implies checking immdiate children for synctree_descendantflagged
+    enum : unsigned
+    {
+        SYNCTREE_RESOLVED = 0,
+        SYNCTREE_DESCENDANT_FLAGGED = 1,
+        SYNCTREE_SCAN_HERE = 2 };  // scan here also implies checking immdiate children for synctree_descendantflagged
 
     struct
     {
@@ -396,11 +401,10 @@ struct MEGA_API LocalNode : public File
         bool checked : 1;
 
         // needs another syncdown after pending changes
-        unsigned syncdownTargetedAction: 2;
+        unsigned syncdownTargetedAction : 2;
 
         // at least one child has needsAnotherSyncdown set
-        unsigned syncupTargetedAction: 2;
-
+        unsigned syncupTargetedAction : 2;
     };
 
     // set the syncupTargetedAction for this, and parents
--- a/include/mega/node.h
+++ b/include/mega/node.h
@@ -403,10 +403,10 @@ struct MEGA_API LocalNode : public File
 
     };
 
-    // set the syncupPartialAction for this, and parents
+    // set the syncupTargetedAction for this, and parents
     void needsFutureSyncup();
 
-    // set the syncupPartialAction for this, and parents
+    // set the syncdownTargetedAction for this, and parents
     void needsFutureSyncdown();
 
     // current subtree sync state: current and displayed
--- a/include/mega/node.h
+++ b/include/mega/node.h
@@ -114,9 +114,9 @@ private:
 };
 
 struct CommandChain
-{   
+{
     // convenience functions, hides the unique_ptr aspect, removes it when empty
-    bool empty() 
+    bool empty()
     {
         return !chain || chain->empty();
     }
@@ -403,7 +403,7 @@ struct MEGA_API LocalNode : public File
         // needs another syncdown after pending changes
         unsigned syncdownTargetedAction : 2;
 
-        // at least one child has needsAnotherSyncdown set
+        // needs another syncup after pending changes
         unsigned syncupTargetedAction : 2;
     };
 
--- a/include/mega/node.h
+++ b/include/mega/node.h
@@ -113,6 +113,46 @@ private:
     m_off_t mSumSizes = 0;
 };
 
+struct CommandChain
+{   
+    // convenience functions, hides the unique_ptr aspect, removes it when empty
+    bool empty() 
+    {
+        return !chain || chain->empty();
+    }
+
+    void push_back(Command* c)
+    {
+        if (!chain) chain.reset(new std::list<Command*>);
+        chain->push_back(c);
+    }
+
+    void erase(Command* c)
+    {
+        if (chain)
+        {
+            for (auto i = chain->begin(); i != chain->end(); ++i)
+            {
+                if (*i == c)
+                {
+                    chain->erase(i);
+                    if (chain->empty())
+                    {
+                        chain.reset();
+                    }
+                    return;
+                }
+            }
+        }
+    }
+
+private:
+    friend class CommandSetAttr;
+
+    // most nodes don't have commands in progress so keep representation super small
+    std::unique_ptr<std::list<Command*>> chain;
+};
+
 
 // filesystem node
 struct MEGA_API Node : public NodeCore, FileFingerprint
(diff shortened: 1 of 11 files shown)


include/mega/node.h
```diff
@@ -347,8 +394,21 @@ struct MEGA_API LocalNode : public File
...
+
+        // needs another syncdown after pending changes
+        unsigned syncdownTargetedAction: 2;
+
+        // at least one child has needsAnotherSyncdown set
+        unsigned syncupTargetedAction: 2;
+
     };
 
+    // set the syncupPartialAction for this, and parents
```
dwmega, line 406: Should this be "syncupTargetedAction"?

mattw-mega: yes thanks - will adjust.  renamed it during the PR construction :-)

tests/integration/SdkTest_test.cpp
```diff
@@ -4559,13 +4561,22 @@ TEST_F(SdkTest, SdkSimpleCommands)
...
+    ASSERT_EQ(MegaError::API_ESID, err) << "Kill session for unknown sessions shoud fail with API_ESID (error: " << err << ")";
+
+    // getMiscFlags() -- not logged in
+    logout(0);
+    err = synchronousGetMiscFlags(0);
+    ASSERT_EQ(MegaError::API_OK, err) << "Get misc flags failed (error: " << err << ")";
+
+    // getUserEmail() test
+    ASSERT_NO_FATAL_FAILURE(getAccountsForTest(1));
+
```
sergiohs84, line 4573: I think this part is useless here, since the emails has already been tested [here](https://github.com/meganz/sdk/pull/2198/files#diff-eb16573401d197a6ba4d030ed6fa3727R4535) 
```suggestion cpp
```

tests/integration/Sync_test.cpp
```diff
@@ -2049,7 +2049,7 @@ void waitonsyncs(chrono::seconds d = std::chrono::seconds(4), StandardClient* c1
     auto totalTimeoutStart = chrono::steady_clock::now();
     auto start = chrono::steady_clock::now();
     std::vector<StandardClient*> v{ c1, c2, c3, c4 };
-    bool onelastsyncdown = true;
+//    bool onelastsyncdown = true;
```
sergiohs84, line 2052: IMHO, if the commented code in this test is not going to be restored, better to remove it.
Otherwise, it would be good to add a comment on why it's commented.

src/megaclient.cpp
```diff
@@ -13268,8 +13315,9 @@ bool MegaClient::syncdown(LocalNode* l, LocalPath& localpath, bool rubbish)
                 }
 
                 // recurse into directories of equal name
-                if (!syncdown(ll, localpath, rubbish) && success)
+                if (!syncdown(ll, localpath, scanWholeSubtree))
                 {
+                    l->syncdownTargetedAction = std::max<int>(originalTargetedAction, LocalNode::synctree_descendantflagged);
```
sergiohs84, line 13320: If both values are unsigned, I think we should use `max<unsigned>`

src/megaclient.cpp
```diff
@@ -13278,7 +13326,7 @@ bool MegaClient::syncdown(LocalNode* l, LocalPath& localpath, bool rubbish)
 
             lit++;
         }
-        else if (rubbish && ll->deleted)    // no corresponding remote node: delete local item
+        else if (ll->deleted)    // no corresponding remote node: delete local item
```
sergiohs84, line 13329: There is an inner block that test exactly the same again (now that the condition has changed). I'll remove the duplicated tests ;)

src/megaclient.cpp
```diff
@@ -14005,12 +14096,12 @@ bool MegaClient::syncup(LocalNode* l, dstime* nds, size_t& parentPending)
     return true;
 }
 
-bool MegaClient::syncup(LocalNode* l, dstime* nds)
-{
-    size_t numPending = 0;
-
-    return syncup(l, nds, numPending) && numPending == 0;
-}
+//bool MegaClient::syncup(LocalNode* l, dstime* nds, bool scanWholeSubtree)
```
sergiohs84, line 14099: If this method has been discarded, we should remove it.
Otherwise, it's good to have a comment indicating why it's commented.

sergiohs84 (changes requested):

include/mega/node.h
```diff
@@ -347,8 +394,21 @@ struct MEGA_API LocalNode : public File
 
         // checked for missing attributes
         bool checked : 1;
+
+        // needs another syncdown after pending changes
+        unsigned syncdownTargetedAction: 2;
```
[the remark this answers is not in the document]
sergiohs84: I've declared the enum values as `unsigned`. Done :)

include/mega/node.h
```diff
@@ -113,6 +113,46 @@ struct Fingerprints
     m_off_t mSumSizes = 0;
 };
 
+struct CommandChain
```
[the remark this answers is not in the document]
mattw-mega: A chain is implicitly plural so that would be a bit odd grammatically, in english at least.

include/mega/node.h
```diff
@@ -347,8 +394,21 @@ struct MEGA_API LocalNode : public File
 
         // checked for missing attributes
         bool checked : 1;
+
+        // needs another syncdown after pending changes
+        unsigned syncdownTargetedAction: 2;
+
+        // at least one child has needsAnotherSyncdown set
```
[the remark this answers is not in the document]
mattw-mega: Yes, I changed those variable names a few times - comment updated, thanks

src/command.cpp
```diff
@@ -41,6 +41,32 @@ void Command::cancel()
...
+void Command::addToNodePendingCommands(Node* node)
+{
+    node->mPendingChanges.push_back(this);
+}
+
+void Command::removeFromNodePendingCommands(handle h, MegaClient* client)
+{
+    if (auto node = client->nodebyhandle(h))
+    {
+        removeFromNodePendingCommands(node);
```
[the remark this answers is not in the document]
mattw-mega: Correct - commands are stored by their pointer, so it'll find and remove the exact one.  Yes,in normal cases the removal will be in order so it'll be found immediately at the front of the chain.

src/commands.cpp
```diff
@@ -1005,6 +1007,15 @@ bool CommandSetAttr::procresult(Result r)
             }
         }
     }
+
+    if (Node* n = client->nodebyhandle(h))
+    {
+        // in case the name of the file/folder changed
+        if (n && n->parent && n->parent->localnode)
```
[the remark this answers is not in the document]
mattw-mega: true, I'll remove that one

tests/integration/SdkTest_test.cpp
mattw-mega: It's needed for the --RESUMESESSIONS case, where we are trying to run the tests more quickly by skipping logout/login inbetween tests.   Because this test cancels it's own session we need to actually log in again.    For the non resume sessions case, we don't care so much about speed.

tests/integration/Sync_test.cpp
mattw-mega: Ok, removed.   I wasn't sure if there were still cases where it might be needed but the tests are passing so let's remove it.

src/megaclient.cpp
mattw-mega: updated

src/megaclient.cpp
```diff
@@ -13298,23 +13346,21 @@ bool MegaClient::syncdown(LocalNode* l, LocalPath& localpath, bool rubbish)
                 }
```
mattw-mega, line 13346: We can't remove the ll->deleted test here - the lines above may have set it false since the last time we tested it.   Unfortunately our auto tests are not thorough enough yet to detect this error.

src/megaclient.cpp
sergiohs84: I will change the other one too, which is equivalent to this one: https://github.com/meganz/sdk/pull/2198/files#diff-29ce12163767d4d38759d549fa619b81R13510

tests/integration/SdkTest_test.cpp
sergiohs84: Oh, I see. I didn't realized about the resume sessions mode :+1:

src/megaclient.cpp
sergiohs84: My sincere apologies for the mistake, Matt. Great you checked my commits :sweat:

sergiohs84 (approved):

PR #2201, opened 2020-08-24
SDK-1191. Merged sync test support code from the OneWay Sync branch

mattw-mega:
Including the One Way Symmetric test, but renamed for Two Way
Disabled for now but it will be updated to actually test combinations of conditions for two-way syncs

### DIFF
--- a/src/posix/net.cpp
+++ b/src/posix/net.cpp
@@ -61,12 +61,12 @@ bool SockInfo::createAssociateEvent()
     return true;
 }
 
-bool SockInfo::checkEvent(bool& read, bool& write)
+bool SockInfo::checkEvent(bool& read, bool& write, bool logErr)
 {
     WSANETWORKEVENTS wne;
     memset(&wne, 0, sizeof(wne));
     auto err = WSAEnumNetworkEvents(fd, NULL, &wne);
-    if (err)
+    if (err && logErr)
     {
         auto e = WSAGetLastError();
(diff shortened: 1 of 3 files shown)


src/posix/net.cpp
```diff
@@ -61,12 +61,12 @@ bool SockInfo::createAssociateEvent()
...
 }
 
-bool SockInfo::checkEvent(bool& read, bool& write)
+bool SockInfo::checkEvent(bool& read, bool& write, bool logErr)
 {
     WSANETWORKEVENTS wne;
     memset(&wne, 0, sizeof(wne));
     auto err = WSAEnumNetworkEvents(fd, NULL, &wne);
-    if (err)
+    if (err && logErr)
```
sergiohs84, line 69: I feel this change may have side effects. For safety, I would change it this way:
``` cpp
if (err)
{
         if (logErr)
        {
            auto e = WSAGetLastError(); 
            LOG_err << "WSAEnumNetworkEvents error " << e;
        }
        return false;
}

sergiohs84 (changes requested):

src/posix/net.cpp
mattw-mega: Good point, I'll do that, thanks.   Also, I saw you removed some trailing spaces - I've now installed a Visual Studio plugin so that I can see those :-)  cheers

sergiohs84 (approved):

PR #2202, opened 2020-08-25
SDK-1193. Hotfix/fix crash on megasync retry upload

mattw-mega:
only two very small changes really, but also I've updated my editor with a tool that shows and removes trailiing whitespace, as per the general unix preference.

### DIFF
--- a/src/filesystem.cpp
+++ b/src/filesystem.cpp
@@ -392,7 +392,7 @@ void DirNotify::setFailed(int errCode, const string& reason)
 
 int DirNotify::getFailed(string& reason)
 {
-    if (mFailed) 
+    if (mFailed)
     {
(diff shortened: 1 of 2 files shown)


sergiohs84 (approved):
I guess the only change in this PR is https://github.com/meganz/sdk/pull/2202/files#diff-a46176812da95138343d663ce59718c6R926, which will have no difference in Windows, but in other OS it will apply `%1` instead of `%2`, since the separator is not 2 chars size.
The removal of empty spaces makes the review quite unconvenient. Please, stop submitting PRs with mixed changes (it's OK to prepare PRs which exclusively remove tailing spaces and alike, but not mixed with changes in the logic). Thanks

polmr (approved):

PR #2204, opened 2020-08-25
Prevent incorrect speeds being reported during network glitches

mattw-mega:
We should never have a negative number of bytes being sent or received since the last update.


icaroalvarez (approved):
The patch solves receiving negative speed values in MEGASync (and other clients). I wonder if (p - progressreported) giving a negative result means there is still an issue but if it is only used to calculate the speed would be enough with the patch.

mattw-mega:
Those other two place the `diff` value is passed, also go to SpeedControllers, so I think it's correct - thanks for the review!  cheers

PR #2206, opened 2020-08-26
SDK-1192: Delete sync test files from previous runs

catalinr-m:
* Old files are being deleted before starting tests.

* The new `TestFS` class should probably be in separate h/cpp files, and that will require modifying the build files, but I'm afraid I'll miss something if I do that. Should I do that?
* Code could be simplified a bit if `std::filesystem` was always available.


catalinr-m:
trigger new run

mattw-mega (approved):

catalinr-m:
trigger compilation

PR #2215, opened 2020-08-31
SDK-1199. Provide a way for apps to call setrlimit

mattw-mega:
The defaults on some platforms are too low for uploading or downloading many files at once, as the SDK may do on request.   Let's allow the app to choose an overall limit that suits the SDK and whatever else the app may be doing.

### DIFF
--- a/src/megaapi_impl.cpp
+++ b/src/megaapi_impl.cpp
@@ -7490,7 +7490,7 @@ bool MegaApiImpl::hasToForceUpload(const Node &node, const MegaTransferPrivate &
     return canForceUpload && (isMedia || isPdf) && !(hasPreview && hasThumbnail);
 }
 
-void MegaApiImpl::platformSetRLimitNumFile(int newNumFileLimit) const
+bool MegaApiImpl::platformSetRLimitNumFile(int newNumFileLimit) const
 {
 #ifndef WIN32
     struct rlimit rl{0,0};
@@ -7498,6 +7498,7 @@ void MegaApiImpl::platformSetRLimitNumFile(int newNumFileLimit) const
     {
         auto e = errno;
         LOG_err << "Error calling getrlimit: " << e;
+        return false;
     }
     else
     {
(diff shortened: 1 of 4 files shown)


src/megaapi_impl.cpp
```diff
@@ -7489,6 +7490,30 @@ bool MegaApiImpl::hasToForceUpload(const Node &node, const MegaTransferPrivate &
...
 }
 
+void MegaApiImpl::platformSetRLimitNumFile(int newNumFileLimit) const
+{
+#ifndef WIN32
+    struct rlimit rl{0,0};
+    if (0 < getrlimit(RLIMIT_NOFILE, &rl))
+    {
+        auto e = errno;
+        LOG_err << "Error calling getrlimit: " << e;
```
sergiohs84, line 7500: We may want to return a boolean in this function, so it returns false in case of error. That way, the app knows its call has failed.

src/megaapi_impl.cpp
```diff
@@ -7489,6 +7490,30 @@ bool MegaApiImpl::hasToForceUpload(const Node &node, const MegaTransferPrivate &
...
+    }
+    else
+    {
+        LOG_info << "rlimit for NOFILE before change is: " << rl.rlim_cur << ", " << rl.rlim_max;
+        rl.rlim_cur = rlim_t(newNumFileLimit);
```
sergiohs84, line 7505: Should we avoid the call `if (newNumFileLimit > rl.rlim_max)`?? maybe not worthy if it simply fails without side-effects, falling into the error's block below.

sergiohs84: Or maybe we could force `rl.rlim_max` if the requested limit is above the actual limit in the platform :thinking:

mattw-mega: Yes, we'll log an error if the OS won't accept the change.   And we're not allowed to change the max, anyway (except with admin privilege).  On iOS that max number is extremely large anyway, it might be max int.

src/megaapi_impl.cpp
mattw-mega: If the OS won't accept the limit then I don't think there's really anything the app can do to improve its situation?  We already logged the issue.

src/megaapi_impl.cpp
sergiohs84: In Android, the max number depends on the device/Android version. In iOS it may change also with the OS version. That's why I suggest the override of the maximum value.

catalinr-m:
Last 2 commits are intended to address the review comments.

sergiohs84 (approved):

PR #2226, opened 2020-09-03
SDK-1195: Smart Banner

catalinr-m:


src/commands.cpp
```diff
@@ -8261,4 +8261,114 @@ bool CommandBackupRemove::procresult(Result r)
...
+                "l": "", ///URL
+                "bimg": "Verified_BG.png", ///background image name.
+                "dsp": "https://web-sandbox3.developers.mega.co.nz/images/mega/" ///Where to get the image.
+            },
+            {
+                ...
+            }
+        ]
+    ]
+    */
```
sergiohs84, line 8296: It's a good idea, IMO, to have a sample JSON here as you added. But let's try to collapse it a bit:
```suggestion
    /*
        [ {
                "id": 2, ///The banner id
                "t": "R2V0IFZlcmlmaWVk", ///Banner title
                "d": "TWFrZSBpdCBlYXNpZXIgZm9yIHlvdXIgY29udGFjdHMgdG8gZmluZCB5b3Ugb24gTUVHQS4", ///Banner description.
                "img": "Verified_image.png", ///Image name.
                "l": "", ///URL
                "bimg": "Verified_BG.png", ///background image name.
                "dsp": "https://web-sandbox3.developers.mega.co.nz/images/mega/" ///Where to get the image.
            }, {"id":3, ...}, ... ]
    */
```

src/commands.cpp
```diff
@@ -8261,4 +8261,114 @@ bool CommandBackupRemove::procresult(Result r)
...
+        return false; // parsing failed
+    }
+
+    vector< tuple<int, string, string, string, string, string, string> > banners;
+
+    // loop array elements
+    while (client->json.enterobject())
+    {
+        int fieldcount = 7,
+            id = 0;
```
sergiohs84, line 8311: We try to avoid declaration of multiple variables in a row like this ones, with initialization.
Also, use cammelCase for naming variables, please.

It's preferred to do:
```suggestion
        int fieldCount = 7;
        int id = 0;
```

src/commands.cpp
```diff
@@ -8261,4 +8261,114 @@ bool CommandBackupRemove::procresult(Result r)
...
+        string title, description, img, url, bimg, dsp;
```
sergiohs84, line 8312: This one is OK :wink:

src/commands.cpp
```diff
@@ -8261,4 +8261,114 @@ bool CommandBackupRemove::procresult(Result r)
...
+            case MAKENAMEID4('b', 'i', 'm', 'g'):
+                ok = client->json.storeobject(&bimg);
+                break;
+
+            case MAKENAMEID3('d', 's', 'p'):
+                ok = client->json.storeobject(&dsp);
+                break;
+
+            case EOO:  // [[fallthrough]];
+            default:
```
sergiohs84, line 8350: We need to be prepared for future changes in the API, where they can add new fields. The parser should skip the unknown items, in that case, so the `default` should consume that part of the JSON (see other commands as reference, you'll see a `storeobject()` in there). In consequence, the `EOO` should be the exit path of the loop (you can use a `bool exit = true` for that purpose).
On the other hand, the command should usually attempt to read the JSON response to the command in full. It means that, in case one of the items cannot be read successfully, we may want to keep going (instead of breaking the loop due to `!ok`). Sometimes, an issue parsing one of the items is not a critical issue. The sanity checks should be done upon `EOO`, where some tests are done against the received values to double-check they are valid and make sense.

src/megaapi_impl.cpp
```diff
@@ -15674,6 +15685,34 @@ void MegaApiImpl::checkfile_result(handle h, error e, byte*, m_off_t, m_time_t,
...
+
+void MegaApiImpl::getbanner_result(vector< tuple<int, string, string, string, string, string, string> >&& banners)
+{
+    auto it = requestMap.find(client->restag);
+    if (it == requestMap.end())
+    {
+        return;
+    }
+
+    MegaRequestPrivate* request = it->second;
```
sergiohs84, line 15709: We should test against `request` is valid (it should not, but it could be nullptr) and against the type of request (sometimes, the same `restag` is used for multiple requests (when they are chained). Just for homogeneity, we test type for all commands.
```suggestion cpp
    if(requestMap.find(client->restag) == requestMap.end()) return;
    MegaRequestPrivate* request = requestMap.at(client->restag);
    if(!request || (request->getType() != MegaRequest::TYPE_GET_BANNERS)) return;
```

sergiohs84 (changes requested):
It looks pretty good overall, @catalinr-m, good work! Please, see my comments and proceed with the implementation of the `dismissBanner()`. It would be great also if you add a command to `megacli` to exercise this feature, something like `banners get | dismiss <id>`

include/megaapi.h
```diff
@@ -2739,6 +2739,25 @@ class MegaRecentActionBucketList
     virtual int size() const;
 };
 
+class MegaBanner
+{
+public:
+    MegaBanner(std::tuple<int, std::string, std::string, std::string, std::string, std::string, std::string>&& details);
+
+    int getId() const;
+    const std::string& getTitle() const;
```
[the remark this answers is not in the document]
catalinr-m: Done.

src/commands.cpp
catalinr-m: Updated.
I can't think of some good sanity checks now, but I can add some upon suggestions.

src/megaapi_impl.cpp
```diff
@@ -3878,6 +3888,7 @@ const char *MegaRequestPrivate::getRequestString() const
         case TYPE_SET_RETENTION_TIME: return "SET_RETENTION_TIME";
         case TYPE_RESET_SMS_VERIFIED_NUMBER: return "RESET_SMS_VERIFIED_NUMBER";
         case TYPE_SEND_DEV_COMMAND: return "SEND_DEV_COMMAND";
+        case TYPE_GET_BANNERS: return "TYPE_GET_BANNERS";
```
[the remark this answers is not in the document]
catalinr-m: Done.

src/commands.cpp
```diff
@@ -8261,4 +8261,114 @@ bool CommandBackupRemove::procresult(Result r)
...
+                "dsp": "https://web-sandbox3.developers.mega.co.nz/images/mega/" ///Where to get the image.
+            },
+            {
+                ...
+            }
+        ]
+    ]
+    */
+
+    if (!client->json.enterarray() ||
```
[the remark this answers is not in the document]
catalinr-m: I'll leave this unresolved until there's a way to test it.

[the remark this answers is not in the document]
catalinr-m: Updated after testing.

catalinr-m:
Trigger compilation

src/megaapi_impl.cpp
```diff
@@ -22749,6 +22811,21 @@ bool MegaApiImpl::tryLockMutexFor(long long time)
...
+{
+    MegaRequestPrivate *request = new MegaRequestPrivate(MegaRequest::TYPE_GET_BANNERS, listener);
+    requestQueue.push(request);
+    waiter->notify();
+}
+
+void MegaApiImpl::dismissBanner(int id, MegaRequestListener *listener)
+{
+    MegaRequestPrivate *request = new MegaRequestPrivate(MegaRequest::TYPE_DISMISS_BANNER, listener);
+    request->setNumber(id);
```
sergiohs84, line 22824: This is the right place to capture the `ts` for the `dban` command ;)

sergiohs84 (changes requested):

src/commands.cpp
```diff
@@ -8250,4 +8252,108 @@ bool CommandBackupRemove::procresult(Result r)
...
+    client->app->getbanners_result(move(banners));
+
+    return true;
+}
+
+CommandDismissBanner::CommandDismissBanner(MegaClient* client, int id)
+{
+    cmd("dban");
+    arg("id", id); // id of the Smart Banner
+    uint64_t msSinceEpoch = std::chrono::system_clock::now().time_since_epoch() / std::chrono::milliseconds(1);
```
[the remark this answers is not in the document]
catalinr-m: Well, at least to avoid using `time_t` because "encoding of calendar time in std::time_t is unspecified", even if _usually_ it is what we need. Do you want me to change it to use time()?

src/megaapi_impl.cpp
```diff
@@ -15871,6 +15883,46 @@ void MegaApiImpl::checkfile_result(handle h, error e, byte*, m_off_t, m_time_t,
...
+    {
+        return;
+    }
+
+    MegaRequestPrivate* request = it->second;
+
+    if (!request || (request->getType() != MegaRequest::TYPE_GET_BANNERS))
+    {
+        return;
+    }
```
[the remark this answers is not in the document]
catalinr-m: Ok, I see the point of making code shorter (even though it contradicts the [conclusion of a slack discussion](https://megaconz.slack.com/archives/C02FB882Q/p1597767279420500) 😃), but are the 2 sets of round brackets a style requirement?

include/megaapi.h
```diff
@@ -2739,6 +2739,40 @@ class MegaRecentActionBucketList
...
+* @brief Represents a set of properties that define a SmartBanner.
+* These are used to display a banner in mobile apps.
+*
+* MegaBanner-s can be retrieved from MegaBannerList
+*
+*/
+class MegaBanner
+{
+public:
+    MegaBanner(std::tuple<int, std::string, std::string, std::string, std::string, std::string, std::string>&& details);
```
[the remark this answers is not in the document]
catalinr-m: Shouldn't these be enforced then, by a common base class/interface?

include/megaapi.h
```diff
@@ -2739,6 +2739,40 @@ class MegaRecentActionBucketList
...
+
+    int getId() const;
+    const char* getTitle() const;
+    const char* getDescription() const;
+    const char* getImage() const;
+    const char* getUrl() const;
+    const char* getBackgroundImage() const;
+    const char* getImageLocation() const;
```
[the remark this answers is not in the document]
catalinr-m: Should these be made pure virtual here while implemented in `MegaBannerPrivate`?

sergiohs84 (changes requested):

src/megaapi_impl.cpp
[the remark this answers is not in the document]
catalinr-m: I made the change to shorten the code, but I'm still puzzled by the second part:
- original code:
> if (!request || `(`request->getType() != MegaRequest::TYPE_GET_BANNERS`)`)
- suggested change:
> if (!request || `((`request->getType() != MegaRequest::TYPE_GET_BANNERS`))`) return;

Was this intended or a typo?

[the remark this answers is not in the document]
sergiohs84: It must be a typo, I don't see a good reason for the double brackets in there.

include/megaapi.h
```diff
@@ -3525,6 +3636,20 @@ class MegaRequest
...
+         * The SDK retains the ownership of the returned value. It will be valid until
+         * the MegaRequest object is deleted.
+         *
+         * This value is valid for these requests in onRequestFinish when the
+         * error code is MegaError::API_OK:
+         * - MegaApi::getBanners - Requests all Smart Banners available for current user
+         *
+         * @return List of all Smart Banners available for current user
+         */
+        virtual MegaBannerList* getBannerList() const;
```
sergiohs84, line 3654: When the getter returns a `MegaXYZ`, we tend to call these getter something like `getMegaXYZ()`.

include/megaapi.h
```diff
@@ -17642,7 +17767,19 @@ class MegaApi
...
-         * @brief Request a list of all Smart Banners available for current user.
+         * @brief Requests a list of all Smart Banners available for current user.
+         *
+         * The response value is stored as a MegaBannerList.
+         *
+         * The associated request type with this request is MegaRequest::TYPE_GET_BANNERS
+         * Valid data in the MegaRequest object received in onRequestFinish when the error code
+         * is MegaError::API_OK:
+         * - MegaRequest::getMegaStringListMap where the keys are two-letter country codes and the
+         *   values a list of calling codes.
```
sergiohs84, line 17780: Outdated copy&paste:
```suggestion
         * - MegaRequest::getMegaBannerList: the list of banners
```

include/megaapi.h
```diff
@@ -17642,7 +17767,19 @@ class MegaApi
...
+         *
+         * For this command, there are currently no command specific error codes returned by the API.
```
sergiohs84, line 17782: Actually, check the documentation:
```
EACCESS if this command is called with an invalid userid or while the user is not logged in.
EINTERNAL if the internally used user attribute exists but can't be decoded.
ENOENT if there are no banners to return to the user. ```

File: include/megaapi_impl.h [hunk 17]
```diff
@@ -1275,7 +1276,7 @@ class MegaRequestPrivate : public MegaRequest
         MegaBackgroundMediaUpload* backgroundMediaUpload;  // non-owned pointer
 
     private:
-        unique_ptr<MegaBannerList> mBannerList;
+        unique_ptr<MegaBannerListPrivate> mBannerList;
```
sergiohs84, line 1279: Why is it private instead of protected, as the rest of the members?
Also, to avoid casting issues, I think it should store a pointer to the base class :)

include/megaapi.h
```diff
@@ -3311,6 +3311,7 @@ class MegaRequest
          * - MegaApi::setUserAttribute - Returns the attribute type
          * - MegaApi::getUserAttribute - Returns the attribute type
          * - MegaApi::setMaxConnections - Returns the direction of transfers
+         * - MegaApi::dismissBanner - Returns the id of the banner
```
sergiohs84, line 3314: It's great you updated it here (despite we are not keeping the list up to date anymore). We usually document only the method in MegaApi, with the information about the members in the request that are valid. Not the other way.
Just for your information, no need to change it

include/megaapi_impl.h
catalinr-m: > Why is it private instead of protected, as the rest of the members?

Probably just a habit, but now that I think of it, I can find a few good reasons:
- Because I could not find any reason for giving it more access -- it's more restrictive and still honors the purpose.
- Because it's an implementation detail, that is managed by a dedicated function that creates it, and accessed by another.
- Because an implementation detail should be allowed to change without affecting functionality -- think of creating a `MegaBannerListPrivate` on the fly;
- Because we can always blame the people that invented encapsulation.

So why not? I don't think "consistency" can be invoked here.

> Also, to avoid casting issues, I think it should store a pointer to the base class :)

I can't think of any issues with upcasting a _pointer_. Surely that's why they made it implicit. What am I missing?
Also, if I do that, then I will need to either explicitly cast it when used throughout the SDK, or add `add()` to the base class, which will make it modifiable by a user, which should not be allowed, right?

> Just picky comments, feel free to disregard those ones

I'll keep current code, as long as it's up to me. If you still want me to change them, please point them out again.

catalinr-m:
Trigger compilation

include/megaapi_impl.h
sergiohs84: :+1: good points for encapsulation. I asked because I thought I saw some others where you used `protected` by "default", but not here. I'm totally fine using private here, and in case a child class needs access, we can always change it :wink: 

As for the casting thing, if you haven't observed issue warnings or alike, it's all good. We definitely don't want to add `add()` to the public class.

sergiohs84 (dismissed):

sergiohs84 (changes requested):

sergiohs84 (approved):

PR #2225, opened 2020-09-02
SDK-1034. Move CACHEDSTATUS to a new table

polmr:
to be able to commit changes immediately

PR #2228, opened 2020-09-03
SDK-1209. Initialize tctable and do not purgeOrphanTransfers if null

polmr:


sergiohs84:
Not a real issue, but introduced in PR https://github.com/meganz/sdk/pull/2225.
Closing

PR #2229, opened 2020-09-04
SDK-1210. Fixes for auto sync tests after sync management merge

mattw-mega:


tests/integration/SdkTest_test.cpp
```diff
@@ -5035,10 +5053,10 @@ TEST_F(SdkTest, SyncResumptionAfterFetchNodes)
 
     LOG_verbose << " SyncResumptionAfterFetchNodes : syncying folders";
 
-    int tag1 = syncFolder(sync1Path);
+    int tag1 = syncFolder(sync1Path);  tag1;
```
catalinr-m, line 5056: Out of curiosity, what is this for?

polmr, line 5056: :upside_down_face: :smile:

tests/integration/SdkTest_test.cpp
```diff
@@ -5107,7 +5125,7 @@ TEST_F(SdkTest, SyncResumptionAfterFetchNodes)
     removeSync(sync4Path);
 
     // wait for the sync removals to actually take place
-    std::this_thread::sleep_for(std::chrono::seconds{20});
+    std::this_thread::sleep_for(std::chrono::seconds{5});
```
polmr, line 5128: Note the time reduction before merging. Is it intentionally commited?

polmr (approved):

tests/integration/SdkTest_test.cpp
polmr: to avoid unused var warnings I guess :-D

catalinr-m: IIRC that won't work when optimizations are enabled, i.e. Release builds will still get the warnings.
Using `(void)tag1;` should work for all cases.

polmr (dismissed):

tests/integration/SdkTest_test.cpp
polmr: wouldn't it? Interesting ... I would have though those warnings were reported before the optimizations take place :thinking:

catalinr-m: Probably (only?) the NOPs are optimized earlier.
At the end of the optimized build unused variables won't be present at all in the final executable, but the warning is there even for those cases, so not all optimization have been applied at that point.

sergiohs84 (approved):

tests/integration/SdkTest_test.cpp
mattw-mega: yes - the tests are pretty slow already, and 5 seconds should be more than enough :-)  thanks

PR #2236, opened 2020-09-09
picked changes that fixed xcode compilation

polmr:
SyncConfig equality was only used by unit tests
And clang on the mac did not like the std::tie/std::tuple method of
implementing equality


mattw-mega (approved):

jnavarrom:
The compilation on iOS works pointing to that branch
Can we merge the PR into develop?
Thanks

PR #2237, opened 2020-09-10
SDK-1217. cURL was not reporting 500 back to the app with Connection: close

mattw-mega:


src/posix/net.cpp
```diff
@@ -1463,6 +1463,8 @@ void CurlHttpIO::send_request(CurlHttpContext* httpctx)
         curl_easy_setopt(curl, CURLOPT_TCP_KEEPINTVL, 60L);
         curl_easy_setopt(curl, CURLOPT_SOCKOPTFUNCTION, sockopt_callback);
         curl_easy_setopt(curl, CURLOPT_SOCKOPTDATA, (void*)req);
+        curl_easy_setopt(curl, CURLOPT_FAILONERROR, 1L);
```
sergiohs84, line 1466: According to cUrl documentation, this option "_tells the library to fail the request if the HTTP code returned is equal to or larger than 400._", while the default behavior would be to return the page as usual, ignoring the error.

I wonder if enabling the option would have any side effect on the management of `HttpReqXfer::status`. When it's set to `REQ_FAILURE`, it handles the error codes 509, 429, 403, 404 and 503. My concern is whether these codepaths will still be visited with this option enabled?

sergiohs84: Maybe we need to handle, in case of failed requests, the `CURLE_HTTP_RETURNED_ERROR`
_When this option is used and an error is detected, it will cause the connection to get closed and CURLE_HTTP_RETURNED_ERROR is returned._

tests/integration/SdkTest_test.cpp
```diff
@@ -4585,9 +4586,6 @@ TEST_F(SdkTest, SdkSimpleCommands)
     logout(0);
     err = synchronousGetMiscFlags(0);
     ASSERT_EQ(MegaError::API_OK, err) << "Get misc flags failed (error: " << err << ")";
-
```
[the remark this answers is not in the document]
mattw-mega: Yes, that's right - the test was still problematic after that last solution though, because if the test fails before the re-login, that is not performed.   This time I'm marking the saved session as invalid so that the next test will perform the full login if needed.  thanks

src/posix/net.cpp
mattw-mega: Yes, this case is handled already in CurlHttpIO::multidoio().   It passes the CURLMSG_DONE test, and assigns CURLE_HTTP_RETURNED_ERROR to errorCode from msg->data.result.   After extracting the http code and storing it in our Req object, it also assigns REQ_FAILURE to the status member after the comment "check httpstatus and response length".   So, I think we are already covered for this case, and it's working already for these test cases, logging the REQ_FAILURE and so on.   I think, for the normal case when the headers didn't have the `Connection: close` line, the old code worked fine.  But, when the request fails (>=400) and the connection close is triggered like that, the way cURL works internally, it just doesn't store a record for the curl_multi_info_read() to report.

sergiohs84: Alright, it sounds good. Thanks for all the investigation, Matt.
Ready to merge, IMO ;)

sergiohs84 (approved):

PR #2243, opened 2020-09-15
SDK-1221. Completion functions instead of share_result() and exportnode_result()

mattw-mega:
Inspired by the need to fix the failing test case for getpubliclink().
This enables us to easily code the test cases for export/share/getpubliclink()
And also makes the chaining of commands very simple - export node needs to share and then getpubliclink().
But those are not always called in that order.   Matching up the response to the request
and then figuring out whether to chain another command was complex.
Now we can use lambdas to supply the actions to perform when a command completes, with no need to match up tags and request types.
This change can serve as an example for discussion as to whether we should adopt the method widely.
Also without needing to use overrides of MegaApp virtual functions, we can keep the requesting and response processing code close together.

### DIFF
--- a/include/mega/command.h
+++ b/include/mega/command.h
@@ -165,7 +165,7 @@ public:
     };
 
     virtual bool procresult(Result) = 0;
-    
+
     // json for the command is usually pre-generated but can be calculated just before sending, by overriding this function
     virtual const char* getJSON(MegaClient* client);
 
@@ -639,12 +639,14 @@ class MEGA_API CommandSetShare : public Command
     string msg;
     string personal_representation;
 
+    std::function<void(Error)> completion;
+
     bool procuserresult(MegaClient*);
 
 public:
     bool procresult(Result) override;
 
-    CommandSetShare(MegaClient*, Node*, User*, accesslevel_t, int, const char*, const char* = NULL);
+    CommandSetShare(MegaClient*, Node*, User*, accesslevel_t, int, const char*, const char*, int tag, std::function<void(Error)> f);
 };
 
 class MEGA_API CommandGetUserData : public Command
@@ -742,11 +744,12 @@ class MEGA_API CommandSetPH : public Command
 {
     handle h;
     m_time_t ets;
+    std::function<void(Error, handle, handle)> completion;
 
 public:
     bool procresult(Result) override;
 
-    CommandSetPH(MegaClient*, Node*, int, m_time_t);
+    CommandSetPH(MegaClient*, Node*, int, m_time_t, int ctag, std::function<void(Error, handle, handle)> f);
 };
 
 class MEGA_API CommandGetPH : public Command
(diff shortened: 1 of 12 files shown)


include/mega/command.h
```diff
@@ -639,12 +639,14 @@ class MEGA_API CommandSetShare : public Command
     string msg;
     string personal_representation;
 
+    std::function<void(Error)> completion;
```
sergiohs84, line 642: Maybe we can name it `callback`? (or `mCallback`, indeed)

sergiohs84: And also, should we simply add this new member to `Command` class? I guess the cost of keeping this object for every command shouldn't be a problem. And it will be available for any new command, just by design. Perhaps a pointer, so we avoid the instantiation :thinking:

src/pubkeyaction.cpp
```diff
@@ -108,16 +109,17 @@ void PubKeyActionCreateShare::proc(MegaClient* client, User* u)
...
     // we have all ingredients ready: the target user's public key, the share
     // key and all nodes to share
-    client->restag = tag;
-    client->reqs.add(new CommandSetShare(client, n, u, a, newshare, NULL, selfemail.c_str()));
+    client->reqs.add(new CommandSetShare(client, n, u, a, newshare, NULL, selfemail.c_str(), tag, move(completion)));
 }
 
 // share node sh with access level sa
-PubKeyActionCreateShare::PubKeyActionCreateShare(handle sh, accesslevel_t sa, int ctag, const char* personal_representation)
+PubKeyActionCreateShare::PubKeyActionCreateShare(handle sh, accesslevel_t sa, int ctag, const char* personal_representation, std::function<void(Error)> f)
```
sergiohs84, line 116: Now that `personal_representation` won't be nullptr by default, we may want to adjust the test:
`if (personal_representation)` to become `if (personal_representation && personal_representation[0])`.

sergiohs84 (dismissed):
Very nice simplification thanks to the callback's param, @mattw-mega .
Please, see my comments. Nevertheless, I'll leave the PR pre-approved, since they are minor stuff.

include/mega/command.h
mattw-mega: I don't think we should add it to the command class, because different commands will have different parameters to pass to the completion function - some may pass a string result along with an Error for example, or node handles etc.   Not all will only need an Error, though that is the most common case.   Since the function signature is part of the type, we can't support any variation if we put it in the base class.

src/pubkeyaction.cpp
mattw-mega: My idea about that was just that anywhere we were using the default before, we would pass nullptr now, so it's all the same in that respect.   Scanning through now, I think it's only used in setshare() anyway, and that parameter is passed from setshare()'s parameter of the same name, and so in fact the default was never used anyway.

sergiohs84:
trigger compilation

sergiohs84 (approved):
Automated tests fail for `mcc` command, but it's because the `bt1` API server does not reply to that command. It's safe to merge this PR into the main branch. Thanks

PR #2244, opened 2020-09-15
SDK-1222 - Minor syncup tweaks.

dwmega:
**Simplify syncup's flow of control.**

Syncup returns to its caller from several points. This is unfortunate as there is book-keeping that must be done on return which is currently duplicated. Instead of returning directly, syncup will now update a 'success' flag and break out of its main loop. This allows us to have the book-keeping logic in one place.

**Correctly update parentPending.**

**Correctly update l->syncupTargetedAction.**


sergiohs84:
@dwmega please, can you update the ticket accordingly, based on the decision you made for this PR/task?? thanks

PR #2249, opened 2020-09-17
SDK-1224. When a node is moved to the trash, remove outshares

mattw-mega:
For that node, and any descendant.  Also pending outshares.

### DIFF
--- a/src/megaclient.cpp
+++ b/src/megaclient.cpp
@@ -7792,6 +7792,12 @@ error MegaClient::rename(Node* n, Node* p, syncdel_t syncdel, handle prevparent,
         return e;
     }
 
+    if (p->firstancestor()->type == RUBBISHNODE)
+    {
+        // similar to the webclient, send `s2` along with `m` if the node is moving to the rubbish
+        removeOutSharesFromSubtree(n);
+    }
+
     Node *prevParent = NULL;
     if (!ISUNDEF(prevparent))
     {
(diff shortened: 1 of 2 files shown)


src/megaclient.cpp
```diff
@@ -7852,6 +7858,36 @@ error MegaClient::rename(Node* n, Node* p, syncdel_t syncdel, handle prevparent,
...
+    }
+
+    if (n->outshares)
+    {
+        for (auto& it : *n->outshares)
+        {
+            if (it.second->user)
+            {
+                setshare(n, it.second->user->email.c_str(), ACCESS_UNKNOWN, nullptr);
+            }
```
sergiohs84, line 7881:
```suggestion
            }
            else    // folder links are a shared folder without user
            {
                setshare(n, nullptr, ACCESS_UNKNOWN, nullptr);
            }
```

sergiohs84 (changes requested):
I'd tried to test the change, as well as the suggestion, but lu1:444 is not responding right now and prod servers are not suitable for v3 of API.
I planned to test:
- Create a pending share folder, remove the PCR and later attempt to move the folder into the rubbish.
- Create a folder link, move the folder to rubbish bin.
- Create a shared folder, remove the user among your contacts, attempt to move the folder into the rubbish.
- Create a shared folder, create a subfolder and share it too, movo root shared folder into the rubbish.

src/megaclient.cpp
mattw-mega: Excellent point.   As lu1 was not available for me either, I've made a new PR with these changes, against `develop`.  And I've incorporated this change too.   All 4 cases that you've listed above are working.   thanks!

mattw-mega:
closing in favour of https://github.com/meganz/sdk/pull/2260 fixing on develop instead

PR #2253, opened 2020-09-18
SDK-1226. Do not use DNS server provided by iOS (release v3.7.3d)

jnavarrom:
**Bugfix**:
- [SDK-1226] - Do not use DNS server provided by iOS


sergiohs84 (approved):

include/mega/http.h
```diff
@@ -91,10 +91,10 @@ namespace mega {
...
 #define APISSLEXPONENT "\x01\x00\x01"
 
-#define MEGA_DNS_SERVERS "2001:678:25c:2215::554,89.44.169.136," \
-                         "2001:67c:1998:2212::13,31.216.148.13," \
-                         "2405:f900:3e6a:1::103,31.216.148.11," \
-                         "2403:9800:c020::43,122.56.56.216"
+#define MEGA_DNS_SERVERS "2001:2:0:1baa::592c:a98d,89.44.169.136," \
+                         "2001:2:0:1baa::42cb:7f0e,89.44.169.141," \
+                         "2001:2:0:1baa::42cb:7f10,66.203.127.16," \
+                         "2001:2:0:1baa::592c:a988,66.203.127.14"
```
sergiohs84, line 97: Actually, the second update we applies for DNS servers was wrong, since the IPs were returned by the system's DNS resolver on a NAT64 network.
We need to revert commit aaef23473f957804350f4672b984261373dea50b

src/posix/net.cpp
```diff
@@ -1399,6 +1404,8 @@ void CurlHttpIO::send_request(CurlHttpContext* httpctx)
         httpctx->posturl.replace(httpctx->posturl.find(httpctx->hostname), httpctx->hostname.size(), httpctx->hostip);
         httpctx->headers = curl_slist_append(httpctx->headers, httpctx->hostheader.c_str());
     }
+    
+#ifndef TARGET_OS_IPHONE
```
sergiohs84, line 1408: I don't understand the reason to skip this code. In case there's no IP nor hostname, do we really want let the code continue?

src/posix/net.cpp
```diff
@@ -306,7 +306,10 @@ CurlHttpIO::CurlHttpIO()
     options.tries = 2;
     ares_init_options(&ares, &options, ARES_OPT_TRIES);
     arestimeout = -1;
+    
+#ifndef TARGET_OS_IPHONE
```
sergiohs84, line 310: If we don't want to filter DNS servers on iOS (which I think it's useless too), maybe it's cleaner to change the method, rather than the calls (ie. add a return inside `filterDNSservers()` in case of iOS).

sergiohs84 (changes requested):

include/mega/http.h
sergiohs84: Done!

src/posix/net.cpp
sergiohs84: Done!

src/posix/net.cpp
sergiohs84: Understood: there should be no proxy (iOS doesn't support it yet) and the IP is not available, since we don't solve hostnames with c-ares in advance for iOS anymore, so there's no chance to replace the hostname by the address. However, we want to proceed and send the request directly to the hostname (cUrl will do the lookup).

sergiohs84 (approved):

PR #2267, opened 2020-09-28
review:  update flags detecting a move in sc_newnodes

mattw-mega:
The case is supposed to detect the move, and cause procsc() to exit for one run of syncdown() before resuming
It was exiting earlier that it used to - inbetween the `d` and `t`, but that should occur after `t` as it used to.

### DIFF
--- a/src/megaclient.cpp
+++ b/src/megaclient.cpp
@@ -4796,18 +4796,18 @@ bool MegaClient::procsc()
                                     }
                                 }
 #endif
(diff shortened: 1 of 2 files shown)


sergiohs84:
trigger compilation

sergiohs84 (approved):

sergiohs84:
The failed test is related to chats, and I think it's an issue related to staging. I don't see any relationship with the changes in this PR, so feel free to merge it.

PR #2287, opened 2020-10-13
SDK-1204. Fix serialization of writable links auth keys giving backwards compatibility

polmr:
added new unit test

### DIFF
--- a/src/node.cpp
+++ b/src/node.cpp
@@ -245,7 +245,6 @@ Node* Node::unserialize(MegaClient* client, const string* d, node_vector* dp)
     int i;
     char isExported = '\0';
     char hasLinkCreationTs = '\0';
-    char hasLinkAuthKey = '\0';
 
     if (ptr + sizeof s + 2 * MegaClient::NODEHANDLE + MegaClient::USERHANDLE + 2 * sizeof ts + sizeof ll > end)
     {
@@ -329,8 +328,15 @@ Node* Node::unserialize(MegaClient* client, const string* d, node_vector* dp)
     hasLinkCreationTs = MemAccess::get<char>(ptr);
     ptr += sizeof(hasLinkCreationTs);
 
-    hasLinkAuthKey = MemAccess::get<char>(ptr);
-    ptr += sizeof(hasLinkAuthKey);
+    auto authKeySize = MemAccess::get<char>(ptr);
+
+    ptr += sizeof authKeySize;
+    const char *authKey = nullptr;
+    if (authKeySize)
+    {
+        authKey = ptr;
+        ptr += authKeySize;
+    }
 
     for (i = 5; i--;)
     {
@@ -431,16 +437,6 @@ Node* Node::unserialize(MegaClient* client, const string* d, node_vector* dp)
         {
             cts = MemAccess::get<m_time_t>(ptr);
             ptr += sizeof(cts);
-        }
-
-        const char *authKey = nullptr;
-        if (hasLinkAuthKey)
-        {
-            auto authKeySize = MemAccess::get<unsigned short>(ptr);
-            ptr += sizeof authKeySize;
-
-            authKey = ptr;
-            ptr += authKeySize;
         }
 
         plink = new PublicLink(ph, cts, ets, takendown, authKey ? authKey : "");
@@ -546,8 +542,16 @@ bool Node::serialize(string* d)
     char hasLinkCreationTs = plink ? 1 : 0;
     d->append((char*)&hasLinkCreationTs, 1);
 
-    char hasLinkAuthKey = plink ? 1 : 0;
-    d->append((char*)&hasLinkAuthKey, 1);
+    if (isExported && plink && plink->mAuthKey.size())
+    {
+        auto authKeySize = (char)plink->mAuthKey.size();
+        d->append((char*)&authKeySize, sizeof(authKeySize));
+        d->append(plink->mAuthKey.data(), authKeySize);
+    }
+    else
+    {
+        d->append("", 1);
+    }
 
     d->append("\0\0\0\0", 5); // Use these bytes for extensions
 
@@ -608,10 +612,6 @@ bool Node::serialize(string* d)
         {
             d->append((char*) &plink->cts, sizeof(plink->cts));
         }
-
-        auto authKeySize = (unsigned short)plink->mAuthKey.size();
-        d->append((char*)&authKeySize, sizeof(authKeySize));
-        d->append(plink->mAuthKey.data(), authKeySize);
     }
 
     return true;
--- a/tests/unit/Serialization_test.cpp
+++ b/tests/unit/Serialization_test.cpp
@@ -749,6 +749,28 @@ TEST(Serialization, Node_forFile_withoutShares)
     checkDeserializedNode(*dn, *n);
 }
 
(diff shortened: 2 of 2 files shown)


src/node.cpp
```diff
@@ -329,8 +328,15 @@ Node* Node::unserialize(MegaClient* client, const string* d, node_vector* dp)
     hasLinkCreationTs = MemAccess::get<char>(ptr);
     ptr += sizeof(hasLinkCreationTs);
 
-    hasLinkAuthKey = MemAccess::get<char>(ptr);
-    ptr += sizeof(hasLinkAuthKey);
+    auto authKeySize = MemAccess::get<char>(ptr);
```
sergiohs84, line 331: Please, don't use `auto` here. It's important that the datatype is `char` and not any other.

src/node.cpp
```diff
@@ -546,8 +542,16 @@ bool Node::serialize(string* d)
...
     d->append((char*)&hasLinkCreationTs, 1);
 
-    char hasLinkAuthKey = plink ? 1 : 0;
-    d->append((char*)&hasLinkAuthKey, 1);
+    if (isExported && plink && plink->mAuthKey.size())
+    {
+        auto authKeySize = (char)plink->mAuthKey.size();
+        d->append((char*)&authKeySize, sizeof(authKeySize));
+        d->append(plink->mAuthKey.data(), authKeySize);
+    }
```
sergiohs84, line 550: When you use expansion flags, you can consume only one by one, never more than one byte. Here, you are consuming 4/8 bytes for the size, plus as many bytes as the size of the authentication key.

Instead, you need to set the flag and, at the end of all the serialized data, append the size and the data.

sergiohs84 (changes requested):

src/node.cpp
mattw-mega: Hi, I saw in the email notifications that serialization is being discussed - since we are changing it for this class, please do consider switching it to use CacheableReader/CacheableWriter which would considerably simplify the code, and those functions are meant to make it easy to get it right, including use of the expansion flags, and resistant to breakages due to someone changing a type, etc.  thanks

polmr: Hi there. We were discussing this yesterday. The problem and motivation of this PR is that using some cache saved with `feature/public-writable-folders` branch, in an older version, turned into some wrong reading of numshared, and then into a crash.
In the end this happens because the former code to skip expansion flags when unserializing does this:
```
    for (i = 6; i--;)
    {
        if (ptr + (unsigned char)*ptr < end)
        {
            ptr += (unsigned char)*ptr + 1;
        }
    }
```
i.e adds the content of *ptr to the number of bytes to skip!!! Which lead me to believe that the original intention of the expansion flags was precisely that: add some memory there in between that will be skipped by former versions.
CacheableReader::unserializeexpansionflags doesn't do the same, it matches with what sergio points in his comment. i.e. something like:
```
    for (i = 6; i--;)
    {
            if (*ptr) return false;
            ptr += 1;
    }
```
To be sincere I'm kind of blocked here, what should we do:
1 - Assume that this PR is correct and the original intent on expansion flags was the one I was believing. Hence, CacheableReader::unserialize is incorrect.
2 - Assume that *ptr should always be 0 and this PR is missconception, and leave former unserialization code as is: it does something funny that leads into crashes, but that's ok since we don't intend to keep backwards compatibility.
3 - Assume that *ptr should always be 0 and this PR is missconception and assume adding *ptr is a bug. and change former unserialization code, so that unserialization fails otherwise (we can use CacheableReader), but at least it does not crash (nor loads the node in this case).
4 -Assume that *ptr should always be 0 and this PR is missconception and assume adding *ptr is a bug. and change former unserialization code, so that unserialization does not read *ptr.

1. will give backwards compatibility for this case.
2 & 3, will break backwards compatibility and will produce the same issue in further similar additions. 
4 - will breack backwards compatibility now, but will work for further expansion flags

mattw-mega: Ah, I see - so fundamentally, this class treats those flags a little differently than the class that I based the existing CacheableReader::unserializexpansionflags on.    I would say the intent of that loop that increments ptr by *ptr is designed to allow fields added in future to be skipped over by the current code (and those future fields are limited to 255 bytes in length).   However I see 'numshares', 'attrs', 'exportedLink' were all added after those flags anyway, without setting any of them (perhaps working out the length was not practical, or the 255 limit was too small?)....  so I would say that was an idea that was implemented but didn't work out.  We can't go back and use those flags because that would change the offsets for those other fields already added.  All serialized nodes in production will have those last 6 flags as \0 anyway.  And the first two of the set of 8 don't cause increments in the same way, so it's not consistent.     I would say, don't even treat those as expansion flags anymore since anyone reading in future might think we could use them that way.  Just use serializebinary()/unserializebinary() instead, with a comment that we unfortunately wasted 6 bytes a long time ago, but they must always be 0 for backward compatibility.   We could add a new set of expansion flags to the end of the record and start using those.

polmr: Aham... thanks matt for taking the time for this, 
So I will add serializebinary at the end, a comment stating the 6 bytes that need to be 0, and also will try to replace by `CacheableReader::unserializexpansionflags`  or add some sensitive comments when doing this
```
    for (i = 6; i--;)
    {
        if (ptr + (unsigned char)*ptr < end)
        {
            ptr += (unsigned char)*ptr + 1;
        }
    }
``` in the code, so that no one gets the temptation of copypasting these for a new set of extension flags ;-)

Comment (mattw-mega) [reply src/node.cpp:550]: You're welcome - I don't think we ever did advance since those 6 bytes would always have been 0 - you can just replace that block with unserializebinary(6 bytes, ignored)
It did occur to me that maybe Node serialization is included as a sub-object of larger serialized objects?  Worth checking as that might mean adding new fields at the end is problematic.  thanks

Comment (polmr) [reply src/node.cpp:550]: I have implemented the suggestions here:
https://github.com/meganz/sdk/pull/2296
I've open a separate PR because in the end it does not fix the issue:
With PR 2296, the failure unserializing a node with authkey in previous versions would occur when checking ptr reached `end` in Node::unserialize former code:
```
    if (ptr == end)
    {
        return n;
    }
    else
    {
        delete n;
        return NULL;
    }
```
Leading to a similar crash.

The way I see it, the only way to keep backwards compatibility would be to use the implementation in this PR. And assume that the use of expansions flags in this PR is correct and the only way for us to provide backwards compatibility with versions that check the end of the registry is to use expansion flags as they were used here:
```
  for (i = 6; i--;)
    {
        if (ptr + (unsigned char)*ptr < end)
        {
            ptr += (unsigned char)*ptr + 1;
        }
    }
```
but no was they are in  `CacheableReader::unserializexpansionflags`.
I leave it for your discretion to either:
A - Use implementation in base branch (public-writable-folders), which writes an expansion flag, saving on byte compared to `feature/public-writable-folder-do-not-use-exp-flags` when there is no auth key.
B - Use https://github.com/meganz/sdk/pull/2296, which disregards expansion flags and writes a string at the end, regardless if it's empty
C - Use this PR and give backwards compatibility for this particular change (which is otherwise broken for the first 2/8 existing expansion flags).
My vote is C.

sergiohs84: My vote goes for C too, as discussed on Slack. Thanks

sergiohs84 (approved):

PR #2289, opened 2020-10-15
WIP: SDK-1139. Improve folder upload/download controllers (Revamped)

jgandres:


include/mega/command.h
```diff
@@ -573,13 +573,14 @@ class MEGA_API CommandPutNodes : public Command
...
     handle targethandle;
+    std::function<void(const Error&, targettype_t , vector<NewNode>&)> mCompletion;
 
     void removePendingDBRecordsAndTempFiles();
 
 public:
     bool procresult(Result) override;
 
-    CommandPutNodes(MegaClient*, handle, const char*, vector<NewNode>&&, int, putsource_t = PUTNODES_APP, const char *cauth = NULL);
+    CommandPutNodes(MegaClient*, handle, const char*, vector<NewNode>&&, int, putsource_t = PUTNODES_APP, const char *cauth = NULL, std::function<void(const Error&, targettype_t , vector<NewNode>&)> = NULLPTR);
```
mattw-mega, line 583: Nice to see the completion functions idea being used - however, please, let's use `nullptr`.   We don't support pre-c++11 anymore so no need for this all-caps version.  thanks

src/megaclient.cpp
```diff
@@ -13158,6 +13164,19 @@ handle MegaClient::nextsyncid()
     return currsyncid;
 }
 
+// this method provides a temporal handle useful to indicate putnodes()-local parent linkage
+handle MegaClient::nextUploadId()
+{
+    byte* ptr = (byte*)&mCurrUploadId;
+
+    while (!++*ptr && ptr < (byte*)&mCurrUploadId + NODEHANDLE)
```
mattw-mega, line 13172: stopping when the incremented byte turns out to be 0?  Not sure if that is what you meant here - and, if no bytes incremented to 0 then we increment them all, which basically means we would repeat the same ID in only 256 iterations... however, I don't think we need this ID anyway, I'll comment elsewhere about that, thanks

include/mega/megaclient.h
```diff
@@ -1373,6 +1373,15 @@ class MEGA_API MegaClient
...
     std::map<handle, handle> mPublicLinks;
 
+    // temporal nodeHandle for uploads from App
+    handle mCurrUploadId;
+
+    // generates a temporal nodeHandle for uploads from App
+    handle nextUploadId();
+
+    // maps temporal nodeHandle (used for uploads from App) to definitive nodeHandle (filled by upload completion)
+    std::map<handle, handle> mTempHandleToNodeHandle;
```
mattw-mega, line 1383: I don't think we need this map - the NewNodes now have mAddedHandle so we know the handle created for a particular folder - you can store a parallel vector with info about each node created - so get the mAddedHandle from nn[i], and lookup yourDetails[i] to figure out which files to upload to that folder, in the completion function.

src/megaapi_impl.cpp
```diff
@@ -25146,12 +25057,176 @@ void MegaFolderUploadController::onTransferFinish(MegaApi *, MegaTransfer *t, Me
...
+
+    // generate fresh random key for this folder node
+    client->rng.genblock(buf,FOLDERNODEKEYLENGTH);
+    newnode.nodekey.assign((char*)buf,FOLDERNODEKEYLENGTH);
+    key.setkey(buf);
+
+    // generate fresh attribute object with the folder name
+    AttrMap attrs;
+    string sname = folderName;
+    client->fsaccess->normalize(&sname);
```
mattw-mega, line 25206: You can use MegaClient::putnodes_prepareOneFolder() to do a lot of this setup, and save repeating code here.  thanks

src/megaapi_impl.cpp
```diff
@@ -18844,11 +18846,17 @@ unsigned MegaApiImpl::sendPendingTransfers()
             fireOnTransferFinish(transfer, make_unique<MegaErrorPrivate>(e), committer);
         }
 
-        if (++count > 100 || std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - t0).count() > 100)
+        auto elapsedTime = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - t0).count();
+        if (elapsedTime > maxPendingTransfersTimeout
+                || (!mProcessingFolderTransfer && (++count > 100 || elapsedTime > minPendingTransfersTimeout)))
         {
+            // if elapsed time exceeds 3000 ms, and we are processing a folder transfer
+            // or we're not processing a folder transfer, and processed transfers exceed 100
```
mattw-mega, line 18854: Three seconds is a long time, especially if the app is trying to draw a window, and needs to wait for the mutex before it can draw something.     Since we are reworking folder uploads/downloads, I think this is the right time - we should perform the directory scanning on a thread (the thread can just exit once it's finished).   As it runs, it scans the folder and all subfolders, building up a data structure of the folders to create, and files in those folders to upload.   Whatever it scans is the snapshot that will actually be uploaded, so no race conditions or callback issues, or problems with that operation taking the mutex for too long - and we can keep the old 100ms condition here.

src/megaapi_impl.cpp
```diff
@@ -26925,51 +27223,117 @@ void MegaFolderDownloadController::downloadFolderNode(MegaNode *node, LocalPath&
...
-        transfer->setLastError(&mLastError);
-        DBTableTransactionCommitter committer(client->tctable);
-        megaApi->fireOnTransferFinish(transfer, make_unique<MegaErrorPrivate>(!mIncompleteTransfers ? API_OK : API_EINCOMPLETE), committer);
+        LocalPath &localpath = it->localPath;
+        mLastError = megaApi->createLocalFolder(localpath);
+        if (mLastError.getErrorCode() && mLastError.getErrorCode() != API_EEXIST)
+        {
+            mIncompleteTransfers++;
+            it = mLocalTree.erase(it); // remove all it's children nodes
+            mPendingFilesToProcess -= it->childrenNodes.size();
```
mattw-mega, line 27270: It looks like this might be a silent failure - shouldn't we inform the user of a failure so that they can decide how to address the problem?

mattw-mega (changes requested):

PR #2296, opened 2020-10-19
SDK-1204. Avoid using expansion flags to store auth key

polmr:
As an alternative to https://github.com/meganz/sdk/pull/2287


polmr:
Closed in favor of #2287

PR #2303, opened 2020-10-21
SDK-1262. Fix buffer overflow after call getextension

jgandres:
New PR: https://github.com/meganz/sdk/pull/2304

### DIFF
--- a/src/posix/fs.cpp
+++ b/src/posix/fs.cpp
@@ -1364,7 +1364,8 @@ bool PosixFileSystemAccess::getextension(const LocalPath& filename, char* extens
     const char* ptr = str->data() + str->size();
     char c;
 
-    size = std::min(size - 1, str->size());
+    // we need to leave space for trailing '.' and '\0'
+    size = std::min(size - 2, str->size());
 
     for (unsigned i = 0; i < size; i++)
     {


src/posix/fs.cpp
```diff
@@ -1364,7 +1364,8 @@ bool PosixFileSystemAccess::getextension(const LocalPath& filename, char* extens
     const char* ptr = str->data() + str->size();
     char c;
 
-    size = std::min(size - 1, str->size());
+    // we need to leave space for trailing '.' and '\0'
+    size = std::min(size - 2, str->size());
```
sergiohs84, line 1368: This `-2` here is out of context. This method receives the `extension` and it will fill it with the extension, up to `size -1` chars. The `-1` is because at the end it adds a `\0`.
However, the _trailing '.'_ is not added by this method, so it shouldn't be adjusted here, but at the caller. It's the caller's responsibility to do whatever it wants with the `extension`, like adding the `.` at the end (and ensuring there's space for it).

sergiohs84 (changes requested):

src/posix/fs.cpp
jgandres: PR closed, uploaded a new one: https://github.com/meganz/sdk/pull/2304

PR #2311, opened 2020-10-23
Fix Move's call to putndoes_result by avoiding it

mattw-mega:
putnodes_result should really only be called back from an actual putnodes.
In this instance, it was trying to rejoin the sequence of commands for a move,
when the move turns out to be a copy, and needs to remvoe the old node.
But, we can have much smaller and clearer code by calling unlink() directly
and also avoid ambiguities with putnodes_result, and be disentangled from it.
All putnodes_result does in this case, is skip through all the tests, and
ultimately execute these same two lines.


sergiohs84:
Closing in pro of the same solution, but applied as a hotfix here: https://github.com/meganz/sdk/pull/2307

PR #2323, opened 2020-10-26
SDK-1139. Improve folder upload/download controllers (worker thread revamped)

jgandres:


jgandres:
trigger compilation

src/megaapi_impl.cpp
```diff
@@ -18425,15 +18436,22 @@ void MegaApiImpl::updateBackups()
...
     SdkMutexGuard guard(sdkMutex);
     DBTableTransactionCommitter committer(client->tctable);
 
-    while(MegaTransferPrivate *transfer = transferQueue.pop())
+    TransferQueue &auxQueue = queue
+            ? *queue            // custom transferQueue used for folder uploads/downloads
+            : transferQueue;    // transfer queue of class MegaApiImpl
+
+    // if we are processing a custom queue, we need to process in one shot
+    bool timeout = !queue;
```
sergiohs84, line 18452: It is not a timeout only, but also the amount of transfers processed. Better to name it differently. Maybe `canSplit` or alike

src/megaapi_impl.cpp
```diff
@@ -24975,9 +24985,12 @@ void MegaFolderUploadController::start(MegaNode*)
         return;
     }
 
-    auto localpath = LocalPath::fromPath(transfer->getPath(), *client->fsaccess);
-    scanFolder(parent->getHandle(), parent->getHandle(), localpath, folderName);
-    createFolder();
+    mThread = std::thread ([this]() {
+        auto localpath = LocalPath::fromPath(transfer->getPath(), *client->fsaccess);
+        scanFolder(transfer->getParentHandle(), transfer->getParentHandle(), localpath, transfer->getFileName());
+        createFolder();
```
sergiohs84, line 24991: I think it would be easier to read if the `wait()` after createFolder and the call to `uploadFiles()` were all in the lambda for the thread.

src/megaapi_impl.cpp
```diff
@@ -25228,21 +25263,42 @@ void MegaFolderUploadController::startFileUploads(const vector<NewNode> &nn)
            vector<LocalPath> &pendingFiles = auxit->second;
            for (size_t j = 0; j < pendingFiles.size(); j++)
            {
-               pendingTransfers++;
                const LocalPath &localpath = pendingFiles.at(j);
                FileSystemType fsType = client->fsaccess->getlocalfstype(localpath);
-               megaApi->startUpload(false, localpath.toPath(*client->fsaccess).c_str(), parentNode.get(), (const char *)NULL, -1, tag, false, NULL, false, false, fsType, this);
-           }
+               MegaTransferPrivate *transfer = megaApi->createUploadTransfer(false, localpath.toPath(*client->fsaccess).c_str(),
```
sergiohs84, line 25268: Both this method `createUploadTransfer()` and the line above (`getlocalfstype()`) access to the `MegaClient::fsaccess` from the worker thread. They represent the mutual-exclusion, which needs to be ensured. In other words, we should avoid accessing MegaClient members from both threads, worker and main.

src/megaapi_impl.cpp
```diff
@@ -26406,32 +26462,35 @@ void MegaFolderDownloadController::start(MegaNode *node)
...
+        {
+            path = LocalPath::fromPath(".", *client->fsaccess);
+            path.appendWithSeparator(LocalPath::fromPath("", *client->fsaccess), true, client->fsaccess->localseparator);
+        }
 
-    FileSystemType fsType = client->fsaccess->getlocalfstype(path);
-    const LocalPath &name = (!transfer->getFileName())
-            ? LocalPath::fromName(node->getName(), *client->fsaccess, fsType)
-            : LocalPath::fromName(transfer->getFileName(), *client->fsaccess, fsType);
+        FileSystemType fsType = client->fsaccess->getlocalfstype(path);
```
sergiohs84, line 26477: Mutual exclusion violated here.

src/megaapi_impl.cpp
sergiohs84, line 24991: Mutual exclusion violated here. The `createFolder()` adds API commands to the internal queue of the SDK, which is not thread-safe.

jgandres:
trigger compilation

src/megaapi_impl.cpp
```diff
@@ -6420,6 +6439,35 @@ bool MegaApiImpl::createLocalFolder(const char *path)
...
 
+MegaErrorPrivate MegaApiImpl::createLocalFolder(LocalPath & localPath)
+{
+    SdkMutexGuard g(sdkMutex);
+    auto da = client->fsaccess->newfileaccess();
+    if (!da->fopen(localPath, true, false))
+    {
+        if (!client->fsaccess->mkdirlocal(localPath))
+        {
+            da.reset();
```
mattw-mega, line 6450: no need for these 4 calls to da.reset(), as da's destructor will always be called before exiting the function.  This is a case where we can rely on RAII to keep our code small and tidy.   thanks

PR #2324, opened 2020-10-27
SDK-1243. Add a new type of sync for Backups into SyncConfig

catalinr-m:


src/megaapi_impl.cpp
```diff
@@ -23107,6 +23148,128 @@ void MegaApiImpl::sendPendingRequests()
...
+            MegaNode* myBackupsNode = getNodeByHandle(h);
+            if (!myBackupsNode) { e = API_ENOENT; break; }
+
+            // get `DEVICE_NAME`, from user attributes
+            if (!u->isattrvalid(ATTR_DEVICE_NAMES)) { e = API_EINCOMPLETE; break; }
+            const string* deviceNameContainerStr = u->getattr(ATTR_DEVICE_NAMES);
+            if (!deviceNameContainerStr) { e = API_EINCOMPLETE; break; }
+
+            tlvRecords.reset(TLVstore::containerToTLVrecords(deviceNameContainerStr, &client->key));
+            if (!tlvRecords || !tlvRecords->find("n")) { e = API_EINCOMPLETE; break; } // is 'n' the right key ??
```
sergiohs84, line 23181: No, the right key is the `device-id`. See the corresponding PR: https://github.com/meganz/sdk/pull/2078/files#diff-5d048e8562b9eef0c31b18c647b312f4abf03481607422ffb7f1b12814524fc0R6669

src/megaapi_impl.cpp
```diff
@@ -23107,6 +23148,128 @@ void MegaApiImpl::sendPendingRequests()
...
+
+            std::unique_ptr<TLVstore> tlvRecords(TLVstore::containerToTLVrecords(handleContainerStr, &client->key));
+            if (!tlvRecords || !tlvRecords->find("h")) { e = API_ENOENT; break; }
+
+            const string& handleStr = tlvRecords->get("h");
+            handle h = MegaApi::base64ToHandle(handleStr.c_str());
+            if (h == UNDEF) { e = API_ENOENT; break; }
+
+            // get Node of remote "My Backups" folder
+            MegaNode* myBackupsNode = getNodeByHandle(h);
```
sergiohs84, line 23172: You may want to use a more efficient approach by using a `Node` instead of `MegaNode` (simply need to call `client->nodebyhandle(h)` ;)

src/megaapi_impl.cpp
```diff
@@ -23107,6 +23148,128 @@ void MegaApiImpl::sendPendingRequests()
...
+            for (int i = 0; i < nodeList->size(); ++i)
+            {
+                MegaNode* child = nodeList->get(i);
+
+                // get folder name
+                const char* name = child->getName();
+                bool gotName = deviceName == name;
+
+                // get 'device-id' tag
+                const char* id = child->getCustomAttr("device-id");
```
sergiohs84, line 23203: We should add support to create the `DEVICE_NAME` folder, which needs to set such node's attribute `"device-id"` (perhaps shorter flavor, like `"dev-id"`.

src/megaapi_impl.cpp
```diff
@@ -23107,6 +23148,128 @@ void MegaApiImpl::sendPendingRequests()
...
+                // determine whether we found our node or not
+                if (!gotName && !gotId) { continue; } // ignore this child folder
+
+                if (gotName && gotId) // got it, use this one to add another backup to it
+                {
+                    deviceNameNode = child;
+                    break;
+                }
+
+                // there's an error if only one of the folder's {name, id} is correct and the other is wrong.
```
sergiohs84, line 23215: Actually, I think it should be fine if the user decides to rename the folder. The node's attribute will allow us to identify the right one (and there should be only one). Hence, should we simply ignore the name and look for the child with tagged with the right node's attribute/device-id?

src/megaapi_impl.cpp
```diff
@@ -23107,6 +23148,128 @@ void MegaApiImpl::sendPendingRequests()
...
+                e = API_EINCOMPLETE;
+                break;
+            }
+
+            if (e != API_OK) { break; }
+
+            if (!deviceNameNode)
+            {
+                // create "My Backups"/`DEVICE_NAME`/
+                // HOW ? Like this?
```
sergiohs84, line 23225: You just need to create the folder, not a sync. It should be something similar to the code for `TYPE_CREATE_FOLDER`. Additionally, the node's attribute should be specified. Once the folder is created, you can continue to create the subfolder for the backup itself. There's a possible way to create subtrees at once (so you could create the `DEVICE_NAME` folder and the backup folder at one shot), which could be better.

src/megaapi_impl.cpp
```diff
@@ -23107,6 +23148,128 @@ void MegaApiImpl::sendPendingRequests()
...
+                auto nextSyncTag = client->nextSyncTag();
+                auto sync = make_unique<MegaSyncPrivate>(nullptr, deviceName.c_str(), h, nextSyncTag);
+                sync->setListener(request->getSyncListener());
+
+                // NOW WHAT ? Something like this?
```
sergiohs84, line 23230: Yes, I think that's the idea :+1:

src/megaapi_impl.cpp
```diff
@@ -23107,6 +23148,128 @@ void MegaApiImpl::sendPendingRequests()
...
+                e = client->addsync(syncConfig, DEBRISFOLDER, nullptr, syncError, true, sync.get());
+
+                fireOnSyncAdded(sync.get(), MegaSync::NEW);
+                fireOnRequestFinish(request, make_unique<MegaErrorPrivate>(e));
+
+                // Update deviceNameNode somehow?
+                // HOW ?
+
+                // tag the new node with `dev:<device-id>` in `attrstring`
+                // HOW ?
```
sergiohs84, line 23245: It would be something like:
``` cpp
node->attrs.map['d'] = device_id;
```
It would set the device id in the map of attributes and those attributes will be serialized and encrypted before sending to API.

src/megaapi_impl.cpp
catalinr-m: I don't understand. Either
a) There's a misunderstanding somewhere.. This is for the case when `DEVICE_NAME` existed. In which case it must have set the attribute. If `DEVICE_NAME` didn't exist, creating it is done later.
Or
b) The attribute should be `"dev-id"` instead of `"device-id"`?

src/megaapi_impl.cpp
catalinr-m: Do you mean something like the code in
```
MegaApiImpl::sendPendingRequests()

    case MegaRequest::TYPE_CREATE_FOLDER:
```
?

src/megaapi_impl.cpp
catalinr-m: So, I should "not create a sync" (as noted in the previous comment), but `client->addsync()` does take a `SyncConfig` as first, and a `MegaSyncPrivate` as 6th argument. I'm failing to match the names with the arguments and the purpose of such calls...

src/megaapi_impl.cpp
catalinr-m: Uhm, it's still not clear what would make the difference between `dev:<device-id>` and `bak:<device-id>`.

src/megaapi_impl.cpp
```diff
@@ -1491,6 +1491,47 @@ char *MegaApiImpl::getBlockedPath()
...
+        // set local path
+        string path(localFolder);
+#if defined(_WIN32) && !defined(WINDOWS_PHONE)
+        if (!PathIsRelativeA(path.c_str()) && ((path.size() < 2) || path.compare(0, 2, "\\\\")))
+            path.insert(0, "\\\\?\\");
+#endif
+        request->setFile(path.c_str());
+
+        // set backup name
+        if (backupName && *backupName)  request->setName(backupName);
```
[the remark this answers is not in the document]
catalinr-m: The checks here are only to figure out whether it should use `backupName` or the last leaf of `localFolder` (see `else` branch).

src/megaapi_impl.cpp
sergiohs84: Let me try to clarify it. The user's attribute that stores the device-id:device-name must exist, otherwise EINCOMPLETE (already done in line 23189). At this point, the DEVICE-NAME folder may or may not exist. If the latter, it needs to be created (as you plan to do in the block in line 23222).
My comment was just mentioning that this part of the puzzle needs it's counterpart: in order to consume a node's attribute with the tag `dev-id` or `device-id` (I think tags are not strings, anyway), you may want to create the folder with the node's attribute first, so you know how to consume it :)

src/megaapi_impl.cpp
sergiohs84: Yes, exactly. Perhaps not every line of code in that `case` makes sense for this use case, so you can probably simplify it. Feel free to create new methods to reuse common parts, if suitable.

src/megaapi_impl.cpp
sergiohs84: In my previous comment I was confused by looking at the comment, I was replying to the comment: 
```
// create "My Backups"/`DEVICE_NAME`/
// HOW ? Like this?
```
Creating a `SyncConfig` and a `Sync` is not the way to create such folder. Hence my comment.
But the final goal of this new interface is to create a sync, so obviously at some point you need to create the aforementioned objects.

src/megaapi_impl.cpp
catalinr-m: Sorry, I still don't understand. This code loops through existing child folders, and looks for any one having that attribute. Nothing found means that `DEVICE-NAME` child folder must be created, along with its attribute.
Should this be changed to "attempt to create `DEVICE-NAME` folder first, and if that failed, it _might_ mean that the folder already existed, so attempt to read the attribute then" ?

The API used here, namely MegaNod::getCustomAttr() does take a string as an argument. Is that the wrong one, and some other taking a different type should be used? If so, where can I find the mapping between "device-id" and a numeric id?

src/megaapi_impl.cpp
sergiohs84: Where that `bak:<device-id>` comes from?

catalinr-m: Last comment in Jira SDK-1243

sergiohs84: Oh, I see. It's just a suggestion. It should be the same thing, we are tagging the node with a node's attribute that stores the device-id. Simply use the same tag, both for the `DEVICE_NAME` folder and for the backup folder. Both folder nodes should add the `device-id` in a node's attribute. The tag is irrelevant.

src/megaapi_impl.cpp
sergiohs84: > _Should this be changed to "attempt to create DEVICE-NAME folder first, and if that failed, it might mean that the folder already existed, so attempt to read the attribute then" ?_
No, it doesn't make any sense. It's good as it is.
> _The API used here, namely MegaNod::getCustomAttr() does take a string as an argument. Is that the wrong one, and some other taking a different type should be used? If so, where can I find the mapping between "device-id" and a numeric id?_
Please, try to implement the node's attribute (the case below, where it's created, and you'll learn how to deal with node's attributes) and then you will understand how to retrieve a node's attribute in the code.

src/megaapi_impl.cpp
```diff
@@ -23166,96 +23166,118 @@ void MegaApiImpl::sendPendingRequests()
...
+                string attrstring;
+                attrs.getjson(&attrstring);
+                newnode->attrstring.reset(new string);
+                client->makeattr(&key, newnode->attrstring, attrstring.c_str());
+
+                // add the newly generated folder node
+                client->putnodes(h, move(newnodes));
+
+                // Get the new node & handle.
+                // HOW ? How can it wait until the remote folder was created?
```
sergiohs84, line 23226: You need to chain multiple API commands. Currently it's a bit tedious: the `putnodes_result()` will check the type of MegaRequest associated to the result (`TYPE_TYPE_BACKUP_FOLDER`) and proceed with the rest of the required steps (in your case, create the backup folder).
A better approach would be to adjust the putnodes to accept a lambda completion function, that can be passed to the `putnodes()` here as a parameter. Instead of the `putnodes_result()`, the `CommandPutNodes::procresult()` would call your completion function, which would proceed with the required steps.
However, the putnodes adapted for completion functions is in a branch that will take time to merge, unfortunately, so I suggest to go the old way, which perhaps is uglier, but we will avoid conflicts with the removal-SIC branch (the one that introduces the new completion function for fetchnodes).
There are examples of many `xxx_result()` that actually don't call the `fireOnRequestFinish()`, but chain another command. You can check for reference. Anyway, we can discuss on the phone, I'll show you how to do it with an existing example, if you want to.

src/megaapi_impl.cpp
```diff
@@ -23166,96 +23166,118 @@ void MegaApiImpl::sendPendingRequests()
...
+
+                // Update deviceNameNode somehow
             }
 
-            if (e != API_OK) { break; }
+            //
+            // create backupName remote dir
```
sergiohs84, line 23232: There is a way to create multiple nodes with a single putnodes, as long as they belong to the same subtree (which is the case). It may be easier and more efficient, since you won't need to wait for the creation of the `DEVICE_NAME` folder in order to create the backup folder. Both would be done at the same time, with a single `p` command.
You will need to investigate the code in order to learn how to do it, I think a good bet is to search for usages of `NewNodes`, were more than one is added to the array. I think that's the way. If you don't find the way, ping the team on the channel.

src/megaapi_impl.cpp
```diff
@@ -23166,96 +23166,118 @@ void MegaApiImpl::sendPendingRequests()
...
+                                    0, {}, true, SyncConfig::TYPE_BACKUP };
 
-            // NOW WHAT ? Something like this?
-            std::unique_ptr<char[]> remotePath{ getNodePathByNodeHandle(request->getNodeHandle()) };
-            SyncConfig syncConfig{ nextSyncTag, request->getFile(), request->getName(), deviceNameNode->getHandle(), remotePath.get(),
-                                   0, {}, true, SyncConfig::TYPE_BACKUP };
-
-            SyncError syncError = NO_SYNC_ERROR;
-            e = client->addsync(syncConfig, DEBRISFOLDER, nullptr, syncError, true, sync.get());
+            // What should be done with syncConfig ?  Something like this?
```
sergiohs84, line 23274: In order to add the new sync for the backup folder, you can use the `case MegaRequest::TYPE_ADD_SYNC:` as reference. Basically, it needs to do the same. Again, feel free to reuse the code by encapsulating common parts in new functions, called from both places

sergiohs84 (changes requested):

src/megaapi_impl.cpp
```diff
@@ -1495,6 +1495,128 @@ char *MegaApiImpl::getBlockedPath()
...
+    string localPath(request->getFile());
+    string backupName(request->getName());
+
+    if (backupName.empty()) // get the last leaf of local path
+    {
+        // trim trailing path separator(s)
+        while (!localPath.empty() && (localPath.back() == '\\' || localPath.back() == '/'))  localPath.pop_back();
+
+        // find the last non-trailing separator
+        auto sep = localPath.find_last_of("\\/");
```
sergiohs84, line 1529: We should reuse the logic from `LocalPath` here. Please, see:
- `LocalPath LocalPath::fromPath()`
- `LocalPath LocalPath::leafName()`
- `string LocalPath::toPath()`

include/mega/types.h
```diff
@@ -865,6 +865,7 @@ class SyncConfig : public Cacheable
         TYPE_UP = 0x01, // sync up from local to remote
         TYPE_DOWN = 0x02, // sync down from remote to local
         TYPE_TWOWAY = TYPE_UP | TYPE_DOWN, // Two-way sync
+        TYPE_BACKUP, // special sync up from local to remote, automatically disabled when remote changed
```
sergiohs84, line 868: Should be explicitly indicate the value?
```suggestion
        TYPE_BACKUP = 0x04, // special sync up from local to remote, automatically disabled when remote changed
```

include/megaapi.h
```diff
@@ -14132,6 +14133,18 @@ class MegaApi
         char *getBlockedPath();
 #endif
 
+        /**
+         * @brief Start backup for the given folder, to "My Backups" remote destination
+         *
+         * The backup's folder name is optional. If not provided, it will take the name of the leaf folder of
+         * the local path. In example, for "/home/user/Documents", it will become "Documents".
+         *
```
sergiohs84, line 14141: Missing documentation about the request params. Please, use doc from other interfaces as reference. It should indicate the valid params, both at any callback and when it returns API_OK. Also, it should list some of the non-obvious errors the request may return.

src/megaapi_impl.cpp
```diff
@@ -1495,6 +1495,136 @@ char *MegaApiImpl::getBlockedPath()
 }
 #endif
 
+void MegaApiImpl::backupFolder(const char *localFolder, const char *backupName, MegaRequestListener *listener)
+{
+    MegaRequestPrivate *request = new MegaRequestPrivate(MegaRequest::TYPE_BACKUP_FOLDER, listener);
+    request->setNodeHandle(0);
```
sergiohs84, line 1501: Why??

sergiohs84 (changes requested):

include/mega/types.h
catalinr-m: I don't think so. It is previous value +1, and future-proof, in case previous values change.

src/megaapi_impl.cpp
catalinr-m: It was probably a debug thing, in order to make it easier to notify if the response changed the handle. I'll remove it.

sergiohs84 (changes requested):
Pending to review the test, but @catalinr-m please, have a look to my comments.

catalinr-m:
Running the test again, it fails with
```
err: Unable to begin transaction on database: C:\Buildpath\sdk_test_mega_cache_0\megaclient_statecache12_status_VHl3VnBXVVNqTFn-TqBYyoXV59fnz6G5cd2n.db [sqlite.cpp:393]
```
I manually removed `sdk_test_mega_cache_*` folders and retried, but I got the same error. I don't know how to fix that.

I used `MegaClient::APIURL = "https://staging.api.mega.co.nz/";`

tests/integration/SdkTest_test.cpp
```diff
@@ -4684,6 +4684,61 @@ TEST_F(SdkTest, SdkGetBanners)
...
+TEST_F(SdkTest, SdkBackupFolder)
+{
+    getAccountsForTest(1);
+    LOG_info << "___TEST BackupFolder___";
+
+    //
+    // make this work even before the remote API has support for My Backups folder
+
+    // create My Backups folder
+    fetchnodes(0, maxTimeout);
```
sergiohs84, line 4696: It is performed as part of the `getAccountsForTest(1)`
```suggestion
```

tests/integration/SdkTest_test.cpp
```diff
@@ -4684,6 +4684,61 @@ TEST_F(SdkTest, SdkGetBanners)
...
+    std::unique_ptr<MegaNode> myBkpsNode{ megaApi[0]->getNodeByPath("/My Backups") };
```
sergiohs84, line 4697: The node relative to the `/My Backups` folder is stored in a user's attribute and it needs to be retrieved from the user's attribute, not by searching by folder name.
Note that there might be a user's attribute pointing to the node, but the node could have been deleted. So you will need to check not only that the user's attribute exists and contains a node's handle, but also that the node's handle exists.

tests/integration/SdkTest_test.cpp
```diff
@@ -4684,6 +4684,61 @@ TEST_F(SdkTest, SdkGetBanners)
...
+        // set My Backups handle attr
+        MegaHandle mh = myBkpsNode->getHandle();
+        int err = synchronousSetMyBackupsFolder(0, mh);
+        ASSERT_TRUE(err == MegaError::API_OK) << "Setting My Backup folder failed (error: " << err << ")";
+        err = synchronousGetMyBackupsFolder(0);
+        ASSERT_TRUE(err == MegaError::API_OK);
+
+        // read the attribute to make sure it was set
+        std::unique_ptr<MegaUser> user(megaApi[0]->getMyUser());
+        ASSERT_NO_FATAL_FAILURE(getUserAttribute(user.get(), MegaApi::USER_ATTR_MY_BACKUPS_FOLDER, maxTimeout, 0));
```
sergiohs84, line 4717: There is a dedicated interface to `getMyBackupsFolder()` that should be used, instead of the generic one for any user attribute (which should work as well, but may not set the MegaRequest members in the same way than the specific one).
You've already called it in line 4712, so you just need to check the value is the intended one. No need to call `getUserAttribute()`.

tests/integration/SdkTest_test.cpp
```diff
@@ -4684,6 +4684,61 @@ TEST_F(SdkTest, SdkGetBanners)
...
+
+        // look for Device Name attr
+        if (synchronousGetDeviceName(0) != MegaError::API_OK || attributeValue.empty())
```
sergiohs84, line 4720: This block should probably be out of the `if (!myBkpsNode)`. It may happen that there's a `/My Backups`, but there's no `/My Backups/<device_name>`

tests/integration/SdkTest_test.cpp
```diff
@@ -4684,6 +4684,61 @@ TEST_F(SdkTest, SdkGetBanners)
...
+            strftime(timebuf, sizeof timebuf, "%c", localtime(&now));
+
+            string deviceName = string("Jenkins ") + timebuf;
+            synchronousSetDeviceName(0, deviceName.c_str());
+        }
+    }
+
+    // make sure Device Name attr was set
+    int err = synchronousGetDeviceName(0);
+    ASSERT_TRUE(err == MegaError::API_OK) << "Getting device name attr failed (error: " << err << ")";
```
sergiohs84, line 4733: It would be good to check the value retrieved by `getDeviceName()` actually matches the expected one (at least when it's created, since it's random :) )

sergiohs84 (changes requested):

tests/integration/SdkTest_test.cpp
```diff
@@ -4684,6 +4684,61 @@ TEST_F(SdkTest, SdkGetBanners)
...
+
+    // request to backup a folder
+    string folderToBackup = string(LOCAL_TEST_FOLDER) + "\\LocalBackedUpFolder";
+    makeNewTestRoot(folderToBackup.c_str());
+    err = synchronousBackupFolder(0, folderToBackup.c_str(), "RemoteBackupFolder");
+    ASSERT_TRUE(err == MegaError::API_OK) << "Backup folder failed (error: " << err << ")";
```
sergiohs84, line 4739: If possible, it would be good to check the node's attribute of the folder for `<device_name>`, which should contain the `dev-id:<device_id>`. Same for the backup folder itself.
(Remember this comment: https://github.com/meganz/sdk/pull/2324#discussion_r513436961)

src/megaapi_impl.cpp
```diff
@@ -1495,6 +1495,136 @@ char *MegaApiImpl::getBlockedPath()
...
+    // get handle of remote "My Backups" folder, from user attributes
+    if (!u || !u->isattrvalid(ATTR_MY_BACKUPS_FOLDER)) { return API_EACCESS; }
+    const string* handleContainerStr = u->getattr(ATTR_MY_BACKUPS_FOLDER);
+    if (!handleContainerStr) { return API_EACCESS; }
+
+    std::unique_ptr<TLVstore> tlvRecords(TLVstore::containerToTLVrecords(handleContainerStr, &client->key));
+    if (!tlvRecords || !tlvRecords->find("h")) { return API_EINTERNAL; }
+
+    std::unique_ptr<MegaStringMap> stringMap(new MegaStringMapPrivate(tlvRecords->getMap(), true));
+    const char *handleStr = stringMap->get("h");
```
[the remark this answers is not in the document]
catalinr-m: Ok. Can you please tell me how to get the actual `handle` from the value returned by `tlvRecords->get("h")`? I remember trying a bunch of `Base64::` transformations and never got the right value. I attempted that again and still without any luck. Surely I must have missed the right call, but using the intermediary `MegaStringMap` just made that work..

Mind that current value of `handleStr`, when passed to `Base64::atob()` gets the `handle` value, while after the proposed change, the value is completely different. Others examples that I could find, also got the value from a `MegaStringMap`.

src/megaapi_impl.cpp
```diff
@@ -13931,6 +14065,31 @@ void MegaApiImpl::putnodes_result(const Error& inputErr, targettype_t t, vector<
...
+            std::unique_ptr<char[]> remotePath{ getNodePathByNodeHandle(backupHandle) };
+            SyncConfig syncConfig{ nextSyncTag, nullptr, deviceName, backupHandle, remotePath.get(),
+                                    0, {}, true, SyncConfig::TYPE_BACKUP };
+
+            const char* backupName = request->getName();
+            const char* localPath = request->getFile();
+
+            // create the Sync
+            auto sync = make_unique<MegaSyncPrivate>(localPath, backupName, backupHandle, nextSyncTag);
+            sync->setListener(request->getSyncListener());
```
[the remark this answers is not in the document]
catalinr-m: Shouldn't it remain null when not initially set?
It's not clear to me whether this is actually not needed. I've seen this being done in `case MegaRequest::TYPE_ADD_SYNC`.
Shall I remove it?

src/megaapi_impl.cpp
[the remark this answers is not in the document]
sergiohs84: You're right. The `stringMap` you created converts the `Node:attr` values to B64, so you get it in B64 in `handleStr`.
My suggestion means that we can directly do:
``` cpp
handle h = 0;
memset(&h, tlvRecords->get(h), sizeof h);
```

src/megaapi_impl.cpp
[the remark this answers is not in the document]
sergiohs84: It's unclear to me too. I think the `MegaSyncListener` might had become obsolete at some point and it is still around in the code. I don't see any place where it gets set. And the app seems to not use it either. @vtmateos can we remove traces of this listener from the SDK?

tests/integration/SdkTest_test.cpp
sergiohs84: As discussed, we need:
1. New `const char *MegaApi::getDeviceId()`
2. New property to store `string MegaNodePrivate::mDeviceId;`
3. New method to get `const char *MegaNode::getDeviceId()`
4. Add the corresponding checks in the test.

include/megaapi_impl.h
```diff
@@ -2406,7 +2406,7 @@ class MegaApiImpl : public MegaApp
         //Sync
         int syncPathState(string *path);
         MegaNode *getSyncedNode(const LocalPath& path);
-        void syncFolder(const char *localFolder, const char *name, MegaHandle megaHandle, MegaRegExp *regExp = NULL, MegaRequestListener* listener = NULL);
+        void syncFolder(const char *localFolder, const char *name, MegaHandle megaHandle, SyncConfig::Type backupType, MegaRegExp *regExp = NULL, MegaRequestListener* listener = NULL);
```
catalinr-m, line 2412: It might be convenient to default the new parameter to a value, i.e. `SyncConfig::TYPE_TWOWAY`

src/megaapi.cpp
```diff
@@ -3300,12 +3300,12 @@ void MegaApi::syncFolder(const char *localFolder, MegaNode *megaFolder, MegaRequ
 
 void MegaApi::syncFolder(const char *localFolder, const char *name, MegaHandle megaHandle, MegaRequestListener *listener)
 {
-    pImpl->syncFolder(localFolder, name, megaHandle, NULL, listener);
+    pImpl->syncFolder(localFolder, name, megaHandle, SyncConfig::Type::TYPE_TWOWAY, NULL, listener);
```
catalinr-m, line 3303: I can do without mentioning `Type::` scope. Not sure which is more or less readable though.

src/megaapi.cpp
```diff
@@ -3300,12 +3300,12 @@ void MegaApi::syncFolder(const char *localFolder, MegaNode *megaFolder, MegaRequ
...
 }
 
 void MegaApi::syncFolder(const char *localFolder, MegaHandle megaHandle, MegaRequestListener *listener)
 {
-    pImpl->syncFolder(localFolder, nullptr, megaHandle, NULL, listener);
+    pImpl->syncFolder(localFolder, nullptr, megaHandle, SyncConfig::Type::TYPE_TWOWAY, NULL, listener);
```
catalinr-m, line 3308: ..same here

src/megaapi_impl.cpp
```diff
@@ -14077,9 +14067,9 @@ void MegaApiImpl::putnodes_result(const Error& inputErr, targettype_t t, vector<
             fireOnRequestFinish(request, make_unique<MegaErrorPrivate>(API_OK));    // even if import fails, notify account was successfuly created anyway
             return;
         }
-
-        else if (request->getType() == MegaRequest::TYPE_BACKUP_FOLDER && e == API_OK)
+        else if (request->getType() == MegaRequest::TYPE_ADD_SYNC && e == API_OK)
         {
+            // folders for the new backup has been created -> now proceed with the sync
```
catalinr-m, line 14072: `has` -> `have`

src/megaapi.cpp
catalinr-m: ..and below, in a couple more places

catalinr-m (commented):
There's one more build error, not sure why it did not show up until now:
```
desktop\src\MEGASync\mega\src\megaapi_impl.cpp(24723): error C2664: 'void mega::MegaSyncPrivate::setType(mega::MegaSync::SyncType)': cannot convert argument 1 from 'int' to 'mega::MegaSync::SyncType'
desktop\src\MEGASync\mega\src\megaapi_impl.cpp(24723): note: Conversion to enumeration type requires an explicit cast (static_cast, C-style cast or function-style cast)
```

Other than that, the tests are still passing, so they should be fine.

src/megaapi.cpp
sergiohs84: There's already a mix of them used all around. For simplicity, I'd keep it as `SyncConfig::TYPE_XYZ` ;)

include/megaapi_impl.h
sergiohs84: I've intentionally make it mandatory in order to avoid setting `TWOWAY` by mistake for a backup

catalinr-m:
This must be after the merge:

```
desktop\src\MEGASync\mega\src\megaapi_impl.cpp(14075): error C2039: 'setListener': is not a member of 'mega::MegaSyncPrivate'
desktop\src\MEGASync\mega\include\megaapi_impl.h(1069): note: see declaration of 'mega::MegaSyncPrivate'
desktop\src\MEGASync\mega\src\megaapi_impl.cpp(14075): error C2039: 'getSyncListener': is not a member of 'mega::MegaRequestPrivate'
desktop\src\MEGASync\mega\include\megaapi_impl.h(1149): note: see declaration of 'mega::MegaRequestPrivate'
```

There is also a missing include file, signaled by cmake too:
```
CMake Error at CMakeLists.txt:595 (add_executable):
  Cannot find source file:

    desktop/src/MEGASync/mega/bindings/qt/QTMegaSyncListener.h
```

Are you still making updates? Shall I attempt to fix them?

src/megaapi_impl.cpp
```diff
@@ -1493,6 +1502,117 @@ char *MegaApiImpl::getBlockedPath()
...
+        request->setFile(localPath.c_str());
+    }
+#endif
+
+    // get current user
+    User* u = client->ownuser();
+
+    // get handle of remote "My Backups" folder, from user attributes
+    if (!u || !u->isattrvalid(ATTR_MY_BACKUPS_FOLDER)) { return API_EACCESS; }
+    const string* handleContainerStr = u->getattr(ATTR_MY_BACKUPS_FOLDER);
```
mattw-mega, line 1541: At what point do we actually create the "My Backups" folder and store its handle in the ATTR_MY_BACKUPS_FOLDER attribute?   I would have thought we would do it here if it didn't exist already?   thanks

src/megaapi_impl.cpp
```diff
@@ -1493,6 +1502,117 @@ char *MegaApiImpl::getBlockedPath()
...
+    const string& deviceId = client->getDeviceid();
+    if (deviceId.empty()) { return API_EINCOMPLETE; }
+
+    // get `DEVICE_NAME`, from user attributes
+    if (!u->isattrvalid(ATTR_DEVICE_NAMES)) { return API_EINCOMPLETE; }
+    const string* deviceNameContainerStr = u->getattr(ATTR_DEVICE_NAMES);
+    if (!deviceNameContainerStr) { return API_EINCOMPLETE; }
+
+    tlvRecords.reset(TLVstore::containerToTLVrecords(deviceNameContainerStr, &client->key));
+    if (!tlvRecords || !tlvRecords->find(deviceId)) { return API_EINCOMPLETE; }
```
mattw-mega, line 1566: I thought we would be adding it here if it wasn't already present?  thanks

src/megaapi_impl.cpp
```diff
@@ -1493,6 +1502,117 @@ char *MegaApiImpl::getBlockedPath()
...
+        if (deviceNameNode->type != FOLDERNODE) { return API_EACCESS; }
+
+        // get 'dev-id' tag
+        const auto& childAttrs = deviceNameNode->attrs.map;
+        auto found = childAttrs.find(attrId);
+
+        // a folder without the attribute, or with it but with different value, will not be accessed
+        if (found == childAttrs.end() || found->second != deviceId) { return API_EACCESS; }
+    }
+    else // create `DEVICE_NAME` remote dir
```
mattw-mega, line 1591: Since the device nodes can be renamed, shouldn't we be iterating the subfolders of "My Backups" and looking for one with the matching dev-id, since that seems to be our definition of matching the device with the folder?  thanks

include/megaapi.h
```diff
@@ -11935,6 +11963,15 @@ class MegaApi
...
         void setRubbishBinAutopurgePeriod(int days, MegaRequestListener *listener = NULL);
 
+        /**
+         * @brief Returns the id of this device
+         *
+         * You take the ownership of the returned value.
+         *
+         * @return The id of this device
+         */
+        const char* getDeviceId() const;
```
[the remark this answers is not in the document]
catalinr-m: This was done only to aid with test code. But I'll let @sergiohs84 decide.

src/megaapi_impl.cpp
catalinr-m: I don't remember where exactly it was written, but the idea was that the folder at the remote location and user attr would already exist when the user will have the possibility of creating backups.

src/megaapi_impl.cpp
catalinr-m: Good point. This should probably be changed to use your suggestion. @sergiohs84 what do you think?

include/megaapi.h
```diff
@@ -5172,6 +5183,15 @@ class MegaSync
         NEW_TEMP_DISABLED = 6, // new sync added as temporarily disabled due to a temporary error
     };
 
+    enum SyncType
+    {
+        TYPE_UNKNOWN = 0x00,
+        TYPE_UP = 0x01, // sync up from local to remote
+        TYPE_DOWN = 0x02, // sync down from remote to local
```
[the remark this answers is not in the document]
sergiohs84: Well, we could comment them out, but then we would need to hardcode the value for `TWO_WAY`.
And in practice, the official apps won't start using those types randomly. Also, the `MegaSync` objects seen by the app are always provided by the SDK, so none of them will have `UP` or `DOWN`, since there's no MegaApi method to create such config :) 
The only inconvenient of keeping UP and DOWN here is... the fake impression that we support one way sync, while we don't. I don't see it a problem.

include/megaapi.h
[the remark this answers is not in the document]
sergiohs84: Clients will add the ability to set the device name in a near future. Hence, they will need to check their own device id.
Imagine a list of devices for the user to see the last session ts, country... (like in Security in Webclient). The idea is that, instead of anonymous sessions, they map to the device id, and the user attribute that stores the names given to each device, will be used to ease the identification of devices by the user. At some point, the app will need to know which is its own device id.

src/megaapi_impl.cpp
sergiohs84: It was agreed with MEGAsync TeamLeader that:

> ...upon creation of a new backup, the app should:
> - Have created the "My Backups" folder first (it needs to be localised) and set the corresponding attribute by calling setMyBackupsFolder()
> - Have set the device name in advance by calling setDeviceName() , even if it's a name given by default

include/megaapi.h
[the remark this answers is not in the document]
sergiohs84: Actually, thinking twice about it, I don't see the need to expose the device id for tests. Instead of calling the generic `getUserAttribute()` and then search for your device id, you could use the dedicated getter `getDeviceName()`, which directly returns the name of your device.
Nevertheless, I think it's not a problem to keep it, and can be beneficial soon. Up to you, I don't have an strong opinion on this one, but I'd keep it.

src/megaapi_impl.cpp
sergiohs84: Yep, same decision than here: https://github.com/meganz/sdk/pull/2324/files/7d6155a379d9d6bfe14fd2c627e46b937b78e723#r540876333

No need to change it

src/megaapi_impl.cpp
sergiohs84: Matt is right. The `<device_name>` folder should be found among children of `My Backups`, but by checking the device id in the node's attribute, not by checking the name.
Maybe a getter `MegaHandle getDeviceNameFolder()` might be useful.

catalinr-m: Ok, I'll make that change

sergiohs84: Just for your information: 
- if the user changes the device name for a device id, the folders won't be renamed.
- if the user renames the `<device_name>` folder, the device name in the user attribute won't be updated.

catalinr-m: > Maybe a getter MegaHandle / MegaNode getDeviceNameFolder() might be useful.

I made the change in a slightly different way, because of the different error codes that could be encountered on the way.

tests/integration/SdkTest_test.cpp
```diff
@@ -4696,6 +4704,117 @@ TEST_F(SdkTest, SdkGetBanners)
...
+    LOG_info << "___TEST BackupFolder___";
+
+    // Attempt to get My Backups folder;
+    // make this work even before the remote API has support for My Backups folder
+    synchronousGetMyBackupsFolder(0);
+    MegaHandle mh = mApi[0].h;
+
+    // create My Backups folder
+    unique_ptr<MegaNode> myBkpsNode{ mh && mh != UNDEF ? megaApi[0]->getNodeByHandle(mh)
+                                                       : megaApi[0]->getNodeByPath("/My Backups") };
```
sergiohs84, line 4719: The `My Backups` folder should be retrieved exclusively from the `mh` handle, not by node's name.
Also, the `UNDEF` is internal. Here, you should use `INVALID_HANDLE` instead
```suggestion cpp
    unique_ptr<MegaNode> myBkpsNode{ mh && mh != INVALID_HANDLE ? megaApi[0]->getNodeByHandle(mh) : nullptr };
```

sergiohs84: Actually, taking into account the `myBkpsNode` is only here, I'd say the `if (!myBkpsNode)` could be replaced directly by:
``` cpp
if (unique_ptr<MegaNode> n = megaApi[0]->getNodeByHandle(mh))
```

tests/integration/SdkTest_test.cpp
```diff
@@ -4696,6 +4704,117 @@ TEST_F(SdkTest, SdkGetBanners)
...
+        synchronousSetDeviceName(0, deviceName.c_str());
+
+        // make sure Device Name attr was set
+        int err = synchronousGetDeviceName(0);
+        ASSERT_TRUE(err == MegaError::API_OK) << "Getting device name attr failed (error: " << err << ")";
+        ASSERT_EQ(deviceName, attributeValue) << "Getting device name attr failed (wrong value)";
+    }
+
+    // request to backup a folder
+    string folderToBackup = string(LOCAL_TEST_FOLDER) + "\\LocalBackedUpFolder";
```
sergiohs84, line 4761: Perhaps it's better to name it `localFolderPath`

tests/integration/SdkTest_test.cpp
```diff
@@ -4696,6 +4704,117 @@ TEST_F(SdkTest, SdkGetBanners)
...
+    makeNewTestRoot(folderToBackup.c_str());
+    mApi[0].h = 0;
+    const char* remoteBackup = "RemoteBackupFolder";
```
sergiohs84, line 4764: Perhaps it's better to name it `backupName`

tests/integration/SdkTest_test.cpp
```diff
@@ -4696,6 +4704,117 @@ TEST_F(SdkTest, SdkGetBanners)
...
+    unique_ptr<char[]> actualRemotePath{ megaApi[0]->getNodePathByNodeHandle(mApi[0].h) };
+    ASSERT_EQ(expectedRemotePath, actualRemotePath.get()) << "Wrong remote path for backup";
+
+    // Verify that the sync was added
+    unique_ptr<MegaSyncList> allSyncs{ megaApi[0]->getSyncs() };
+    ASSERT_TRUE(allSyncs && allSyncs->size()) << "API reports 0 Sync instances";
+    MegaSync* megaSync = allSyncs->get(0);
+    ASSERT_EQ(string(remoteBackup), megaSync->getName()) << "Sync instance points to a backup with wrong name";
+    ASSERT_EQ(string(actualRemotePath.get()), megaSync->getMegaFolder()) << "Sync instance points to wrong remote path";
+    ASSERT_EQ(mApi[0].h, megaSync->getMegaHandle()) << "Sync instance points to wrong MegaHandle";
```
sergiohs84, line 4786: Instead of assuming that your sync will be the only one (today perhaps that's guaranteed, but not necessarily in the future), I think it would be better to iterate through the list, and find the one/s that have `MegaSync::getType() == TYPE_BACKUP`, then check for the name/path/handle has you do. If no sync of type backup matches, then fail.

tests/integration/SdkTest_test.cpp
```diff
@@ -4696,6 +4704,117 @@ TEST_F(SdkTest, SdkGetBanners)
...
+
+    // Verify sync after logout / login
+    string session = dumpSession();
+    locallogout();
+    auto tracker = asyncRequestFastLogin(0, session.c_str());
+    ASSERT_EQ(API_OK, tracker->waitForResult()) << " Failed to establish a login/session for accout " << 0;
```
sergiohs84, line 4792:
```suggestion
    ASSERT_EQ(API_OK, tracker->waitForResult()) << " Failed to establish a login/session for account " << 0;
```

sergiohs84 (changes requested):
We are almost there, @catalinr-m. Please, have a look to the open discussions, and let's try to get it ready for the merge. Thanks

include/megaapi.h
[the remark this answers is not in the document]
catalinr-m: Is there anything to change after this discussion?

include/megaapi.h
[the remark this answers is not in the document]
catalinr-m: Is there anything that must be changed after this discussion?

tests/integration/SdkTest_test.cpp
catalinr-m: How would this work? It needs only the negated condition.

tests/integration/SdkTest_test.cpp
catalinr-m: renamed

tests/integration/SdkTest_test.cpp
catalinr-m: renamed

tests/integration/SdkTest_test.cpp
catalinr-m: Updated.
I thought that tests had a guaranteed initial state, which included _no syncs_.

include/megaapi.h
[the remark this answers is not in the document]
mattw-mega: Ok, no need to change it, let's see how it plays out.  thanks

include/megaapi.h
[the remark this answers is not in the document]
mattw-mega: Ok let's leave it as is for now, thanks

tests/integration/SdkTest_test.cpp
sergiohs84: Usually that's true, tests start with a clean state. However, the cache of configured syncs persist from one session to another. So in case a test fails/crashes and leaves leftovers, it may happen that some other syncs are restored from cache, potentially leading to a fake alarm on your test.
Thanks for the update

sergiohs84 (approved):

Daryle1K (commented):
I don’t know how to do nun of this

PR #2327, opened 2020-10-30
SDK-1273. Keep subtransfer resumption after exit/locallogout

polmr:
avoid stopping subtransfers when aborting actions.
that entailed the removal of subtransfers from cache.
abortactions will now only cause a fireOnTransferFinish and its
consequent deletion of the MegaTransferPrivate object, keeping the cache
& temporary files untouched.


jgandres (commented):
It's needed to implement same checkups in MegaFolderDownloadController::cancel in MegaFolderUploadController class.

mattw-mega (approved):

PR #2331, opened 2020-11-03
SDK-1255. Google Ads Support

alber2510:


include/megaapi.h
```diff
@@ -18285,6 +18318,25 @@ class MegaApi
         */
         void sendBackupHeartbeat(MegaHandle backupId, int status, int progress, int ups, int downs, long long ts, MegaHandle lastNode);
 
+        /**
+         * @brief Allow fetch google ads
```
sergiohs84, line 18322:
```suggestion
         * @brief Fetch Google ads
```
Additionally, the doc about the request and its params when API_OK is missing.

include/megaapi.h
```diff
@@ -18285,6 +18318,25 @@ class MegaApi
...
+         *
+         * @param adFlags A bitmap flag used to communicate with the API
+         * @param adUnits A list of the adslot ids to fetch
+         * @param publicHandle Provide the public handle that the user is visiting (optionally)
+         * @param listener MegaRequestListener to track this request
+         */
+        void fetchGoogleAds(int adFlags, MegaStringList *adUnits, MegaHandle publicHandle, MegaRequestListener *listener = nullptr);
+
+        /**
+         * @brief Allows you to simply query if we should show an ad or not
```
sergiohs84, line 18332:
```suggestion
         * @brief Check if Google ads should show or not
```
Also, missing doc about the request. And we probably should include some context and possible values for `adFlags`.

include/megaapi.h
```diff
@@ -18285,6 +18318,25 @@ class MegaApi
         */
         void sendBackupHeartbeat(MegaHandle backupId, int status, int progress, int ups, int downs, long long ts, MegaHandle lastNode);
 
+        /**
+         * @brief Allow fetch google ads
+         *
+         * @param adFlags A bitmap flag used to communicate with the API
+         * @param adUnits A list of the adslot ids to fetch
+         * @param publicHandle Provide the public handle that the user is visiting (optionally)
```
sergiohs84, line 18326: Since it's optional, it could be useful to set a default value in the signature (aka. `publicHandle = INVALID_HANDLE`).
Same for the other interface.

src/commands.cpp
```diff
@@ -8335,4 +8335,115 @@ bool CommandDismissBanner::procresult(Result r)
     return r.wasErrorOrOK();
 }
 
+bool CommandFetchGoogleAds::procresult(Command::Result r)
+{
+    string_map result;
+    if (r.wasErrorOrOK())
+    {
+        mCompletion(r.errorOrOK(), result);
```
sergiohs84, line 8343: Is it possible that the command returns no ads, so it simply returns an `API_OK`?? <-- `[0]`
If so, then this code is fine. If not possible, then we may want to check for strictly error and return an API_INTERNAL if not error or object with ads.

src/commands.cpp
```diff
@@ -8335,4 +8335,115 @@ bool CommandDismissBanner::procresult(Result r)
...
+                        result[id] = iu;
+                    }
+                    else
+                    {
+                        error = true;
+                        result.clear();
+                    }
+                    break;
+
+                default:
```
sergiohs84, line 8379: Here you need to consume the rest of the JSON object. Imagine the API decides to add extra fields other than `id` and `iu`. The parser should ignore them, but finish the parsing of it, so the outer loop keeps reading the next object.

src/commands.cpp
```diff
@@ -8335,4 +8335,115 @@ bool CommandDismissBanner::procresult(Result r)
...
+                    client->json.storeobject(&id);
+                    break;
+
+                case MAKENAMEID2('i', 'u'):
+                    client->json.storeobject(&iu);
+                    break;
+
+                case EOO:
+                    exit = true;
+                    if (id.size())
```
sergiohs84, line 8368:
```suggestion
                    if (!id.empty() && !iu.empty())
```

src/commands.cpp
```diff
@@ -8335,4 +8335,115 @@ bool CommandDismissBanner::procresult(Result r)
...
+    {
+        arg("p", publicHandle);
+    }
+
+    beginarray("au");
+    for (const std::string& adUnit : adUnits)
+    {
+        element(adUnit.c_str());
+    }
+
```
sergiohs84, line 8409:
```suggestion
```

src/commands.cpp
```diff
@@ -8335,4 +8335,115 @@ bool CommandDismissBanner::procresult(Result r)
...
+                    break;
+            }
+        }
+
+        client->json.leaveobject();
+    }
+
+    mCompletion((error ? API_EINTERNAL : API_OK), result);
+
+    return true;
```
sergiohs84, line 8389: If there's been an error parsing the command, I think the right return value is false.

src/megaapi_impl.cpp
```diff
@@ -4128,46 +4145,29 @@ MegaStringMapPrivate::MegaStringMapPrivate(const MegaStringMapPrivate *megaStrin
...
-    for (int i = 0; i < s; i++)
-        list[i] = MegaApi::strdup(stringList->get(i));
+    mList = stringList->mList;
 }
 
 MegaStringListPrivate::MegaStringListPrivate(char **newlist, int size)
 {
-    list = NULL;
-    s = size;
     if (!size)
```
sergiohs84, line 4157: We can also get rid of this `if (!size)` :)

src/megaapi_impl.cpp
```diff
@@ -23073,6 +23081,45 @@ void MegaApiImpl::sendPendingRequests()
                                                            request->getNodeHandle()));
             break;
         }
+        case MegaRequest::TYPE_FETCH_GOOGLE_ADS:
+        {
+                MegaStringList* stringList = request->getMegaStringList();
+                std::vector<std::string> vectorString;
```
sergiohs84, line 23087: Maybe, instead of adding a copy, we could add a getter at the private class, and get the `MegaRequestPrivate::mList`  directly thanks to a casting+getter, avoiding the copy.

src/megaapi_impl.cpp
```diff
@@ -23073,6 +23081,45 @@ void MegaApiImpl::sendPendingRequests()
...
+                for (int i = 0; i < stringList->size(); i++)
+                {
+                    vectorString.push_back(stringList->get(i));
+                }
+
```
sergiohs84, line 23092: Once you add an enumeration with the possible `adFlags`, we may want to add sanity checks here.

src/megaapi_impl.cpp
```diff
@@ -23073,6 +23081,45 @@ void MegaApiImpl::sendPendingRequests()
...
+                        request->setMegaStringMap(stringMap.get());
+                    }
+
+                    fireOnRequestFinish(request, make_unique<MegaErrorPrivate>(e));
+                }));
+            break;
+        }
+        case MegaRequest::TYPE_QUERY_GOOGLE_ADS:
+        {
```
sergiohs84, line 23111: Sanity checks for flags

sergiohs84 (changes requested):
Additionally to my comments, we should add a test to exercise this 2 commands, if possible

src/megaapi_impl.cpp
alber2510: Fixed

src/megaapi_impl.cpp
alber2510: Fixed

src/commands.cpp
alber2510: I only set a check for `id` because I thought that it is more relevant for parsing information but use both condition I think it's ok

sergiohs84 (approved):

PR #2333, opened 2020-11-04
SDK-1269. Notify when API sends the node to the rubbish bin

sergiohs84:
When the target node of a putnode does not exist, the API will put the node into the rubbish bin.


sergiohs84:
Failed execution in Jenkins is not related to this PR, but a failed assert related to heartbeats

mattw-mega (approved):

PR #2337, opened 2020-11-06
SDK-1284. DB Tweaks / Improvements

dwmega:


src/db/sqlite.cpp
```diff
@@ -319,7 +395,17 @@ void SqliteDbTable::commit()
     }
 
     LOG_debug << "DB transaction COMMIT " << dbfile;
-    sqlite3_exec(db, "COMMIT", 0, 0, NULL);
+
+    if (sqlite3_get_autocommit(db))
```
sergiohs84, line 399: Wouldn't it be a programmer's fault if the code calls `commit()` on a non-transactional mode DB?
This addition will hide that mistake.
I guess that you added the call because it's already occurring?

src/db/sqlite.cpp
```diff
@@ -341,10 +437,7 @@ void SqliteDbTable::remove()
         return;
     }
 
-    if (pStmt)
-    {
-        sqlite3_finalize(pStmt);
-    }
+    sqlite3_finalize(pStmt);
```
sergiohs84, line 440: From documentation:
> ^Invoking sqlite3_finalize() on a NULL pointer is a harmless no-op.

Good spotting ;)

tests/unit/Sync_test.cpp
```diff
@@ -1300,13 +1300,18 @@ class MockDbAccess : public mega::DbAccess
     MockDbAccess(std::vector<std::pair<uint32_t, std::string>>& data)
         : mData{data}
     {}
-    mega::DbTable* open(mega::PrnGen &rng, mega::FileSystemAccess*, std::string*, bool recycleLegacyDB, bool checkAlwaysTransacted) override
+    mega::DbTable* open(mega::PrnGen &rng, mega::FileSystemAccess&, const std::string&, const int flags) override
     {
-        auto table = new MockDbTable{rng, checkAlwaysTransacted};
+        auto table = new MockDbTable{rng, (flags & mega::DB_OPEN_FLAG_RECYCLE) > 0};
```
sergiohs84, line 1305:
```suggestion
        auto table = new MockDbTable{rng, (flags & mega::DB_OPEN_FLAG_TRANSACTED) > 0};
```

sergiohs84 (changes requested):

src/db/sqlite.cpp
dwmega: Aye, I added the guard as commit() is being called unconditionally in some places.

Perhaps a better solution would have been to expose a function that allows the client to query whether the database is currently within a transaction. That way, we could call commit only if necessary.

That'd solve this problem nicely.

tests/unit/Sync_test.cpp
dwmega: Nice catch!

sergiohs84 (approved):

sergiohs84 (approved):

PR #2349, opened 2020-11-10
SDK-1272. Recheck MAC on download MAC failure, in case it was the result of a m…

mattw-mega:
…ac calculation race

In rare cases, some uploads were missing the macs of one of the final upload pieces in the mac-of-macs calculation.
Recognize these and reconfirm the MAC, approve the download.

### DIFF
--- a/src/transferslot.cpp
+++ b/src/transferslot.cpp
@@ -347,7 +347,100 @@ int64_t TransferSlot::macsmac(chunkmac_map* m)
     return m->macsmac(transfer->transfercipher());
 }
 
-bool TransferSlot::checkTransferFinished(DBTableTransactionCommitter& committer, MegaClient* client)
+int64_t TransferSlot::macsmac_gaps(chunkmac_map* m, size_t g1, size_t g2, size_t g3, size_t g4)
+{
+    return m->macsmac_gaps(transfer->transfercipher(), g1, g2, g3, g4);
+}
+
+bool TransferSlot::checkMetaMacWithMissingLateEntries()
+{
+    // Due to an old bug, some uploads attached a MAC to the node that was missing some MAC entries 
+    // (even though the data was uploaded) - this occurred when a ultoken arrived but one other 
+    // final upload connection had not completed at the local end (even though it must have 
+    // completed at the server end).  So the file's data is still complete in the cloud.
+    // Here we check if the MAC is one of those with a missing entry (or a few if the connection had multiple chunks)
+
+    // last 3 connections, up to 32MB (ie chunks) each, up to two completing after the one that delivered the ultoken
+    size_t end = transfer->chunkmacs.size();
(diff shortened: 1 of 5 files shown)


src/transferslot.cpp
```diff
@@ -347,7 +347,100 @@ int64_t TransferSlot::macsmac(chunkmac_map* m)
...
+{
+    // Due to an old bug, some uploads attached a MAC to the node that was missing some MAC entries 
+    // (even though the data was uploaded) - this occurred when a ultoken arrived but one other 
+    // final upload connection had not completed at the local end (even though it must have 
+    // completed at the server end).  So the file's data is still complete in the cloud.
+    // Here we check if the MAC is one of those with a missing entry (or a few if the connection had multiple chunks)
+
+    // last 3 connections, up to 32MB (ie chunks) each, up to two completing after the one that delivered the ultoken
+    size_t end = transfer->chunkmacs.size();
+    size_t finalN = std::min<int>(32 * 3, end);
```
sergiohs84, line 365: I guess that `3` comes from the fact that the maximum number of connections per transfer is 6, so there will be, at maximum, up to 3 gaps / missing entries. If that's the case, rather than `3`:
```suggestion
    size_t finalN = std::min<int>(32 * MegaClient::MAX_NUM_CONNECTIONS/2, end);
```

src/transferslot.cpp
```diff
@@ -347,7 +347,100 @@ int64_t TransferSlot::macsmac(chunkmac_map* m)
...
+                updateMacInKey(correctMac);
+                return true;
+            }
+        }
+    }
+
+    // now check for two separate pieces missing (much less likely)
+    // limit to checking up to 16Mb pieces wtih up to 8Mb between to avoid excessive CPU
+    // takes about 1 second on a fairly modest laptop for a 100Mb file (in a release build)
+    int caseschecked = 0;
```
sergiohs84, line 388: Unused variable :)
```suggestion
```

src/transferslot.cpp
```diff
@@ -347,7 +347,100 @@ int64_t TransferSlot::macsmac(chunkmac_map* m)
...
+    }
+    return false;
+}
+
+void TransferSlot::updateMacInKey(int64_t correctMac)
+{
+    // Update the Node's key to be correct (needs some API additions before enabling)
+    //for (file_list::iterator it = transfer->files.begin(); it != transfer->files.end(); it++)
+    //{
+    //    if ((*it)->hprivate && !(*it)->hforeign)
```
sergiohs84, line 419: Probably we want to allow sharees to rewrite keys too, not only the owner of the node.

sergiohs84 (dismissed):

src/transferslot.cpp
sergiohs84: We keep this part to a different PR ;)

sergiohs84 (approved):

PR #2356, opened 2020-11-11
SDK-1283. Crash in MegaClient::getRecentActions using Android app

jgandres:


jgandres:
trigger compilation

src/posix/fs.cpp
```diff
@@ -1343,15 +1343,13 @@ bool PosixFileSystemAccess::chdirlocal(LocalPath& name) const
 }
 
 // return lowercased ASCII file extension, including the . separator
-bool PosixFileSystemAccess::getextension(const LocalPath& filename, char* extension, size_t size) const
```
polmr, line 1346: same change should be carried out in Windows fs.cpp

include/mega/win32/megafs.h
```diff
@@ -78,7 +78,7 @@ class MEGA_API WinFileSystemAccess : public FileSystemAccess
     bool mkdirlocal(LocalPath&, bool) override;
     bool setmtimelocal(LocalPath&, m_time_t) override;
     bool chdirlocal(LocalPath&) const override;
-    bool getextension(const LocalPath&, char*, size_t) const override;
+    bool getextension(const LocalPath&, string) const override;
```
polmr, line 81:
```suggestion
    bool getextension(const LocalPath&, string&) const override;
```

polmr (changes requested):

polmr:
Now the issue for windows is in megacli:
```
megacli.cpp(6414): error C2660: 'mega::FileSystemAccess::getextension': function does not take 3 arguments 
include\mega\filesystem.h(506): note: see declaration of 'mega::FileSystemAccess::getextension'
```
almost there :sweat_smile:

PR #2364, opened 2020-11-12
macOS\Homebrew build adjustments

shinnok:
Hi,

I'm trying to build the sdk and megacli example program on macOS 10.15.7 and XCode clang++ version 11.0.3 and these are the changes I made to achieve it. I'm offering these changes in this PR hoping they are helpful and save time for the next one. Please review.


vtmateos:
Thanks for the collaboration @shinnok, it is really appreciated. We will check out.

shinnok:
Quick clarification, the lib freeimage fix might be obsolete with the current develop HEAD. Requires re-testing.

src/gfx/freeimage.cpp
```diff
@@ -252,7 +252,7 @@ bool GfxProcFreeImage::readbitmapFfmpeg(FileAccess* fa, const LocalPath& imagePa
 
     string extension;
     if (client->fsaccess->getextension(imagePath, extension)
-            && extension.compare(".mp3") && seek_target > 0
+            && strcmp(extension.c_str(),".mp3") && seek_target > 0
```
shinnok: I'm a bit confused about what git applied in this merge, or maybe this is a third version of achieving the same comparison, that is now in develop head? i.e. using c_str() with unsafe strcmp

shinnok:
@sergiohs84 any consensus on the HomeBrew situation?

sergiohs84:
> @sergiohs84 any consensus on the HomeBrew situation?

Yes, we should avoid to install software on our computers via `brew`. If you feel you need to do so, contact Shaun in advance for approval.

PR #2365, opened 2020-11-13
SDK-1291.  Make localseparator a const static public member of LocalPath.

jgandres:


src/file.cpp
```diff
@@ -484,12 +484,12 @@ void SyncFileGet::prepare()
                 transfer->localfilename = sync->localdebris;
                 sync->client->fsaccess->mkdirlocal(transfer->localfilename, true);
 
-                transfer->localfilename.appendWithSeparator(tmpname, true, sync->client->fsaccess->localseparator);
+                transfer->localfilename.appendWithSeparator(tmpname, true, LocalPath::localPathSeparator);
```
mattw-mega, line 487: Hi Javier, since this is constant now, we don't need to pass it at all - we can simplify these functions and the calls by removing the parameter, and just use the constant in the implementation.   thanks!

jgandres: Yeah, first I refactored automatically and then I were going to remove not necessary parameters in methods.

jgandres:
trigger compilation

mattw-mega (approved):

sergiohs84 (approved):

PR #2368, opened 2020-11-13
WIP: SDK-1258. Detection of external drives plugged in

catalinr-m:
For now this is only for a quick review. The first commit introduced 2 new files that are not included in the build system (is there a HOWTO for that?).

This implementation was tested with a simple POC app, on Windows, and worked as expected when connecting and disconnecting usb drives.

### DIFF
--- a/include/mega/volumewatcher.h
+++ b/include/mega/volumewatcher.h
@@ -1,0 +1,75 @@
+/**
+ * @file mega/volumewatcher.h
+ * @brief Mega SDK various utilities and helper classes
+ *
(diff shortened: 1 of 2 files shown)


catalinr-m:
Alternatively, this code could be added to `utils`.`h`|`cpp` files, maybe?
Should the dependency on `Qt` be specially handled? What if the build was done without `Qt`?

catalinr-m:
Replaced by [PR-2441](https://github.com/meganz/sdk/pull/2441).

PR #2378, opened 2020-11-19
SDK-1260 - Identify backups already configured

dwmega:
This branch adds support for storing configuration information about external backup syncs on external media.


aabellagm:
trigger compilation

PR #2379, opened 2020-11-19
SDK-1330. A fix for the occasional crashes reported, with the call stack indicating notifynode() [release v3.7.3g]

mattw-mega:
…ting notifynode().

Careful analysis of the exact condition required shows a Node's localnode pointer must be dangling.
Reused the crossref_ptr class to ensure that Node and LocalNode pointers to each other are consistent and never dangle.

### DIFF
--- a/src/megaclient.cpp
+++ b/src/megaclient.cpp
@@ -482,7 +482,7 @@ void MegaClient::setsid(const byte* newsid, unsigned len)
     size_t t = auth.size();
     auth.resize(t + len * 4 / 3 + 4);
     auth.resize(t + Base64::btoa(newsid, len, (char*)(auth.c_str() + t)));
-    
+
     sid.assign((const char*)newsid, len);
 }
 
@@ -580,7 +580,7 @@ SCSN::SCSN()
     clear();
 }
 
-void SCSN::clear() 
+void SCSN::clear()
 {
     memset(scsn, 0, sizeof(scsn));
     stopsc = false;
@@ -1199,7 +1199,7 @@ MegaClient::MegaClient(MegaApp* a, Waiter* w, HttpIO* h, FileSystemAccess* f, Db
     usehttps = true;
     orderdownloadedchunks = true;
 #endif
-    
+
     fetchingnodes = false;
     fetchnodestag = 0;
 
@@ -1273,7 +1273,7 @@ MegaClient::MegaClient(MegaApp* a, Waiter* w, HttpIO* h, FileSystemAccess* f, Db
         reqid[i] = static_cast<char>('a' + rng.genuint32(26));
     }
 
-    nextuh = 0;  
+    nextuh = 0;
     reqtag = 0;
 
     badhostcs = NULL;
@@ -2036,7 +2036,7 @@ void MegaClient::exec()
             case REQ_FAILURE:
                 if (pendingscUserAlerts->httpstatus == 200)
                 {
-                    error e = (error)atoi(pendingscUserAlerts->in.c_str());  
+                    error e = (error)atoi(pendingscUserAlerts->in.c_str());
                     if (e == API_EAGAIN || e == API_ERATELIMIT)
                     {
                         btsc.backoff();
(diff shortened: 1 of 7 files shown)


src/node.cpp
```diff
@@ -1597,7 +1588,7 @@ LocalNode::~LocalNode()
         // shutting down
         if (sync->state < SYNC_INITIALSCAN)
         {
-            node->localnode = NULL;
+            node.reset();
```
polmr, line 1591: This wouldn't be required as it'll happen upon crossref_ptr dtor, wouldn't it?

src/megaclient.cpp
```diff
@@ -12862,6 +12870,12 @@ bool MegaClient::syncdown(LocalNode* l, LocalPath& localpath, bool rubbish)
                 }
             }
         }
+
+        if (rit->second->localnode.get() == (LocalNode*)~0)
+        {
+            // make sure we are not leaving this marker pointer in the localnode
+            rit->second->localnode.store_unchecked(nullptr);
```
polmr, line 12877: I wonder if the lack of this could have cause the issue. Although looking into the code it appears it wouldn't happen for FILENODE which was the case of some reported case...
Anyway, perhaps we can have this assertion in `crossref_ptr`
```
    TO* operator->() const
   {
        assert (ptr != ~0);
        return ptr; 
}
```

src/node.cpp
```diff
@@ -1443,19 +1442,11 @@ treestate_t LocalNode::checkstate()
...
-
     deleted = false;
 
-    node = cnode;
-
-    if (node)
-    {
-        node->localnode = this;
-    }
+    if (cnode) cnode->localnode.reset();
```
polmr, line 1447: I'm not 200% sure if if's feasible, but we might want to check for (LocalNode*)~0:
```suggestion
    if (cnode) 
    {
        if (cnode->localnode.get() == (LocalNode*)~0)
        {
            cnode->localnode.store_unchecked(nullptr);
        }
        else
        {
            cnode->localnode.reset();
        }
    }
```

polmr:
```tests/unit/Serialization_test.cpp: In member function ‘virtual void Serialization_LocalNode_forFiles_oldLocalNodeWithoutSyncable_Test::TestBody()’:
tests/unit/Serialization_test.cpp:546:16: error: no match for ‘operator=’ (operand types are ‘mega::crossref_ptr<mega::Node, mega::LocalNode>’ and ‘mega::Node*’)
     l->node = &n;
                ^
In file included from ./include/mega.h:33:0,
                 from tests/unit/Serialization_test.cpp:26:
./include/mega/types.h:914:10: note: candidate: void mega::crossref_ptr<TO, FROM>::operator=(mega::crossref_ptr<TO, FROM>&&) [with TO = mega::Node; FROM = mega::LocalNode]
     void operator=(crossref_ptr&& p) { assert(!p.ptr); ptr = p; }
          ^~~~~~~~
./include/mega/types.h:914:10: note:   no known conversion for argument 1 from ‘mega::Node*’ to ‘mega::crossref_ptr<mega::Node, mega::LocalNode>&&’```

src/node.cpp
mattw-mega: Yes - though, for a hotfix I think it's good to keep the changes to a minimum and have a clear one-to-one mapping from old to new code.   Though, we could reduce the last 15 lines of this function to just 4.  Should we make the change for this hotfix though?

mattw-mega: And it's interesting that the other case did not clear those pointers - that's one possible case where the problem could have occurred.   Now if we haven't called reset() ourselves that other case will be covered.

src/megaclient.cpp
mattw-mega: Yes, I think it's a possibility.  The assert is a good idea, I'lll add that.

src/node.cpp
mattw-mega: We've prevented the ~0 leaving the function that uses that idea, so I don't think we need this change - if it does occur it's surely a bad memory access and we'll know from crash stacks.  And also, with the upcoming sync rework these pointers go away anyway.

dwmega (approved):

src/node.cpp
polmr: yep, that was what I was looking out when I added the comment :-), but MegaClient::movetosyncdebris did set the localnode to NULL.
We can keep the code as is. worst case scenario: we will wonder why in a few months when we forget about this PR ;-P

sergiohs84 (approved):

sergiohs84:
Replaced by https://github.com/meganz/sdk/pull/2454, to solve conflicts

PR #2382, opened 2020-11-20
SDK-1298. Add new flag to MegaTransferPrivate

sergiohs84:
The `startChatForUpload()` used the flag for streaming, which should be in use only for downloads. However, there's generic code that takes into account that flag for downloads too. In result, transfers started with the aforementioned method cannot be cancelled and are returned by `MegaApi::getStreamingTransfers()`.

This commits adds a new flag to the transfers, so it can be used for this type of uploads without interference with streaming transfers.

### DIFF
--- a/include/megaapi_impl.h
+++ b/include/megaapi_impl.h
@@ -705,7 +705,7 @@ class MegaTransferPrivate : public MegaTransfer, public Cacheable
         void setStartFirst(bool startFirst);
         void setBackupTransfer(bool backupTransfer);
         void setForeignOverquota(bool backupTransfer);
-        void setForceNewUpload(bool backupTransfer);
+        void setForceNewUpload(bool forceNewUpload);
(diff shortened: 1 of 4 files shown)


include/megaapi_impl.h
```diff
@@ -705,6 +705,7 @@ class MegaTransferPrivate : public MegaTransfer, public Cacheable
         void setStartFirst(bool startFirst);
         void setBackupTransfer(bool backupTransfer);
         void setForeignOverquota(bool backupTransfer);
+        void setForceNewUpload(bool backupTransfer);
```
jgandres, line 708:
```suggestion
        void setForceNewUpload(bool forceNewUpload);
```

jgandres (approved):

lhr000lhrmega:
https://github.com/meganz/iOS_dev/pull/971
It worked.

PR #2401, opened 2020-11-25
SDK-1301. Introduce UnifiedSync which brings syncs/configs/heartbeats together

mattw-mega:
Also include the heartbeat manager in SDK core, so it's avaliable in megacli and sync tests
Also avoid entangling the intermediate layer in SDK Core
Also unify syncs + syncconfigs so we don't have to rely on tag to look up one from the other.
Several sync-specific functions moved from MegaClient to syncs.h/cpp, and adapted.


src/megaclient.cpp
```diff
@@ -14816,163 +14489,43 @@ void MegaClient::failSync(Sync* sync, SyncError syncerror)
...
+        });
 
         Node *n = nodebyhandle(nodeHandle);
         nodeHandle = n ? n->parenthandle : UNDEF;
     }
-    return nullptr;
+    return syncFound;
 }
 
 void MegaClient::failSyncs(SyncError syncError)
```
polmr, line 14524: I wonder why aren't you moving this one to Syncs too?

src/heartbeats.cpp
```diff
@@ -446,329 +305,113 @@ MegaBackupMonitor::MegaBackupMonitor(MegaClient *client)
...
+void MegaBackupMonitor::digestPutResult(handle backupId, UnifiedSync* syncPtr)
+{
+    mClient->syncs.forEachUnifiedSync([&](UnifiedSync& us){
+        if (&us == syncPtr)
         {
-            LOG_warn << "discarding heartbeat update for pending sync: no valid id received for sync: " << pendingSyncPair->second->localFolder();
-
+            us.mConfig.setBackupId(backupId);
+            mClient->syncs.saveAndUpdateSyncConfig(&us.mConfig, us.mSync->state, us.mSync->errorCode);
         }
```
polmr, line 316: Why the loop? couldn't we use syncPtr directly?

src/megaclient.cpp
mattw-mega: I didn't need to change it, so I thought it would make review easier if it's left where it is for now, as the PR is already very large.

include/mega/megaclient.h
```diff
@@ -606,67 +605,41 @@ class MEGA_API MegaClient
...
     error isLocalPathSyncable(std::string newPath, int newSyncTag = 0, SyncError *syncError = nullptr);
 
+    /**
+     * @brief check config. Will fill syncError in the SyncConfig case there is one.
+     * Will fill syncWarning in the SyncConfig case there is one.
+     * Does not persist the sync configuration.
+     * @param syncError filled with SyncError with the sync error that prevented the addition
+     * @param delayInitialScan delay the initial scan
+     * @return API_OK if added to active syncs. (regular) error otherwise.
+     */
```
polmr, line 615: needs adjustment

include/mega/types.h
```diff
@@ -358,45 +356,18 @@ enum SyncError {
...
-    case LOCAL_IS_HGFS:
-        return false;
-    default:
-        return true;
-    }
-}
+enum SyncWarning {
+    NO_SYNC_WARNING = 0,
+    LOCAL_IS_FAT = 23,                      // Found FAT (not a failure per se)
+    LOCAL_IS_HGFS= 24,                      // Found HGFS (not a failure per se)
```
polmr, line 368: There's a _problem_ with separate warnings from errors in enumeration: if we ever start considering these an error (i.e. we cease to support that), we'd need to adjust apps accordingly. Not a big issue anyway ;-)

polmr (changes requested):

include/mega/megaclient.h
```diff
@@ -606,67 +605,41 @@ class MEGA_API MegaClient
...
+    error checkSyncConfig(SyncConfig& syncConfig, LocalPath& rootpath, std::unique_ptr<FileAccess>& openedLocalFolder, Node*& remotenode, bool& inshare, bool& isnetwork);
 
     /**
-     * @brief add sync. Will fill syncError in case there is one.
+     * @brief add sync. Will fill syncError/syncWarning in case there is one.
      * It will persist the sync configuration if everything goes fine.
      * @param syncError filled with SyncError with the sync error that prevented the addition
      * @param delayInitialScan delay the initial scan
      * @return API_OK if added to active syncs. (regular) error otherwise.
      */
```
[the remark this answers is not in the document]
mattw-mega: I've updated these two - though, for internal functions that we may edit quite frequently, this level of documentation seems like a lot - we can easily see what the parameters do by reading the function, for example.

src/heartbeats.cpp
```diff
@@ -452,330 +299,120 @@ MegaBackupMonitor::MegaBackupMonitor(MegaClient *client)
...
-        }
-        else
-        {
-            hbs->setPendingDowns(hbs->pendingDowns() + 1);
-        }
-        hbs->updateTransferInfo(transfer);
-    }
-}
-
-void MegaBackupMonitor::onTransferUpdate(MegaApi *api, MegaTransfer *transfer)
```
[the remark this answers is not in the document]
mattw-mega: We only send heartbeats every 30 seconds, so is there any value in keeping track of that?  It seems to me that just counting wholly completed transfers is sufficient.    If it really is needed then we can do something similar.

include/mega/megaclient.h
mattw-mega: updated

include/mega/types.h
mattw-mega: I did that mainly so we didn't have to call isError() all the time - presumably the apps would need to have a version of isError() too under the old system, and so they would need to be adjusted regardless.

src/megaclient.cpp
polmr: ok, it seems odd to have those there, but we can live with that ;-)

src/heartbeats.cpp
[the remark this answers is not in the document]
polmr: To me that is a must:
e.g: I have a 4GB file uploading  and my sync does not progress at all. WIth bigger files that would be even more flagrant.
It'll be a pity to lose that over having heartbeats support in megacli and others :relaxed: .
Would it be too much fuss? Feel free to open a discussion about it and hear others opinions ;-)

sergiohs84:
We should refactor some classes (right before merge, so discussions don't get out of context):

- `UnifiedSync` -> `SyncManager` (it was discussed that it's a better naming)
- `MegaBackupMonitor` -> `BackupMonitor` (it's not anymore a MegaApi object)
- `MegaBackupInfo` -> `BackupInfo` (it's not anymore a MegaApi object)

include/mega/megaclient.h
```diff
@@ -606,67 +605,45 @@ class MEGA_API MegaClient
...
     error isLocalPathSyncable(std::string newPath, int newSyncTag = 0, SyncError *syncError = nullptr);
 
+    /**
+     * @brief check config. Will fill syncError in the SyncConfig case there is one.
+     * Will fill syncWarning in the SyncConfig case there is one.
```
sergiohs84, line 610:
```suggestion cpp
     * @brief Will fill syncError and/or syncWarning in the SyncConfig, in case there is one.
```
(I miss the *in*)

include/mega/megaclient.h
```diff
@@ -606,67 +605,45 @@ class MEGA_API MegaClient
...
-     * It will persist the sync configuration if everything goes fine.
-     * @param syncError filled with SyncError with the sync error that prevented the addition
+     * @brief add sync. Will fill syncError/syncWarning in the SyncConfig in case there are any.
+     * It will persist the sync configuration if its call to checkSyncConfig succeeds
+     * @param syncConfig the Config to attempt to add
+     * @param debris Name of the debris folder on this platform (perhaps we could just use the macro)
+     * @param localdebris Alternate debris folder path - not used at all to my knowledge
      * @param delayInitialScan delay the initial scan
-     * @return API_OK if added to active syncs. (regular) error otherwise.
+     * @param unifiedSync If the syncConfig is added, this parameter wll be filled in with a pointer to the created UnifiedSync.
```
sergiohs84, line 626:
```suggestion
     * @param unifiedSync If the syncConfig is added, this parameter will be filled in with a pointer to the created UnifiedSync.
```

src/megaapi_impl.cpp
```diff
@@ -21503,82 +21400,49 @@ void MegaApiImpl::sendPendingRequests()
...
                     sync->setLocalFingerprint(fsfp);
                     request->setNumber(fsfp);
                 }
-                sync->setError(syncError);
                 sync->setMegaFolderYielding(remotePath.release());
                 request->setTransferTag(nextSyncTag);
-                syncMap[nextSyncTag] = sync;
 
-                fireOnSyncAdded(sync, e ? MegaSync::NEW_TEMP_DISABLED : MegaSync::NEW);
+                fireOnSyncAdded(sync.get(), e ? MegaSync::NEW_TEMP_DISABLED : MegaSync::NEW);
```
polmr, line 21431: Note, this is happening now after fireOnSyncStateChanged: 
```1  mega::MegaApiImpl::fireOnSyncStateChanged megaapi_impl.cpp 16638 0x55dee8db1715 
2  mega::MegaApiImpl::syncupdate_stateconfig megaapi_impl.cpp 13093 0x55dee8d9c7d3 
3  mega::UnifiedSync::changedConfigState     sync.cpp         2146  0x55dee8cc10c4 
4  mega::UnifiedSync::enableSync             sync.cpp         1996  0x55dee8cc035b 
5  mega::MegaClient::addsync                 megaclient.cpp   13110 0x55dee8c4f452 
...
```
which unfortunately breaks the regular flow the app shall expect :-(

polmr (changes requested):

src/megaclient.cpp
```diff
@@ -4611,7 +4536,7 @@ bool MegaClient::procsc()
                             //TODO: remove android control after android gives green light to this.
                             enabletransferresumption();
 #endif
-                            resumeResumableSyncs();
+                            syncs.resumeResumableSyncsOnStartup();
```
sergiohs84, line 4539: This call results in a segmentation fault upon login into a folder link. It attempts to access to `Syncs::mSyncConfigDb` and it's invalid. Folder links cannot be synced (no permission for uploads), so my suggestion is to protect the code with a `if (mClient->loggedinfolderlink()) return;` inside the method.

sergiohs84 (changes requested):
@mattw-mega I've pushed a couple of commits, one for fixing build without MegaApi, with qmake. I've just started the review, but found a segfault. Please, see my comment. I'll continue the review next week.

include/mega/sync.h
```diff
@@ -68,14 +75,45 @@ class MEGA_API SyncConfigBag
...
+    unique_ptr<Sync> mSync;
+
+    // High level info about this sync, sent to backup centre
+    std::unique_ptr<MegaBackupInfoSync> mBackupInfo;
+
+    // The next detail heartbeat to send to the backup centre
+    std::shared_ptr<HeartBeatSyncInfo> mNextHeartbeat;
+
+    // Reference to client
+    MegaClient& mClient;
```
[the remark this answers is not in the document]
sergiohs84: Fixed

src/heartbeats.cpp
```diff
@@ -452,330 +299,120 @@ MegaBackupMonitor::MegaBackupMonitor(MegaClient *client)
...
 
-void MegaBackupMonitor::updateBackupInfo(const MegaBackupInfo &info)
+void MegaBackupMonitor::updateBackupInfo(handle backupId, const MegaBackupInfo &info)
 {
     string localFolderEncrypted(mClient->cypherTLVTextWithMasterKey("lf", info.localFolder()) );
     string deviceIdHash = mClient->getDeviceidHash();
 
-    mClient->reqs.add(new CommandBackupPut(mClient, info.backupId(), info.type(), info.megaHandle(),
+    mClient->reqs.add(new CommandBackupPut(mClient,
+                                           info.type(),
```
sergiohs84, line 322: This is an update, so the client needs to provide the `backupId`. Otherwise, the API will assume that it's a new registration attempt.
```suggestion
    mClient->reqs.add(new CommandBackupPut(mClient,
                                           backupdId,
                                           info.type(),
```

src/heartbeats.cpp
```diff
@@ -452,330 +299,120 @@ MegaBackupMonitor::MegaBackupMonitor(MegaClient *client)
...
+                                           info.backupName(),
```
sergiohs84, line 323: The backup name is not updated this way, but in the corresponding user's attribute.
```suggestion
```

sergiohs84 (changes requested):

src/heartbeats.cpp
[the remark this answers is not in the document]
mattw-mega: Ok, it's straightforward enough, I've pushed the change here:  https://github.com/meganz/sdk/pull/2401/commits/8269ebf39344f1828df049655d8595c1f69c529f

src/megaapi_impl.cpp
mattw-mega: Hi, from the state of the code I had thought that is how it was supposed to work - so that the app can track the state of all syncs just from callbacks for one consistent method.  Or should it be that we don't notify of a change if the app asked for the change itself?   I started making code alterations for that second way (by all relevant functions taking a `bool notifyApp` parameter) but then thought, better to clarify first :-)  thanks

include/mega/megaclient.h
mattw-mega: updated, thanks

include/mega/megaclient.h
mattw-mega: upated, thanks

src/megaclient.cpp
mattw-mega: Thanks - updated with exactly that change

src/sync.cpp
```diff
@@ -1950,5 +1959,542 @@ bool Sync::movetolocaldebris(LocalPath& localpath)
...
+{
+    mSyncVec.clear();
+    resetSyncConfigDb();
+    isEmpty = true;
+}
+
+void Syncs::resetSyncConfigDb()
+{
+    mSyncConfigDb.reset();
+    if (mClient.dbaccess && !mClient.uid.empty())
```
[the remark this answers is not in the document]
mattw-mega: Yes I think that's a bit tidier - updated

src/commands.cpp
```diff
@@ -3177,7 +3148,7 @@ bool CommandGetUA::procresult(Result r)
                                     std::unique_ptr<std::string> container(tlvRecords->tlvRecordsToContainer(client->rng, &client->key));
                                     client->putua(at, (byte *)container->data(), unsigned(container->size()));
                                 }
-                                else
+                                else if (!client->mPendingBackupNames.empty())
```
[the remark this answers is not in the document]
mattw-mega: From memory, there was something going wrong there - I think I was seeing that LOG_warn output a lot, even when it shouldn't have been occurring.  But looking at the code now, it seems ok.   I'll remove that change and see if any problem recurs there.

src/commands.cpp
```diff
@@ -8355,6 +8328,10 @@ bool CommandBackupPut::procresult(Result r)
         }
     }
 
+    LOG_debug << "backup put result: " << error(e) << " " << backupId;
+
```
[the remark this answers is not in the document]
mattw-mega: That was probably just for convenient debugging - or to keep an eye on whether the API rejects our requests  -anyway, happy to remove it

src/heartbeats.cpp
mattw-mega: excellent point!  updated

src/heartbeats.cpp
mattw-mega: It looks like the command is saving that name (but not sending it) and in CommandBackupPut::procresult() it will do something complicated with it, that maybe adds it to client->mPendingBackupNames to be updated in the user attribute.  But, this scheme was not introduced by this PR.

src/heartbeats.cpp
```diff
@@ -452,330 +299,120 @@ MegaBackupMonitor::MegaBackupMonitor(MegaClient *client)
...
-        else // registration in-flight: backupId not received yet, let's queue the update
-        {
-            LOG_debug << " Queuing sync update, register is on progress for sync: " << info->localFolder();
-            mPendingSyncUpdates[syncTag].reset(info.release()); // we replace any previous pending updates
-        }
+        us.mBackupInfo = ::mega::make_unique<MegaBackupInfoSync>(us);
+        registerBackupInfo(currentInfo, &us);
     }
-    else //update
+    else if (us.mConfig.getBackupId() != UNDEF  &&
```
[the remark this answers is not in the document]
mattw-mega: Yes, I totally agree - we need the backupId changes.   I've left it like this on the assumption that initially we create a backup with id UNDEF, if it' still UNDEF then we try to register it, once that succeeds we set the Id and after that the normal heartbeats can be sent.     But all that can be addressed by the upcoming work to use backupId instead of `tag`.

src/megaclient.cpp
mattw-mega: Actually, how did any syncs get loaded if it's not the same account that saved them?   Surely the file/db should be encrypted to this account to prevent accidents like that.

sergiohs84: Well, I think the issue is that logging into a folder link, the `me` handle is the public handle of the folder link. The db is named like: `megaclient_statecache<version>_syncconfigs_<user_handle>.db`, so it attempts to find a file that doesn't exist, so `mSyncConfigDb` is invalid.
FYI, the records are encrypted to the user's Master Key ;)

src/heartbeats.cpp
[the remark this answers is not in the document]
sergiohs84: Well, there was some code in place from Pablo that aimed to keep those updates in a queue, so once the backup gets registered, the updates got sent out. However, with this branch, those changes are lost.
@polmr probably that's not needed anymore with the pre-registration, since the user won't be able to change the backup params until it's registered in the API, right? But something to take into account for the registration rework.

src/heartbeats.cpp
sergiohs84: Yes. The backup name is a param of the command to register a new backup, and once we have a backup id, we add an item to the TLV of the user's attribute storing backup names (`backupd_id:backup_name` records). However, since multiple new backups can be registered in a single HTTP request (with multiple `sp` commands in parallel), and whenever any of them finishes there could be a `upv` in flight, we need to enqueue the updates and apply them only with a fresh version of the attribute (to avoid dirty writes).

Anyway, the comment was only relative to the fact that this call is for an update of the backup via `sp`, and the name is not part of the ctor for updates (only for registrations). So it needs to be deleted, as you already did :smiley:

src/sync.cpp
```diff
@@ -1950,5 +1959,572 @@ bool Sync::movetolocaldebris(LocalPath& localpath)
 
     return false;
 }
+
+m_off_t Sync::getInflightProgress()
```
sergiohs84, line 1963: Not sure how the previous implementation was done, but this one requires to iterate through all transfer slots for every sync, every 30s. Taking into account the loop only takes into account active transfers, it shouldn't be time-consuming, right?

src/megaapi_impl.cpp
```diff
@@ -8876,16 +8860,16 @@ void MegaApiImpl::setExcludedRegularExpressions(MegaSync *sync, MegaRegExp *regE
...
-    }
-    MegaSyncPrivate* megaSync = syncMap.at(tag);
-    megaSync->setRegExp(regExp);
-    sdkMutex.unlock();
+    SdkMutexGuard g(sdkMutex);
+    client->syncs.forEachUnifiedSync([&](UnifiedSync& s) {
+        if (s.mConfig.getTag() == sync->getTag())
+        {
+            s.mConfig.setRegExps(regExpToVector(regExp));
+            client->syncs.saveSyncConfig(s.mConfig);
```
sergiohs84, line 8868: :raised_eyebrow:  Did we have a bug here? The update of the regular expression seems to not be saved in the DB previously. Good that they are only used globally, not per sync.

src/megaapi_impl.cpp
```diff
@@ -8930,33 +8914,27 @@ bool MegaApiImpl::isSyncable(const char *path, long long size)
...
+    client->syncs.forEachRunningSync([&](Sync* sync) {
+
         if (sync->localnodebypath(NULL, localpath, &parent) || parent)
         {
             if (sync->localdebris.isContainingPathOf(localpath))
             {
-                break;
+                auto temp = localpath.leafName();
+                auto name = temp.toName(*fsAccess, sync->mFilesystemType);
+                result = is_syncable(sync, name.c_str(), localpath);
```
sergiohs84, line 8934: Now that we don't break the loop, but iterate through all the syncs, I think we could fake the `result` with any of the other syncs.

Apart from that, not that `temp`, `name`, `result` used to get set when `!sync->localdebris.isContainingPathOf(localpath)` and now it's the other way :thinking:

sergiohs84 (changes requested):

src/heartbeats.cpp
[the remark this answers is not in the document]
polmr: looks fair enough for me

src/megaapi_impl.cpp
polmr: I'm not sure we are in the same page here:. 
Changes should always be notified and onSyncStateChange should be the one consistent method, but the app is expecting first to receive:
fireOnSyncAdded (to start tracking the sync), and then the corresponding state fireOnSyncStateChanged if changes occur.  With the revamping, the state changes happen before the first onSyncAdded arrives. Am I making sense?

src/heartbeats.cpp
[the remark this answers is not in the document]
polmr: Yep, we should remove the `Register` from this `updateOrRegisterSync` and this will always be updates and backupId will be always != UNDEF if there exists one unifiedSync. No need to control updates in between and all.

src/sync.cpp
mattw-mega: Yes - as it's once per 30 seconds it doesn't cost much - and we only ever have max 30 items (maybe more in the future but never an excessive amount) in the tslots list, so this is pretty trivial anyway.    The older version was based on a listener callback for every progress update of every transfer, all the time, and then looking that up in a map of active transfers, so probably considerably more costly.

src/megaapi_impl.cpp
mattw-mega: Ah I see, thanks for explaining the concern, that makes sense.   I've made a small adjustment (a bool notifyApp parameter to addsync() ) that addresses the issue, and prevents fireOnSyncStateChanged notifications before fireOnSyncAdded.  Apps like megacli using the SDK directly may prefer to be pass true to be notified immediately, but intermediate layer apps pass false. thanks

src/megaapi_impl.cpp
mattw-mega: that's a good point about it those being a global setting, I also was thinking it may have been a bug.  However, the SyncConfig serialization is storing all the regExp strings.   Does this mean we have a mismatch between design and implementation?   Or, are we transitioning to per-Sync settings for the exclusion rules?

src/megaapi_impl.cpp
mattw-mega: We don't allow nested syncs so we should only get one match maximum anyway - and even if that wasn't the case, the sync we matched first would have been nondeterministic -  but you're right, I didn't reverse the condition on the test for the debris folder that used to `break`, and should have.  I'll fix that.

src/megaapi_impl.cpp
sergiohs84: Actually, we do (were?) planning to provide exclusion rules per sync at some point, rather than globally. It's a popular user's demand. It's in line with the `.megaignore` (that we should finalize at some point) :smiley:

src/megaapi_impl.cpp
sergiohs84: Could we better add a new member to the `UnifiedSync::isInitialized` or alike, that helps the `fireOnSyncStateChanged()` to decide whether the notification should reach the app or not? I think it would be cleaner than the flag `notifyApp` in these 3 methods passing around.

src/megaapi_impl.cpp
mattw-mega: Ok, sounds good to me :-)

src/megaapi_impl.cpp
```diff
@@ -13085,43 +13063,51 @@ void MegaApiImpl::folderlinkinfo_result(error e, handle owner, handle /*ph*/, st
 }
 
 #ifdef ENABLE_SYNC
-void MegaApiImpl::syncupdate_state(int tag, syncstate_t newstate, SyncError syncerror, bool fireDisableEvent)
+
+MegaSyncPrivate* MegaApiImpl::cachedMegaSyncPrivateByTag(int tag)
 {
-    auto syncpair = syncMap.find(tag);
-    if( syncpair == syncMap.end())
+    if (tag == mCachedMegaSyncPrivateTag && mCachedMegaSyncPrivate)
```
sergiohs84, line 13069: Why not:
```suggestion cpp
    if (mCachedMegaSyncPrivate && tag == mCachedMegaSyncPrivate.getTag())
```
That way, we can directly get rid of the `mCachedMegaSyncPrivateTag`.
I'll fix it ;)

sergiohs84 (changes requested):

src/megaapi_impl.cpp
```diff
@@ -19918,6 +19810,9 @@ void MegaApiImpl::sendPendingRequests()
             {
                 client->locallogout(false);
                 client->restag = nextTag;
+#ifdef ENABLE_SYNC
+                mCachedMegaSyncPrivate.reset();
```
sergiohs84, line 19814: It is probably tidier to reset this member in the `logout_result()`, as the rest of members.
I'll move it there

src/megaapi_impl.cpp
sergiohs84: Nevermind, I've seen that Pablo already replaced the flag by a lambda completion function :)

src/megaclient.cpp
```diff
@@ -11889,14 +11812,7 @@ void MegaClient::fetchnodes(bool nocache)
...
 #ifdef ENABLE_SYNC
-        // lets remove the active syncs. There could be some when enforcing fetchnodes(true)
-        // after API_ETOOMANY (too many action packets)
-        for (sync_list::iterator it = syncs.begin(); it != syncs.end(); it++)
-        {
-            (*it)->changestate(SYNC_CANCELED);//we set it as cancelled. It will be removed from active sync latter on
-            // Note: this does not cause the sync to be removed from syncMap.
-            // However, after processing the new FetchNodes, they will be resumed, overriding the existing ones in the map.
-        }
+        assert(!syncs.hasRunningSyncs()); // not loaded yet - deferred to fetchnodes reply
```
sergiohs84, line 11815: This assert fails, @mattw-mega (try to `sync` a folder with megacli and run `reload`).
```
MEGAcli: ../../../src/megaclient.cpp:11815: void mega::MegaClient::fetchnodes(bool): Assertion `!syncs.hasRunningSyncs()' failed.
```
The comment you removed still applies:
>  // lets remove the active syncs. There could be some when enforcing fetchnodes(true)

Could we simply keep the existing behavior? or would it be problematic?

src/megaclient.cpp
```diff
@@ -12907,20 +12817,26 @@ error MegaClient::isnodesyncable(Node *remotenode, bool *isinshare, SyncError *s
...
         {
-            n = (*it)->localroot->node;
+            n = sync->localroot->node;
 
             do {
                 if (n == remotenode)
                 {
-                    if(syncError) *syncError = ACTIVE_SYNC_BELOW_PATH;
-                    return API_EEXIST;
+                    anyBelow = true;
```
sergiohs84, line 12830: We could break the loop and avoid going upwards in the tree, since we already found a sync below.
```suggestion
                    anyBelow = true;
                    break;
```

sergiohs84 (changes requested):

src/sync.cpp
```diff
@@ -1950,5 +1959,575 @@ bool Sync::movetolocaldebris(LocalPath& localpath)
...
+    }
+    return API_OK;
+}
+
+
+void UnifiedSync::changedConfigState(bool notifyApp)
+{
+    if (!mConfig.mEverKnown ||
+        (mConfig.mError != mConfig.mKnownError) ||
+        (mConfig.getEnabled() != mConfig.mKnownEnabled))
```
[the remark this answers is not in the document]
sergiohs84: Fixed

src/megaclient.cpp
```diff
@@ -2496,13 +2490,12 @@ void MegaClient::exec()
...
                             sync->initializing = false;
                             LOG_debug << "Initial delayed scan finished. New / modified files: " << sync->dirnotify->notifyq[DirNotify::DIREVENTS].size();
-                            saveAndUpdateSyncConfig(&sync->getConfig(), sync->state, NO_SYNC_ERROR);
                         }
                         else
                         {
                             LOG_err << "Initial delayed scan failed";
                             failSync(sync, INITIAL_SCAN_FAILED);
-                            sync->changestate(SYNC_FAILED, INITIAL_SCAN_FAILED); //note, this only causes fireOnSyncXXX if there's a MegaSync object in the map already
+                            sync->changestate(SYNC_FAILED, INITIAL_SCAN_FAILED, false, true); //note, this only causes fireOnSyncXXX if there's a MegaSync object in the map already
```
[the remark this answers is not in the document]
mattw-mega: yes, good point - removed three instances of it.

src/megaclient.cpp
```diff
@@ -3083,7 +3066,9 @@ void MegaClient::exec()
 
         if (btheartbeat.armed())
         {
-            app->heartbeat();
+#ifdef ENABLE_SYNC
+            syncs.mHeartBeatMonitor->beat();
```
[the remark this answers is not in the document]
mattw-mega: yes, this is the only usage of btheartbeat so we can safely do that - done.

src/megaclient.cpp
mattw-mega: This is a tricky one - really, reloading like this is not supported by the current sync algorithm, and should be impossible for MEGAsync.   Given that we will execute fetchnodes and get the last cached tree (not matching our current SCSN, most likely), and then receive actionpackets from there, some of which were already applied to the sync and some not - there's really no way to get this right (until the sync rework branch is ready).   I've opted to disable any active syncs at this point, with an explanatory new error code.

src/megaclient.cpp
mattw-mega: I've resolved this one with a new function already added (based on review comments) in Pablo's branch - forEachRunningSyncContainingNode

src/megaapi_impl.cpp
```diff
@@ -21503,82 +21399,42 @@ void MegaApiImpl::sendPendingRequests()
...
             SyncConfig syncConfig{nextSyncTag, localPath, name, request->getNodeHandle(), remotePath.get(),
                                   0, regExpToVector(request->getRegExp())};
 
-            e = client->addsync(syncConfig, DEBRISFOLDER, NULL, syncError, true, sync);
-            request->setNumDetails(syncError);
-            if (!e || !isSyncErrorPermanent(syncError))
+            UnifiedSync* unifiedSync = nullptr;
+            e = client->addsync(syncConfig, DEBRISFOLDER, NULL, true, unifiedSync, false);  // notifyApp = false since so we don't notify until after fireOnSyncAdded
+            request->setNumDetails(syncConfig.getError());
+            if (unifiedSync)
```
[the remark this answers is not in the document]
mattw-mega: Testing unifiedSync checks if the syncConfig was added to our set or not (and if it was then we should notify the app with fireOnSyncAdded() )  - !e isn't quite the same because some types of error may mean we didn't even add it, but other types of error mean we added it to the set of syncConfigs, but were not able to start it - or there may even be some errors (now or in future) where we can't differentiate those just on error code.   We could split those two cases into separate functions - effectively separating the checkSyncConfig(), and the remainder - though we already discussed and decided not to do that a while back, I'm happy to revisit that though.

src/megaclient.cpp
```diff
@@ -13024,203 +12946,152 @@ error MegaClient::isLocalPathSyncable(string newPath, int newSyncTag, SyncError
...
-                << syncConfig.getLocalPath()
-                << ": "
-                << LOG_NODEHANDLE(syncConfig.getRemoteNode());
+        LOG_warn << "Sync root does not exist in the cloud: "
+                 << syncConfig.getLocalPath()
+                 << ": "
+                 << LOG_NODEHANDLE(syncConfig.getRemoteNode());
 
-        syncError = REMOTE_NODE_NOT_FOUND;
+        syncConfig.mError = REMOTE_NODE_NOT_FOUND;
```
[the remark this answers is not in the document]
mattw-mega: This function checkSyncConfig() is called from two places - addsync, in which case the sync is not even added to our set of syncs yet (so can't be disabled), and also enablesync(), where it is in the set of syncs, but not running.  We do notify the app for the second case (if the error is different to one previously reported), but it wouldn't make sense for the first case.    This one also goes back to the possibility of separating out checkSyncConfig() for clearer code.

src/megaclient.cpp
```diff
@@ -14823,168 +14504,48 @@ void MegaClient::failSync(Sync* sync, SyncError syncerror)
...
+bool MegaClient::disableSyncContainingNode(mega::handle nodeHandle, SyncError syncError, bool newEnabledFlag)
 {
     auto sync = getSyncContainingNodeHandle(nodeHandle);
     if (sync)
     {
-        disableSync(sync, syncError);
-        return true;
+        syncs.disableSelectedSyncs([&](SyncConfig&, Sync* s) {
+            return s == sync;
+        }, syncError, newEnabledFlag);
```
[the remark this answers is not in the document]
mattw-mega: I've updated the return type to void, thanks

src/sync.cpp
```diff
@@ -1950,5 +1959,575 @@ bool Sync::movetolocaldebris(LocalPath& localpath)
...
+    {
+        if (s->mConfig.getTag() == tag)
+        {
+            syncPtrRef = s.get();
+            if (!s->mSync)
+            {
+                return s->enableSync(resetFingerprint, true);
+            }
+            s->mConfig.setError(ACTIVE_SYNC_BELOW_PATH);
+            s->mConfig.setEnabled(false);
```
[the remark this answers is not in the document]
mattw-mega: A good point - actually I'm not sure where these two lines came from, I think I must have copied them from somewhere erroneously.   The case here is that the app asked to enable a sync, but it's already running.  We shouldn't interfere with the already running sync or it's config, we can just return EEXIST.   updated, thanks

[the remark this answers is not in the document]
mattw-mega: Actually the tests were checking for ACTIVE_SYNC_BELOW_PATH in this case.  I've modified that to NO_ERROR.   I think originally that value was not stored in the config, but passed around in function parameters.  However, it's not clear from the megaapi.h documentation as to how the error should be interpreted - I would have thought any error passed back represented the state of the sync.    Another possibility would be to not consider this an error even for the return code (ie, not return API_EEXIST) - after all, the sync is enabled already so the calling app should be satisfied.

src/megaclient.cpp
sergiohs84: Yes, I think that disabling syncs is the right (conservative) approach in this case, though we may get an increase of user complains that get their syncs disabled upon `API_ETOOMANY` scenarios. Better safe than sorry

src/megaapi_impl.cpp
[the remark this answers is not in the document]
sergiohs84: Ouch! I didn't noticed about the reassignment of the `error` upon `enableSync()` here:
``` cpp
    if (!e)
    {
        // if we got this far, the syncConfig is kept (in db and in memory)
        unifiedSync = syncs.appendNewSync(config, *this);

        e = unifiedSync->enableSync(false, notifyApp);

        syncactivity = true;
        config = unifiedSync->mConfig;  // so the caller can easily check the config they passed in
    }
```
My bad, sorry. You're very correct. Let's keep it as it is right now. Thanks.

src/megaclient.cpp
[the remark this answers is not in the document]
sergiohs84: My suggestion was more in the direction of moving the `mEnabled = false` triggered from the usage of `checkSyncConfig()` at `UnifiedSync::enableSync()`. Also, take into account that for the rest of errors managed in `checkSyncConfig()`, the method already sets the config as disabled (I should have said "we should set the config as disabled", instead of "we should disable the sync").
I still think my suggestion is appropriate, without requiring to separate out `checkSyncConfig()`.

src/sync.cpp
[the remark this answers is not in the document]
sergiohs84: These cases are always tricky. I tend to prefer the error (`API_EEXIST`), since it provides extra information to the app about what's actually going on, which is positive for a nice UX. Imagine we silently return: all good. Then, the app could notify: "Your sync has been enabled" and do anything else that should ONLY happen when a sync has been enabled.
Instead, the app should say: "Your sync is already enabled" and nothing else. So good to go as it is now, by returning `API_EEXIST`, IMO.

src/posix/fs.cpp
```diff
@@ -1842,6 +1842,14 @@ DirNotify* PosixFileSystemAccess::newdirnotify(LocalPath& localpath, LocalPath&
     return dirnotify;
 }
 
+bool PosixFileSystemAccess::issyncsupported(const LocalPath& localpathArg, bool& isnetwork, SyncError& syncError, SyncWarning& syncWarning)
```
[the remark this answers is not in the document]
mattw-mega: Prior to this PR, the function took fewer parameters, and the base class had a default implementation in the header of just { return true; }, which is what Posix builds were using.   With the added parameters, I've defaulted that to =0 in the base class, and provided an implementation for posix (essentially still "return true" but sets the error parameters also).     The issyncsupported() function has been around since 2015 so I don't think we necessarily need to adjust per filesystem for Posix, unless we hear of some that are definitely not working?  Or perhaps if a ticket is created, it should be to investigate which filesystems do and don't work for Mac/Linux.  thanks

src/megaclient.cpp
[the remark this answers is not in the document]
mattw-mega: Oh, I see what you mean, thanks for the further explanation.  yes I think that's fine for both use cases - from addsync() it doesn't matter, the config object is discarded as it was never added, and for the enableSync case it keeps the setting of error and enable flag together.  I'll make the change.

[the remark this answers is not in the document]
mattw-mega: While making that update, I noticed addsync() always called enableSync(), even for configs that have mEnabled==false.   I've also added a check there that allows for adding syncs in a disabled state (ie, skip the enableSync() in that case)

mattw-mega:
trigger compilation

mattw-mega:
trigger compilation

mattw-mega:
trigger compilation

mattw-mega:
trigger compilation

src/posix/fs.cpp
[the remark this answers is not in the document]
sergiohs84: Given the expected workload for the coming months, let's simply keep it in mind when a user complains about sync issues on Mac/Linux. A network/vbox drive and/or a FAT/HGFS formatted one could be the reason.

sergiohs84 (approved):

sergiohs84 (approved):

PR #2402, opened 2020-11-25
SDK-1233. Avoid race condition upon set/remove backup for `upv`

sergiohs84:


ms-mega (approved):

include/mega/megaclient.h
```diff
@@ -511,6 +511,11 @@ class MEGA_API MegaClient
     // keep sync configuration after logout
     bool mKeepSyncsAfterLogout = false;
 
+    // backup names pending to be sent
+    string_map mPendingBackupNames;
+
+    // true if setting the backup name for any backup id is in progress
+    bool mSendingBackupName = false;
```
polmr, line 518: why not check emptyness of mPendingBackupNames?

src/commands.cpp
```diff
@@ -2867,6 +2867,28 @@ bool CommandPutUAVer::procresult(Result r)
...
             }
+            else if (at == ATTR_BACKUP_NAMES && client->mSendingBackupName)
+            {
+                if (client->mPendingBackupNames.empty())
+                {
+                    client->mSendingBackupName = false;
+                }
+                else    // more names arrived during `upv`
+                {
+                    string_map attrMap = move(client->mPendingBackupNames);
```
polmr, line 2878: why not use the mPendingBackupNames directly and have it emptied after User::mergeUserAttribute?

src/commands.cpp
```diff
@@ -8223,40 +8258,46 @@ bool CommandBackupPut::procresult(Result r)
...
         {
-            string_map attrMap { {key, Base64::btoa(mBackupName)} };
-            tlv.reset(TLVstore::containerToTLVrecords(oldValue, &client->key));
-            if (User::mergeUserAttribute(attrtype, attrMap, *tlv.get()))
+            // send backup name for this backup directly
+            std::unique_ptr<TLVstore> tlv { !oldValue
+                        ? new TLVstore()
+                        : TLVstore::containerToTLVrecords(oldValue, &client->key) };
+
+            string_map attrMap = move(client->mPendingBackupNames);
```
polmr, line 8292: mPendingBackupNames would be empty here, wouldn't it?

src/commands.cpp
```diff
@@ -8319,33 +8360,43 @@ bool CommandBackupRemove::procresult(Result r)
...
     client->app->backupremove_result(r.errorOrOK(), mBackupId);
 
-    if (r.succeeded())  // remove the corresponding backup name from the user's attribute
-    {
-        User *ownUser = client->finduser(client->me);
-        attr_t attrtype = ATTR_BACKUP_NAMES;
+
+    // Upon removal of backup successfully --> remove the backup name silently for the user's attribute
+    if (r.succeeded())
+    {
```
polmr, line 8366:
```suggestion
    {
```

src/commands.cpp
```diff
@@ -8319,33 +8360,43 @@ bool CommandBackupRemove::procresult(Result r)
...
-            string_map attrMap;
-            attrMap[key] = "";
+            // send backup name for this backup directly
+            std::unique_ptr<TLVstore> tlv { TLVstore::containerToTLVrecords(oldValue, &client->key) };
 
-            std::unique_ptr<TLVstore> tlv;
-            tlv.reset(TLVstore::containerToTLVrecords(oldValue, &client->key));
-
-            if (User::mergeUserAttribute(attrtype, attrMap, *tlv.get()))
+            string_map attrMap = move(client->mPendingBackupNames);
```
polmr, line 8391: same as before. I also wonder if it isn't worth it to encapsulate all the code that gets the old attribute, create the tlv and merges the pending changes (addition/removal), and calls putua with the new data.

src/megaclient.cpp
```diff
@@ -6024,19 +6024,24 @@ void MegaClient::sc_userattr()
...
                                         break;
                                     }
-                                    case ATTR_AUTHRING:     // fall-through
-                                    case ATTR_AUTHCU255:    // fall-through
-                                    case ATTR_AUTHRSA:
+                                    case ATTR_AUTHRING:         // fall-through
+                                    case ATTR_AUTHCU255:        // fall-through
+                                    case ATTR_AUTHRSA:          // fall-through
+                                    case ATTR_MY_BACKUPS_FOLDER:    // fall-through
+#ifdef ENABLE_SYNC
```
polmr, line 6039: this shouldn't be restricted to SYNCS, shouldn't it?

polmr (changes requested):

src/megaclient.cpp
sergiohs84: Agree. Fixed

src/commands.cpp
sergiohs84: Not necessarily. First time, yes, but later it could happen that we just finished sending pending backup names and we got some new pending backup names.

src/commands.cpp
sergiohs84: Agree. Fixed.

include/mega/megaclient.h
sergiohs84: It cannot be done, since we may have pending backup names while we are sending others, and need to decide whether to send the new backup names directly, or not, depending on the current state: sending or not, rather than empty map (which may or may not accumulate new names).

src/megaclient.cpp
sergiohs84: Yep, done.

src/commands.cpp
sergiohs84: I think it doesn't worth. Each case is slightly different.

PR #2404, opened 2020-11-26
fix memory leaks in recentActions and CameraUploads

simonwng:
In iOS 5.4 build, we have detected big memory leaks in recent actions and some obvious small leaks in camera uploads. This PR is to fix the leaks.


jnavarrom (approved):

sergiohs84 (approved):

sergiohs84:
trigger compilation

PR #2405, opened 2020-11-26
fix memory leaks in recentActions and CameraUploads

simonwng:
In iOS 5.4 build, we have detected big memory leaks in recent actions and some obvious small leaks in camera uploads. This PR is to fix the leaks.

### DIFF
--- a/bindings/ios/MEGASdk.mm
+++ b/bindings/ios/MEGASdk.mm
@@ -2136,6 +2136,7 @@ using namespace mega;
     
     return [MEGANodeList.alloc initWithNodeList:self.megaApi->searchByType(node ? [node getCPtr] : NULL, searchString.UTF8String, cancelToken ? [cancelToken getCPtr] : NULL, recursive, (int)orderType, (int)nodeFormatType, (int)folderTargetType) cMemoryOwn:YES];
 }
+
 - (NSMutableArray *)recentActions {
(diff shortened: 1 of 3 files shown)


jnavarrom (approved):

bindings/ios/MEGASdk.mm
```diff
@@ -2145,6 +2146,8 @@ - (NSMutableArray *)recentActions {
         [recentActionBucketMutableArray addObject:recentActionBucket];
     }
     
+    delete megaRecentActionBucketList;
```
joel-meng-nz, line 2149: wow, if have to manually deallocate the objects like this, there are more, at least the banner API binding, need to be updated as well.

jnavarrom: It will depend, you need to do it if the SDK saids: "* You take the ownership of the returned value."

sergiohs84: If the SDK documentation states that you take the ownership of the returned object, @joel-meng-nz , you do need to deallocate them, yes.

Is the SmartBanner feature included in 5.4? If not, please, fix your bindings in a different branch/PR. Thanks

sergiohs84: Navarro already told me it does not. Merging this PR...

PR #2409, opened 2020-12-01
SDK-1312. Interface to get the account ctime

sergiohs84:


src/megaapi_impl.cpp
```diff
@@ -5394,6 +5394,11 @@ char* MegaApiImpl::getMyEmail()
     return result;
 }
 
+int64_t MegaApiImpl::getAccountCreationTs()
+{
+    return client->accountsince;
```
jgandres, line 5399: Although is only getting access to **accountsince** for read, wouldn't it be appropriate to lock sdkMutex, to avoid possible issues in the future

jgandres: It doesn't have sense for the moment, as most of the CPU's read int vars atomically, so it doesn't represent a risk.

jgandres (approved):

src/megaapi_impl.cpp
sergiohs84: `accountsince` is an integer, which is read atomically by most CPUs. Also, the addition of the mutex here would imply that the app may get locked if the SDK is busy, just for reading the value. On top of that, there is a bunch of other public interfaces that behave the same (assume read is atomic), so I think we should keep it without requiring a lock on the `sdkMutext`.
Nevertheless, thanks for the warning ;)

mattw-mega: I think many machines nowadays would still read and write a 64 bit value in RAM/cache as two separate 32 bit read/writes, so you could theoretically read a value that was only half updated.   Now that we can use C++11, you can easily solve this for all cases by making it an atomic<int64_t>.   thanks

PR #2413, opened 2020-12-03
Add missing identifier field for MegaBanner

joel-meng-nz:
Add the missing field for MEGABanner in iOS binding.

### DIFF
--- a/bindings/ios/MEGABanner.mm
+++ b/bindings/ios/MEGABanner.mm
@@ -58,7 +58,7 @@ using namespace mega;
 }
 
 - (NSUInteger)identifier {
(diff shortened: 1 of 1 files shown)


bindings/ios/MEGABanner.mm
```diff
@@ -57,7 +57,11 @@ - (MegaBanner *)getCPtr {
     return self.megaBanner;
 }
 
-- (NSString *)title{
+- (NSUInteger)identifier {
+    return self.megaBanner->getId() ? self.megaBanner->getId() : 0;
```
simonwng, line 61: we can do self.megaBanner->getId() directly, or cast it to `NSUInteger `

simonwng (approved):

PR #2414, opened 2020-12-03
#IOS-4387. iOS Binding - Implemented the equality method for MEGANode.

lancy-mega:
Implemented the equality method for MEGANode.

### DIFF
--- a/bindings/ios/MEGANode.mm
+++ b/bindings/ios/MEGANode.mm
@@ -53,6 +53,14 @@ using namespace mega;
     return self.megaNode ? [[MEGANode alloc] initWithMegaNode:self.megaNode->copy() cMemoryOwn:YES] : nil;
 }
 
+- (BOOL)isEqual:(id)object {
+    if (![object isKindOfClass:[MEGANode class]]) {
+        return false;
+    }
+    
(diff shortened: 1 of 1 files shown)


bindings/ios/MEGANode.mm
```diff
@@ -53,6 +53,14 @@ - (instancetype)clone {
     return self.megaNode ? [[MEGANode alloc] initWithMegaNode:self.megaNode->copy() cMemoryOwn:YES] : nil;
 }
 
+- (BOOL)isEqual:(id)object {
```
simonwng, line 56: We must override `hash` method to make sure the hash values are same for 2 equal objects. Otherwise, a hash based collection will malfunction.

simonwng: Please checkout this link, it could be helpful to define your hash code.
https://nshipster.com/equality/#hashing

simonwng (approved):

bindings/ios/MEGANode.mm
lancy-mega: Right, Thanks. I have added the hash method.

jnavarrom (approved):

PR #2433, opened 2020-12-09
SDK-1321. Task/build 3rdparty script improvements

mattw-mega:


contrib/cmake/build3rdParty/build3rdParty.cpp
```diff
@@ -0,0 +1,279 @@
...
+    std::vector<char*> myargv1(argv + 1, argv + argc);
+    std::vector<char*> myargv2;
+
+    for (auto it = myargv1.begin(); it != myargv1.end(); ++it)
+    {
+        if (std::string(*it) == "--ports")
+        {
+            if (++it == myargv1.end()) return showSyntax();
+            portsFile = *it;
+            argc -= 2;
```
polmr, line 192: no need for this

contrib/cmake/preferred-ports-megacmd.txt
```diff
@@ -0,0 +1,54 @@
+
```
polmr, line 1: Would it make sense for these to go into MEGAcmd/MEGAsync/... projects instead of here ... :thinking:

contrib/cmake/build3rdParty/build3rdParty.cpp
mattw-mega: thanks, removed

contrib/cmake/preferred-ports-megacmd.txt
mattw-mega: yes, but for ease of development and review I've left them here for now - if and when this new way to build 3rdParty is approved, let's move them to the appropriate repos.  thanks

contrib/cmake/build3rdParty/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+
+
+cmake_minimum_required(VERSION 3.13)
```
mega-lia, line 3: There is no need to re-specify `cmake_minimum_required` and `project` in individual CMLs; it should be done once in the topmost CML

contrib/cmake/build3rdParty/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+endif()
+
+if (APPLE)
+    set(CMAKE_CXX_STANDARD 11)
+    SET(CMAKE_CXX_STANDARD_REQUIRED ON)
+endif()
+
+add_executable(build3rdParty build3rdParty.cpp)
+if (NOT WIN32 AND NOT APPLE)
+    target_link_libraries(build3rdParty stdc++fs)
```
mega-lia, line 18: Do we support building with libc++? This becomes a bit more complicated if so

contrib/cmake/build3rdParty/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+
+
+cmake_minimum_required(VERSION 3.13)
+project(Build3rdParty)
+
+if (WIN32)
+    # accurate __cplusplus macro for vc++, selecting c++17 here for windows builds though the MEGA SDK library must build for older c++ standards also
```
mega-lia, line 7: We should consider using the `target_meow` approach here, e.g.,

```cmake
target_compile_features(build3rdParty cxx_std_11 REQUIRED)
target_compile_features(build3rdParty cxx_std_17 REQUIRED)
target_compile_options(build3dParty /Zc:__cplusplus)
```

Also I haven't developed on MSVC in a bit over a year--is support for `/Zc:__cplusplus` more widespread now? There used to be some libraries that detected if they were running on MSVC by checking for `__cplusplus` being 1997

contrib/cmake/build3rdParty/build3rdParty.cpp
```diff
@@ -0,0 +1,273 @@
...
+#include <filesystem>
+#include <vector>
+#include <map>
+#include <string>
+#include <fstream>
+
+using namespace std;
+
+
+#if (__cplusplus >= 201703L && (!defined(__GNUC__) || (__GNUC__*100+__GNUC_MINOR__) >= 800))
```
mega-lia, line 11: If we're checking for `__cplusplus >= 2017` that means we should have access to `__has_include`, might be worth simplifying some of the logic that way? 

```cpp
#if (__cplusplus >= 201703L)
#if __has_include(<filesystem)>
#include <filesystem>
namespace fs = std::filesystem
#else // __has_include(<filesystem)>
#include <experimental/filesystem>
namespace fs = std::experimental::filesystem
#else // (__cplusplus >= 201703L)
// ...
#endif 
```

contrib/cmake/build3rdParty/build3rdParty.cpp
```diff
@@ -0,0 +1,273 @@
...
+                #else
+                    execute("./bootstrap-vcpkg.sh -disableMetrics");
+                #endif
+            }
+            else
+            {
+                fs::current_path("vcpkg");
+            }
+
+            if (fs::exists(sdkRootPath / "contrib" / "cmake" / "vcpkg_extra_triplets" / (triplet + ".cmake")))
```
mega-lia, line 72: Minor but `std::filesystem::path::operator/` is implemented as a string concatenation and I'm pretty sure `'/'` is acceptable as a separator on any platform so we may want to just hardcode `"contrib/cmake/vcpkg_extra_triplets"` here and elsewhere

contrib/cmake/build3rdParty/build3rdParty.cpp
```diff
@@ -0,0 +1,273 @@
...
+                cout << "Copying triplet from SDK: " << triplet << endl;
+                fs::copy(sdkRootPath / "contrib" / "cmake" / "vcpkg_extra_triplets" / (triplet+".cmake"), vcpkgDir / "triplets" / (triplet + ".cmake"));
+            }
+            else if (!fs::exists(vcpkgDir / "triplets" / (triplet + ".cmake")))
+            {
+                cout << "triplet not found in the SDK or in vcpkg: " << triplet << endl;
+                exit(1);
+            }
+
+            for (auto portPair : ports)
```
mega-lia, line 87: Let's do `const auto&` to avoid two string copies here?

contrib/cmake/build3rdParty/build3rdParty.cpp
```diff
@@ -0,0 +1,273 @@
...
+            {
+                const string& portname = portPair.first;
+                const string& portversion = portPair.second;
+
+                if (fs::is_directory(fs::path("ports") / portname))
+                {
+                    cout << "Removing " << (vcpkgDir / "ports" / portname).u8string() << endl;
+                    fs::remove_all(vcpkgDir / "ports" / portname);
+                }
+                if (portversion.size() == 40)
```
mega-lia, line 97: 40  strikes me as a magic number, maybe could define it as a `constexpr` var or add a comment to explain that this branch (as I understand it) means "looks like a commit SHA?"

contrib/cmake/build3rdParty/build3rdParty.cpp
```diff
@@ -0,0 +1,273 @@
...
+                else
+                {
+                    cout << "Copying port for " << portname << " from SDK customized port " << portversion << endl;
+                    fs::copy(sdkRootPath / "contrib" / "cmake" / "vcpkg_extra_ports" / portname / portversion, vcpkgDir / "ports" / portname, fs::copy_options::recursive);
+                }
+            }
+
+            if (removeUnusedPorts)
+            {
+                for (auto dir = fs::directory_iterator(vcpkgDir / "ports"); dir != fs::directory_iterator(); ++dir)
```
mega-lia, line 114: I believe this can be done with range-based `for`: `for (const auto& dir : fs::directory_iterator(vcpkgDir / "ports"))`

contrib/cmake/build3rdParty/build3rdParty.cpp
```diff
@@ -0,0 +1,273 @@
...
+                cout << "This command should be run from just outside 'vcpkg' folder - maybe it is not set up?" << endl;
+                return 1;
+            }
+            else
+            {
+                fs::current_path("vcpkg");
+            }
+
+
+            for (auto portPair : ports)
```
mega-lia, line 138: `const auto&` here as well

contrib/cmake/build3rdParty/build3rdParty.cpp
```diff
@@ -0,0 +1,273 @@
...
+                #endif
+            }
+        }
+
+
+    }
+
+    return 0;
+}
+catch (exception& e)
```
mega-lia, line 153: Might as well catch by `const exception&` since we're not mutating it?

contrib/cmake/build3rdParty/build3rdParty.cpp
```diff
@@ -0,0 +1,273 @@
...
+{
+    cout << "exception: " << e.what() << endl;
+    return 1;
+}
+
+
+void execute(string command)
```
mega-lia, line 160: This can be `const string&`

contrib/cmake/build3rdParty/build3rdParty.cpp
```diff
@@ -0,0 +1,273 @@
...
+bool readCommandLine(int argc, char* argv[])
+{
+    if (argc <= 1) return showSyntax();
+
+    std::vector<char*> myargv1(argv + 1, argv + argc);
+    std::vector<char*> myargv2;
+
+    for (auto it = myargv1.begin(); it != myargv1.end(); ++it)
+    {
+        if (std::string(*it) == "--ports")
```
mega-lia, line 188: We might want to do `const std::string currentArg(*it)` or similar; in the case we hit the final `else ` at the bottom, we've constructed the same temporary string six times. 

Also untested, maybe overkill, but maybe more maintainable:

```cpp
using namespace std::string_literals;

map<string, reference_wrapper<bool> flags{
  {"--setup"s, ref(setup)},
  {"--removeUnusedPorts"s, ref(removeUnusedPorts)},
  {"--build"s, ref(build)}
};

map<string, reference_wrapper<string>> options{
 {"--ports"s, ref(portsFile)},
 // ...
};

for (auto it = myargv1.begin(); it != myargv1.end(); ++it) {
  const string currentArg(*it);
  auto flagIt = flags.end();
  auto optIt = options.end();

  if ((flagIt = flags.find(currentArg)) != flags.end()) {
    if (++it == myargv1.end()) return showSyntax();
    flagIt->second = *it;
  } else if ((optIt = options.find(currentArg)) != options.end()) {
    // same check
    optIt->second = true;
  } else {
    myargv2.push_back(*it);
  }
}
```

contrib/cmake/build3rdParty/build3rdParty.cpp
```diff
@@ -0,0 +1,273 @@
...
+    cout << "build3rdParty --build --ports <ports override file> --triplet <triplet>" << endl;
+    return false;
+}
+
+bool readCommandLine(int argc, char* argv[])
+{
+    if (argc <= 1) return showSyntax();
+
+    std::vector<char*> myargv1(argv + 1, argv + argc);
+    std::vector<char*> myargv2;
```
mega-lia, line 184: Based on what happens below, looks like `unrecognizedArgs` might be a better name for `myargv2`

CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+# This file is here in the root folder so it can be opened as a CMake project in Visual Studio, eg in order to debug a Linux build in WSL2.
+cmake_minimum_required(VERSION 3.13)
+project(TryWSL)
```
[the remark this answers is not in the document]
mattw-mega: Using cmake was a bit of an experiment I started when I joined, so the main CMakeLists.txt started of in that contrib/cmake folder.   If we get agreement from the team, perhaps we could move that here, but in the meantime this is sort of a pointer to that folder.    That particular 'project' line was needed to prevent cmake complaining - I think the later 'project' statement in the real CMakeLists.txt overwrites it anyway.   And that one can be user-specified through cache variables, I use that so that when I have multiple Visual Studios open, the project name reminds me what I'm working on in each one.

CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
...
+include(${CMAKE_CURRENT_LIST_DIR}/contrib/cmake/CMakeLists.txt)
```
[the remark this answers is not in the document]
mattw-mega: Indeed, this one is not following that model - it's just referencing the main CMakeLists.txt file.   That one isn't using the cmake-per-subdirectory method either, I found it much easier for our relatively small repo to have the entire script in a single file, it's much easier to see how it works, especially while learning cmake.

contrib/cmake/build3rdParty/CMakeLists.txt
mattw-mega: Since some people will still be using this one directly, and some perhaps using the one in the root, I think they are both needed or cmake will complain.

contrib/cmake/build3rdParty/CMakeLists.txt
mattw-mega: We should support them all, I think, but probably we just haven't run into any issue with this one yet.   Actually we can't really use std::filesystem for production stuff since we are on C++11 for now.   However it is very convenient and we use it in some test apps etc, referencing the early not-quite-standard experimental versions that were available with the C++11 compilers.

contrib/cmake/build3rdParty/CMakeLists.txt
mattw-mega: I think this is the sort of thing we would adjust if any problem emerged - for windows we're using VS 2017/2019 and both seem fine.

contrib/cmake/build3rdParty/build3rdParty.cpp
mattw-mega: This particular block of #ifs etc is our current method of figuring out which compilers have some version of std::filesystem or std::experimental::filesystem that we can use - it's been through many, many revisions and it's quite difficult to get right across all the platforms, I'd be wary of editing it unless there's a case it doesn't cover.

contrib/cmake/build3rdParty/build3rdParty.cpp
mattw-mega: From my understanding of the filesystem library, the separated out  / is the platform-independent way to code - some implementations may work that way, but we can't say they all do.   I just double checked at https://en.cppreference.com/w/cpp/filesystem, ```The native format (e.g. which characters are used as separators) and character encoding of the string representation of a path (the pathname) is implementation-defined, this library provides portable representation of paths.```

contrib/cmake/build3rdParty/build3rdParty.cpp
mattw-mega: No objection; this is a super small program where efficiency is not important though.

contrib/cmake/build3rdParty/build3rdParty.cpp
mattw-mega: yes exactly - a quick and dirty check for a commit SHA.   Feel free to improve it, thanks

contrib/cmake/build3rdParty/build3rdParty.cpp
mattw-mega: That would be cool, I never tried it as it seemed unlikely since it doesn't have begin()/end() - if it works, does it require C++17?

contrib/cmake/build3rdParty/build3rdParty.cpp
mattw-mega: same

contrib/cmake/build3rdParty/build3rdParty.cpp
mega-lia: Yeah I think it works without a member begin/end! They provide [non-member overloads](https://en.cppreference.com/w/cpp/filesystem/directory_iterator/begin) for range-for compatibility

contrib/cmake/build3rdParty/build3rdParty.cpp
mattw-mega: Personally I find the extra syntax for 'const' unnecessary - you can see at a glance that it's not changed in such a small piece of code.   And if we did change it, well, there's no other piece of code that we would be breaking by doing that.   If it's a longer function where one might legitimately worry about something being modified when it shouldn't be then of course const is useful in the declaration, but I don't think we need to bulk up the code by adding it in every place that it could fit.

contrib/cmake/build3rdParty/build3rdParty.cpp
mattw-mega: same

contrib/cmake/build3rdParty/build3rdParty.cpp
mattw-mega: For this small program where performance is unimportant, I'd lean towards keeping the code small rather than trying to be efficient.

contrib/cmake/build3rdParty/build3rdParty.cpp
mattw-mega: no objection :-)

contrib/cmake/build3rdParty/build3rdParty.cpp
mattw-mega: That's great, thanks for the info.   Hopefully that is part of the pre-17 experimental implementations too - this is a good project to try that and find out.  thanks

contrib/cmake/CMakeLists.txt
```diff
@@ -448,15 +468,7 @@ if (USE_THIRDPARTY_FROM_VCPKG)
...
-                check_cxx_compiler_flag(-fopenmp HAS_FOPENMP)
-
-                set(Mega_PlatformSpecificLibs
-                      ${Mega_PlatformSpecificLibs}
-                      $<$<BOOL:${HAS_FGOMP}>:-fgomp>
-                      $<$<BOOL:${HAS_FOPENMP}>:-fopenmp>)
+                set(Mega_PlatformSpecificLibs ${Mega_PlatformSpecificLibs} -fopenmp -fgomp)
             ENDIF()
 
         ENDIF(USE_LIBRAW)
```
mattw-mega, line 474: Is this change intentional?  Since you were working specifically on Mac, and this would be a change for Linux  - and it seems like checking the compiler supports those flags would be a good idea.  thanks

tests/integration/SdkTest_test.cpp
```diff
@@ -4839,7 +4839,7 @@ TEST_F(SdkTest, SdkBackupFolder)
     }
     else
     {
-        auto now = m_time(nullptr);
+        const time_t now = m_time(nullptr);
         char timebuf[32];
         strftime(timebuf, sizeof timebuf, "%c", localtime(&now));
```
mattw-mega, line 4845: It turns out time_t is a bit inconsistent across platforms and 32/64 bit compilers.  We have smoothed that out with m_time and some other related functions all beginning with m_, which operate on m_time_t instead.  I think the right solution here is to call m_localtime instead?    Also, do we need that 'const' ?  In C++ the default it that variables are variable - I think we should go with the C++ flow, rather than fight the language, and only worry about const where we want to remind other devs that something would break if they decided to actually modify it.  thanks

contrib/cmake/vcpkg_extra_ports/ffmpeg/4.2-3/0001-create-lib-libraries.patch
```diff
@@ -1,25 +0,0 @@
-diff --git a/configure b/configure
-index 34c2adb..1c8008a 100755
---- a/configure
-+++ b/configure
```
mattw-mega, line 4: github is showing this file and a lot of other vcpkg_extra_port files being removed now, if we merge to develop - but we don't want to break any existing usage of those files.   Can we arrange for no pending changes to those files except where necessary for this PR?   thanks

contrib/cmake/CMakeLists.txt
mega-lia: hm don't remember changing these, wonder if they snuck in during a dev pull

tests/integration/SdkTest_test.cpp
mega-lia: I looked at the header with `m_localtime` defined (and other uses in the SDK) and I think I've replaced this to use it correctly, let me know.

The issue of `const` is now sidestepped entirely (if I've used the functions correctly, was following [this](https://github.com/meganz/sdk/blob/13df5bc543ed21b1edd15010593790ae095c1bed/src/sync.cpp#L2005))but my personal conventions (and justifications) lie here: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#con1-by-default-make-objects-immutable 
I think a lot of C++ defaults are inherited for C compatibility or arguably dated conventions, and we should evaluate them for safety/clarity/performance wherever possible--as a developer glancing at someone else's code it's much easier to see `const` and know the variable cannot change (and that I will get a compiler error if I try to do so) rather than see a mutable variable and have to mentally keep in mind that it might change somewhere

contrib/cmake/vcpkg_extra_ports/ffmpeg/4.2-3/0001-create-lib-libraries.patch
mega-lia: Restored, let me know if I missed any

tests/integration/SdkTest_test.cpp
mattw-mega: thanks for the fix, and the explanation - I've sent you a counterargument for that one - the article is advice for beginners really, and doesn't take into account whether you're in a complicated situation where it matters, or a simple one where it's just redundant and distracting extra syntax.

mega-lia (approved):

PR #2441, opened 2020-12-15
WIP: SDK-1258. Detection of external drives plugged in (Win impl, posix stubs)

catalinr-m:


contrib/cmake/CMakeLists.txt
```diff
@@ -469,14 +469,36 @@ endif(USE_THIRDPARTY_FROM_VCPKG)
...
+unset(HAVE_DRIVE_NOTIFY CACHE)
+
 if(WIN32)
+
+    # Test for WMI / WBEM / DriveNotify support
+    include(CheckCXXSourceCompiles)
+    set (CMAKE_REQUIRED_LIBRARIES "wbemuuid")
+    check_cxx_source_compiles("
+    #include <wbemcli.h>
+    int main()
```
mattw-mega, line 482: Hi, for the building of release software, we really want to know that all expected features are present.  I suggest using the normal USE_DRIVE_NOTIFY cached variable, which the person configuring the build would set.   If they choose to turn that on, the build should fail if it's not present.  And if they choose not to use it, it doesn't matter if the headers etc are present or not.   We wouldn't want to accidentally release software that we thought contained this feature but the compile system had silently skipped.  thanks!

examples/megacli.cpp
```diff
@@ -6652,6 +6656,134 @@ void exec_enabletransferresumption(autocomplete::ACState& s)
...
+
+    // notifydrive [connect|disconnect] N
+    else if (s.words.size() == 3)
+    {
+        int t = 0;
+        try
+        {
+            t = stoi(s.words[2].s);
+        }
+        catch (...) {}
```
mattw-mega, line 6704: How about the much shorter  `int t = atoi(...);` if we are going to ignore any exception anyway ;-)

mattw-mega: Or, much better - let the exception be thrown, and have a single catch(std::exception& e) in the main loop to report e.what()  for any command that throws such an exception?  thanks

include/mega/drivenotify.h
```diff
@@ -0,0 +1,85 @@
...
+    // Windows: "mega/win32/drivenotify.h" contains the platform specific definition.
+    class IDriveNotify
+    {
+    public:
+        using NotificationFunc = std::function<void(DriveInfo&&)>;
+
+        // Start receiving notifications for every drive [dis]connection.
+        // driveConnected: function called to notify about a connected drive
+        // driveDisconnected: function called to notify about a disconnected drive
+        virtual bool start(NotificationFunc driveConnected, NotificationFunc driveDisconnected) = 0;
```
mattw-mega, line 60: Rather than passing in functions here, the thread-safe queue to put events on can just be a public member of IDriveNotify.

mattw-mega (changes requested):
This is looking very good and very promising, nice one!   I've left some specific comments

contrib/cmake/CMakeLists.txt
catalinr-m: Ok, replaced `HAVE_DRIVE_NOTIFY` with `USE_DRIVE_NOTIFICATIONS`.

examples/megacli.cpp
catalinr-m: Actually the intended behavior was different than both suggested changes. It should indeed ignore errors and _continue_, using the default value, when an error has occurred.
`atoi()` does not throw exceptions, but it can cause undefined behavior.
By letting exception be thrown, it would still stop the execution of the command.
I can of course change the code the way you prefer; just pointing out the intended behavior.

mattw-mega: If the user supplied a parameter, we shouldn't silently ignore it - if we can't use it we should tell them there's been an error, either they mistyped or we misunderstood.

include/mega/megaapp.h
```diff
@@ -426,6 +426,11 @@ struct MEGA_API MegaApp
     virtual void dismissbanner_result(error) { }
 
     virtual ~MegaApp() { }
+
+    // External drive notifications
+    bool startDriveMonitor();
+    void stopDriveMonitor();
```
mattw-mega, line 432: MegaApp is just for callbacks - start/stop should be invoked by client->mDriveInfoCollector.start()/stop()

include/mega/posix/drivenotifyposix.h
```diff
@@ -0,0 +1,47 @@
...
+
+namespace mega {
+
+    // Posix: Platform specific definition
+    //
+    // Not implemented.
+    class DriveNotifyPosix : public IDriveNotify
+    {
+    public:
+        bool start(NotificationFunc driveConnected, NotificationFunc driveDisconnected) override;
```
mattw-mega, line 38: do we need parameters here?   All the implmentations of this class should be doing the same thing - queueing the details to be delivered, we don't need that to be indirect

include/mega/drivenotify.h
mattw-mega: I think DriveInfoCollector and this IDriveNotify can be the same class, with the derived implementations just detecting and queueing.

mattw-mega (commented):
Looking good, I just left some comments about how the class structure and start/stop can be a bit simpler.  thanks

include/mega/posix/drivenotifyposix.h
catalinr-m: Yeah, I thought it'd be useful to have them separated, but if they'll never be used like that, I'll just remove one of them.

include/mega/megaapp.h
catalinr-m: But... it's not, whether there are functions for that or direct access to a member, and IHMO the former is cleaner.


> DriveInfoCollector and this IDriveNotify can be the same class

(interestingly, I cannot reply to that comment directly)
Yup, you're right. Earlier, it seemed to make more sense like it currently is, but not so much anymore. I'll quickly update it.

include/mega/drivenotify.h
catalinr-m: Yup, you're right. Earlier, it seemed to make more sense like it currently is, but not so much anymore. I'll quickly update it.

catalinr-m: I've merged the 2 classes with the minimum amount of changes.
But if there should be any class & file renaming that will affect the rest of the code, I think now is the best time to do it (i.e. `DriveInfoCollectorBase` -> `DriveNotifyBase` or `DriveNotifyWin` -> `DriveInfoCollectorWin`, file names too).

Also, while at this, the new classes have no dependency upon other `mega` code. Should they be inside `mega` namespace or should I move them out?

contrib/cmake/CMakeLists.txt
sergiohs84: Does the same logic applies to the building with qmake? Please, @catalinr-m, make sure ;)

include/mega/drivenotify.h
sergiohs84: IMO they should belong to `mega` namespace.
As for refactoring, my vote goes for the `DriveNotifyBase`, or even `DriveNotify` (we don't have the suffix `Base` in existing classes), with the corresponding specializations `DriveNotifyWin`, etc

contrib/cmake/CMakeLists.txt
catalinr-m: Sorry, which logic are you asking about?
The build test was removed, and `HAVE_DRIVE_NOTIFY` -> `USE_DRIVE_NOTIFICATIONS` change was done in `qt\sdk.pri` too (though in a later commit).

include/mega/megaapp.h
sergiohs84: You can wrap them if you want, but MegaApp is only for callbacks.
If it feels cleaner, you can prepare something like:
``` cpp
void MegaClient::startDriveMonitor()
{
    mDriveInfoCollector.start();
}
```

examples/megacli.cpp
```diff
@@ -3165,7 +3164,8 @@ autocomplete::ACN autocompleteSyntax()
     p->Add(exec_metamac, sequence(text("metamac"), localFSPath(), remoteFSPath(client, &cwd)));
     p->Add(exec_banner, sequence(text("banner"), either(text("get"), sequence(text("dismiss"), param("id")))));
 
-    p->Add(exec_notifydrive, sequence(text("notifydrive"), opt(wholenumber(60))));
+    p->Add(exec_drivenotificationson, sequence(text("drivenotificationson")));
+    p->Add(exec_drivenotificationsoff, sequence(text("drivenotificationsoff")));
```
sergiohs84, line 3168: Probably nicer to have something like the existing `handles on|off` option :smiley:

include/mega/megaapp.h
catalinr-m: That's more or less what is already implemented. So IIUC, no change for this one, right?

include/mega/drivenotify.h
catalinr-m: Renamed symbols and files.

examples/megacli.cpp
catalinr-m: Updated.

include/mega/megaapp.h
mattw-mega: Every function declared in MegaApp is a callback to be overridden by the client, and called by the SDK core when the relevant event has occurred.   These two new functions break that design, so please move them elsewhere - either Sergio's suggestion or mine or something else is ok, but MegaApp is designed to be only for callbacks.    thanks

mattw-mega: Actually I think in the current version, these two function declarations are unused anyway?  They don't seem to be called or have a definition - it's the ones in MegaClient that are used - you can just remove those two lines.  thanks

include/mega/drivenotify.h
mattw-mega: File and class names look good to me, thanks

include/mega/megaapp.h
catalinr-m: They had definitions, and they were used from `MegaApi` and `megacli`.
Removed now.

include/mega/drivenotify.h
```diff
@@ -0,0 +1,97 @@
...
+    // Platform specific implementations:
+    // - DriveNotifyWin;
+    // - DriveNotifyPosix.
+    class DriveNotify
+    {
+    public:
+        bool start(std::function<void()> notify);
+        void stop();
+
+        std::pair<std::wstring, bool> get();
```
sergiohs84, line 64: This getter may need to be declared within an `#ifdef WIN32`:
```suggestion cpp
#ifdef WIN32
    std::pair<std::wstring, bool> get();
#elseif
    std::pair<std::string, bool> get();
#endif
```
Otherwise, on non-windows, the call to the callback in the `exec()` ([here](https://github.com/meganz/sdk/pull/2441/files#diff-fe6b11e559efca285b6c1c20b3de9a4cd3aba4c65837f872665ec883fdfa03b4R3141)) will throw a compiler error:
```
/home/sergio/dev/sdk/src/megaclient.cpp:3141: error: no matching function for call to ‘mega::LocalPath::fromPlatformEncoded(std::remove_reference<std::__cxx11::basic_string<wchar_t>&>::type)’
         app->drive_presence_changed(di.second, LocalPath::fromPlatformEncoded(move(di.first)));
                                                                                             ^
```
(Note: there `fromPlatformEncoded()` is overloaded for WIN32 in order to accept wide characters strings)

sergiohs84 (changes requested):

include/mega/drivenotify.h
```diff
@@ -0,0 +1,97 @@
...
+#pragma once
+
+#ifdef USE_DRIVE_NOTIFICATIONS
+
+#include <functional>
+#include <string>
+#include <queue>
+#include <mutex>
+
+namespace mega {
```
sergiohs84, line 31: _Code within a namespace is not indented._ 
Please, review the style guide

sergiohs84, line 31: _Code within a namespace is not indented._ 
Please, review the style guide

include/mega/drivenotify.h
sergiohs84: Fixed in a4f24c1

include/mega/drivenotify.h
sergiohs84: Fixed in e910844

include/mega/drivenotify.h
```diff
@@ -61,7 +68,7 @@ namespace mega {
         bool start(std::function<void()> notify);
         void stop();
 
-        std::pair<std::wstring, bool> get();
+    std::pair<DriveInfo::StringType, bool> get();
```
catalinr-m: Indentation was removed for this line

contrib/cmake/CMakeLists.txt
sergiohs84: Yes, I see the `sdk.pri` contains conditional code depending on the `USE_DRIVE_NOTIFICATIONS` is defined or not. However, while I can see that you set that config [here](https://github.com/meganz/sdk/pull/2441/files#diff-7851744e91538f5741554ee3a428b0901d30dce2c98dea27272330a817259f84R45) for building with cmake, I don't see anything analog for qmake+autotools. I guess that it needs to touch the `configure.ac` in a way that it defines it in the same way than for building with CMake. By doing so, Jenkins will start building + testing, since currently it's not building with `USE_DRIVE_NOTIFICATIONS`.

sergiohs84: Another option would be to hardcode the option in the `sdk.pri`, but I think that's not the right way (probably not for cmake either), since we don't want to build the SDK with this feature for all targets.

catalinr-m:
This PR shows 2 requested changes, but I think they are both outdated and no longer applicable. Please re-review them when convenient. Thanks.

PR #2445, opened 2020-12-15
SDK-1333. Have Posix Waiter use poll instead of select if USE_POLL defined

polmr:
This circumvents FD_SETSIZE limit for number of file descriptors allowed

### DIFF
--- a/include/mega/posix/megawaiter.h
+++ b/include/mega/posix/megawaiter.h
@@ -29,7 +29,7 @@
     #define MEGA_FD_ZERO FD_ZERO
     #define MEGA_FD_SET FD_SET
     #define MEGA_FD_ISSET FD_ISSET
-    #define MEGA_FD_SET_T fd_set
+    typedef fd_set mega_fd_set_t;
 #else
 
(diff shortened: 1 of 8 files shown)


mattw-mega (dismissed):

include/mega/posix/megawaiter.h
```diff
@@ -25,17 +25,52 @@
 #include "mega/waiter.h"
 #include <mutex>
 
+#ifndef USE_POLL
+    #define MEGA_FD_ZERO FD_ZERO
+    #define MEGA_FD_SET FD_SET
+    #define MEGA_FD_ISSET FD_ISSET
+    #define MEGA_FD_SET_T fd_set
```
sergiohs84, line 32: This one is not a macro, but a datatype. Perhaps we should use a `typedef`, and a naming in lowcase: `mega_fd_set_t` (or simply `fd_set_t`). Same for the definition if `USE_POLL`

polmr: agree. changed in new commit

sergiohs84 (approved):

PR #2446, opened 2020-12-16
Update Android CU heartbeat feature related bindings

ash-mega:
- Update methods document comments
- Remove state and status constants from bindings, now they are defined in Android layer with same value in `megaapi.h`
- Add other sync types to bindings


Yenel (approved):
Thanks Ash, changes seems fine.

PR #2449, opened 2020-12-16
iOS CU Heartbeat - add being logout notification to iOS binding

simonwng:
Add being logout notification so we can have a chance to remove backup registration before logout.

If we call remove backup after logout, it returns error Bad session id - API_ESID
If we call remove backup when logout request starts, it will return error Access Denied - API_EACCESS
So the solution is post a notification before we start logout.


lancy-mega (dismissed):

simonwng:
I will create a new PR to avoid the version changes introduced in the release/v3.7.9 branch.

PR #2450, opened 2020-12-16
iOS CU Heartbeat - add being logout notification to iOS binding

simonwng:
Add being logout notification so we can have a chance to remove backup registration before logout.

- If we call remove backup after logout, it returns error Bad session id - `API_ESID`
- If we call remove backup when logout request starts, it will return error Access Denied - `API_EACCESS`

So the solution is post a notification before we start logout.

### DIFF
--- a/bindings/ios/MEGASdk.h
+++ b/bindings/ios/MEGASdk.h
@@ -51,6 +51,11 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+/**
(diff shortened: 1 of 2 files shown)


lancy-mega (approved):

jnavarrom:
I'm not sure we should be doing this. 
For the correct functioning of an application written with the SDK using the Objective-c binding, these changes are not necessary. They are specific and necessary changes for our application and we should have fixed the problem in the application layer code IMHO

PR #2451, opened 2020-12-18
SDK-1336 - Implement more reliable KillSession SDK integration test.

dwmega:


mattw-mega (approved):

tests/integration/SdkTest_test.cpp
```diff
@@ -1265,6 +1293,67 @@ bool veryclose(double a, double b)
...
+      << "Unable to kill second client's session (error: "
+      << result
+      << ")";
+
+    // Wait for the second client to become logged out.
+    ASSERT_TRUE(WaitFor([&]()
+                        {
+                            return mApi[1].megaApi->isLoggedIn()  == 0;
+                        },
+                        8 * 1000));
```
sergiohs84, line 1351: Perhaps it would had been more consistent/robust to wait for the `API_ESID` received by the second client, which is notified to the app via a `onRequestFinish(TYPE_LOGOUT)`. In other words, closing the current session remotely, via another client, triggers a logout. See `void MegaApiImpl::request_error(error e)`

PR #2455, opened 2020-12-18
SDK-1328. Cookie dialog: new API flag and user attr

catalinr-m:


include/megaapi.h
```diff
@@ -18523,6 +18525,43 @@ class MegaApi
...
+         * @param listener MegaRequestListener to track this request
+         */
+        void setCookieSettings(int settings, MegaRequestListener *listener = nullptr);
+
+        /**
+         * @brief Get a bitmap to indicate whether some cookies are enabled or not
+         *
+         * The associated request type with this request is MegaRequest::TYPE_GET_COOKIE_SETTINGS
+         * Valid data in the MegaRequest object received on callbacks:
+         *  - MegaRequest::getNumDetails A bitmap to indicate which cookies are enabled or not
```
sergiohs84, line 18552: It's incorrect. Only at `onRequestFinish()` with `API_OK` such param will be set.
We also should document the other fields of the request that are valid on (all) callbacks, like the `getParamType()`.

include/megaapi.h
```diff
@@ -18523,6 +18525,43 @@ class MegaApi
          */
         void queryGoogleAds(int adFlags, MegaHandle publicHandle = INVALID_HANDLE, MegaRequestListener *listener = nullptr);
 
+        /**
+         * @brief Set a bitmap to indicate whether some cookies are enabled or not
+         *
+         * The associated request type with this request is MegaRequest::TYPE_SET_COOKIE_SETTINGS
```
sergiohs84, line 18531: Incorrect: `TYPE_SET_ATTR_USER`
```suggestion
         * The associated request type with this request is MegaRequest::TYPE_SET_ATTR_USER
```

include/megaapi.h
```diff
@@ -18523,6 +18525,43 @@ class MegaApi
...
+         *      - Bit 3: ads
+         *      - Bit 4: thirdparty
+         * @param listener MegaRequestListener to track this request
+         */
+        void setCookieSettings(int settings, MegaRequestListener *listener = nullptr);
+
+        /**
+         * @brief Get a bitmap to indicate whether some cookies are enabled or not
+         *
+         * The associated request type with this request is MegaRequest::TYPE_GET_COOKIE_SETTINGS
```
sergiohs84, line 18550: Incorrect: `TYPE_GET_ATTR_USER`
```suggestion
         * The associated request type with this request is MegaRequest::TYPE_GET_ATTR_USER
```

src/megaapi_impl.cpp
```diff
@@ -23582,6 +23597,50 @@ void MegaApiImpl::queryGoogleAds(int adFlags, MegaHandle publicHandle, MegaReque
...
+    strncpy(startptr, (const char*)data, len);
+
+    // get the value
+    char* endptr;
+    long value = strtol(startptr, &endptr, 10);
+
+    // validate the value
+    if (endptr == startptr || *endptr != '\0' || value == LONG_MAX || value == LONG_MIN)
+    {
+        value = -1;
```
sergiohs84, line 23638: It would be good to log an error message here. If we fail to validate the value, we should even report an error (`API_EINTERNAL`), instead of silently ignore the error. An example: https://github.com/meganz/sdk/pull/2455/files#diff-5d048e8562b9eef0c31b18c647b312f4abf03481607422ffb7f1b12814524fc0R15501

sergiohs84 (changes requested):
Please, @catalinr-m, see my comments and proceed with the implementation of the API flag `cspe`

include/megaapi.h
```diff
@@ -18523,6 +18525,48 @@ class MegaApi
...
+
+        /**
+         * @brief Get a bitmap to indicate whether some cookies are enabled or not
+         *
+         * The associated request type with this request is MegaRequest::TYPE_GET_ATTR_USER
+         * Valid data in the MegaRequest object received on callbacks:
+         *  - MegaRequest::getParamType - Returns the value USER_ATTR_COOKIE_SETTINGS
+         *  - MegaRequest::getListener - Returns the MegaRequestListener to track this request
+         *
+         * When the error code is MegaError::API_OK:
```
sergiohs84, line 18556: Let's use the same template that is used in the rest of the documentation, please:
```suggestion cpp
         * Valid data in the MegaRequest object received in onRequestFinish when the error code
         * is MegaError::API_OK:
```

sergiohs84 (changes requested):

include/megaapi.h
catalinr-m: That was actually the intention, so I scrolled a bit up, to the neighboring (and new) `queryGoogleAds`, and picked it up from there... in case that should be fixed too.

catalinr-m:
I've added implementation of the API flag `cspe`. Please let me know if there is anything else left here.

include/megaapi.h
sergiohs84: Yes, please. Adjust it to be in line with the most common phrasing :)

include/megaapi.h
```diff
@@ -18523,6 +18527,63 @@ class MegaApi
          */
         void queryGoogleAds(int adFlags, MegaHandle publicHandle = INVALID_HANDLE, MegaRequestListener *listener = nullptr);
 
+        /**
+         * @brief Set a bitmap to indicate whether some cookies are enabled or not
+         *
+         * The associated request type with this request is MegaRequest::TYPE_SET_ATTR_USER
+         * Valid data in the MegaRequest object received on callbacks:
+         *  - MegaRequest::getParamType - Returns the attribute type MegaApi::USER_ATTR_COOKIE_SEETINGS
```
javisanesp, line 18535: Typo

```suggestion
         *  - MegaRequest::getParamType - Returns the attribute type MegaApi::USER_ATTR_COOKIE_SETTINGS
```

include/megaapi.h
```diff
@@ -18523,6 +18527,63 @@ class MegaApi
...
+         * - MegaRequest::getNumDetails Return the bitmap with cookie settings
+         *   Valid bits are:
+         *      - Bit 0: essential
+         *      - Bit 1: preference
+         *      - Bit 2: analytics
+         *      - Bit 3: ads
+         *      - Bit 4: thirdparty
+         *
+         * On the onRequestFinish error, the error code associated to the MegaError can be:
+         * - MegaError::API_EINTERNAL - If the value for cookie settings bitmap was invalid
```
javisanesp, line 18569: The `param` line on the documentation is missing for this method.

```suggestion
         * - MegaError::API_EINTERNAL - If the value for cookie settings bitmap was invalid
         *
         * @param listener MegaRequestListener to track this request
```

src/megaapi_impl.cpp
```diff
@@ -23582,6 +23597,60 @@ void MegaApiImpl::queryGoogleAds(int adFlags, MegaHandle publicHandle, MegaReque
...
+    MegaRequestPrivate* request = new MegaRequestPrivate(MegaRequest::TYPE_SET_ATTR_USER, listener);
+    request->setParamType(MegaApi::USER_ATTR_COOKIE_SETTINGS);
+    request->setNumDetails(settings);
+    requestQueue.push(request);
+    waiter->notify();
+}
+
+void MegaApiImpl::setCookieSettings_sendPendingRequests(MegaRequestPrivate* request)
+{
+    auto tmp = std::to_string(request->getNumDetails());
```
sergiohs84, line 23611: Is this conversion really what we want to? I think it's not. We want to send the bytes corresponding to the integer passed by the app. We don't want a C-string representation of the number, do we?

src/megaapi_impl.cpp
```diff
@@ -23582,6 +23597,60 @@ void MegaApiImpl::queryGoogleAds(int adFlags, MegaHandle publicHandle, MegaReque
...
+{
+    // make a copy to make sure we don't read too much from a non-null terminated string
+    unique_ptr<char[]> buff(new char[len + 1]);
+    buff[len] = 0;
+    char* startptr = buff.get();
+    strncpy(startptr, (const char*)data, len);
+
+    // get the value
+    char* endptr;
+    long value = strtol(startptr, &endptr, 10);
```
sergiohs84, line 23633: Actually, I see that you undo the previous `int` -> `string` here, by a `string` -> `long` (which is wrong, since we don't need such a wide range of values, and later you re-convert `long` -> `int`).

sergiohs84 (changes requested):
I miss the parsing of this new attribute in the response of `ug` command. It should be included, when it exists, and we do want to initialize it there, as it happens with many other attributes.

catalinr-m:
> I miss the parsing of this new attribute in the response of `ug` command. It should be included, when it exists, and we do want to initialize it there, as it happens with many other attributes.

For what is worth, most of variables are initialized and reset, in the following functions:
- `MegaClient::init()` - some of them;
- `MegaClient()` `constructor`, and `MegaClient::locallogout()` all of them; some through calling `init()`.

`CommandGetMiscFlags::procresult()` does _not_ initialize any variable, but calls `MegaClient::readmiscflags()` which will only give them new values when received.
`CommandGetUserData::procresult()` does initialize some variables, but it also calls `MegaClient::readmiscflags()` which behaves as above. So should `readmiscflags()` init that flag?

For now I've added the init to `CommandGetUserData::procresult()` (hopefully in the right place), ...and again to the constructor, which apparently missed it.

src/megaapi_impl.cpp
catalinr-m: It may very well be so, however, it was an attempt to follow the code for `ATTR_RUBBISH_TIME` (a.k.a. _an integer 32 bits is the number of days for the auto-purge Rubbish bin ( ^!rubbishtime). It can be used as reference_), as indicated in the ticket description.

src/megaapi_impl.cpp
catalinr-m: It could at most be slightly ineficient, i.e. on LP64 systems where sizeof(int) < sizeof(long), but even that should happen only for bad values, where instead of stopping after the first 4 bytes it would continue to interpret another 4. But the real reason for using `strtol` was that it is C++ and there is no `strtoi`...

Is there a preferred function for converting `string` to `int`?

sergiohs84:
> > I miss the parsing of this new attribute in the response of `ug` command. It should be included, when it exists, and we do want to initialize it there, as it happens with many other attributes.
> 
> For what is worth, most of variables are initialized and reset, in the following functions:
> 
>     * `MegaClient::init()` - some of them;
> 
>     * `MegaClient()` `constructor`, and `MegaClient::locallogout()` all of them; some through calling `init()`.
> 
> 
> `CommandGetMiscFlags::procresult()` does _not_ initialize any variable, but calls `MegaClient::readmiscflags()` which will only give them new values when received.
> `CommandGetUserData::procresult()` does initialize some variables, but it also calls `MegaClient::readmiscflags()` which behaves as above. So should `readmiscflags()` init that flag?
> 
> For now I've added the init to `CommandGetUserData::procresult()` (hopefully in the right place), ...and again to the constructor, which apparently missed it.

It's not the initialization of the API flag `cspe`, which is already done in the `readmiscflags()`, but the early initialization of the user's attribute `^!csp`, which should be returned by the `ug` command when it exists (like it happens for `^clv`, in example).

src/megaapi_impl.cpp
sergiohs84: Good point, Catalin. Thanks for double-checking. Let's stick to the current approach :+1:

sergiohs84 (dismissed):

sergiohs84 (approved):

PR #1639, opened 2019-10-03
#10991: Allow mute unmute chats individually feature iOS SDK bindings.

lancy-mega:
Allow mute unmute chats individually feature iOS SDK bindings.

PR #2463, opened 2020-12-24
IOS-4485: iOS binding for PSA.

lancy-mega:
IOS-4485: iOS binding for PSA.

### DIFF
--- a/bindings/ios/MEGASdk.h
+++ b/bindings/ios/MEGASdk.h
@@ -2503,6 +2503,63 @@ typedef NS_ENUM(NSUInteger, BackupHeartbeatStatus) {
 - (void)getPSA;
 
 /**
+ * @brief Get the next PSA (Public Service Announcement) that should be shown to the user
+ *
+ * After the PSA has been accepted or dismissed by the user, app should
+ * use [MEGASdk setPSAWithIdentifier:] [MEGASdk setPSAWithIdentifier:delegate:] to notify API servers about
+ * this event and do not get the same PSA again in the next call to this function.
+ *
+ * The associated request type with this request is MEGARequestTypeGetPSA.
+ *
+ * Valid data in the MEGARequest object received in onRequestFinish when the error code
+ * is MEGAErrorTypeApiOk:
+ * - [MEGARequest number] - Returns the id of the PSA (useful to call [MEGASdk setPSAWithIdentifier:]
+ *                          [MEGASdk setPSAWithIdentifier:delegate:] later)
+ * - [MEGARequest email] - Returns the URL (or an empty string)
+ * ...or for the old format:
+ * - [MEGARequest name] - Returns the title of the PSA
+ * - [MEGARequest text] - Returns the text of the PSA
+ * - [MEGARequest file] - Returns the URL of the image of the PSA
+ * - [MEGARequest password] - Returns the text for the possitive button (or an empty string)
+ * - [MEGARequest link] - Returns the link for the possitive button (or an empty string)
+ *
+ * If there isn't any new PSA to show, onRequestFinish will be called with the error
+ * code MEGAErrorTypeApiENoent
+ *
+ * @param delegate MEGARequestDelegate to track this request
+ * @see [MEGASdk setPSAWithIdentifier:] [MEGASdk setPSAWithIdentifier:delegate:]
+ */
+- (void)getPSAIncludingUrlWithDelegate:(id<MEGARequestDelegate>)delegate;
+
+/**
+ * @brief Get the next PSA (Public Service Announcement) that should be shown to the user
+ *
+ * After the PSA has been accepted or dismissed by the user, app should
+ * use [MEGASdk setPSAWithIdentifier:] [MEGASdk setPSAWithIdentifier:delegate:] to notify API servers about
+ * this event and do not get the same PSA again in the next call to this function.
+ *
+ * The associated request type with this request is MEGARequestTypeGetPSA.
+ *
+ * Valid data in the MEGARequest object received in onRequestFinish when the error code
+ * is MEGAErrorTypeApiOk:
+ * - [MEGARequest number] - Returns the id of the PSA (useful to call [MEGASdk setPSAWithIdentifier:]
+ *                          [MEGASdk setPSAWithIdentifier:delegate:] later)
+ * - [MEGARequest email] - Returns the URL (or an empty string)
+ * ...or for the old format:
+ * - [MEGARequest name] - Returns the title of the PSA
+ * - [MEGARequest text] - Returns the text of the PSA
+ * - [MEGARequest file] - Returns the URL of the image of the PSA
+ * - [MEGARequest password] - Returns the text for the possitive button (or an empty string)
+ * - [MEGARequest link] - Returns the link for the possitive button (or an empty string)
+ *
+ * If there isn't any new PSA to show, onRequestFinish will be called with the error
+ * code MEGAErrorTypeApiENoent
+ *
+ * @see [MEGASdk setPSAWithIdentifier:] [MEGASdk setPSAWithIdentifier:delegate:]
+ */
+- (void)getPSAIncludingUrl;
+
+/**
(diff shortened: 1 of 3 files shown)


bindings/ios/MEGASdk.h
```diff
@@ -2502,6 +2502,63 @@ typedef NS_ENUM(NSUInteger, BackupHeartbeatStatus) {
  */
 - (void)getPSA;
 
+/**
+ * @brief Get the next PSA (Public Service Announcement) that should be shown to the user
+ *
+ * After the PSA has been accepted or dismissed by the user, app should
+ * use [MEGASdk setPSAWithIdentifier:] [MEGASdk setPSAWithIdentifier:delegate:] to notify API servers about
```
simonwng, line 2509:
```suggestion
 * use [MEGASdk setPSAWithIdentifier:] or [MEGASdk setPSAWithIdentifier:delegate:] to notify API servers about
```

bindings/ios/MEGASdk.h
```diff
@@ -2502,6 +2502,63 @@ typedef NS_ENUM(NSUInteger, BackupHeartbeatStatus) {
...
+ * this event and do not get the same PSA again in the next call to this function.
+ *
+ * The associated request type with this request is MEGARequestTypeGetPSA.
+ *
+ * Valid data in the MEGARequest object received in onRequestFinish when the error code
+ * is MEGAErrorTypeApiOk:
+ * - [MEGARequest number] - Returns the id of the PSA (useful to call [MEGASdk setPSAWithIdentifier:]
+ *                          [MEGASdk setPSAWithIdentifier:delegate:] later)
+ * - [MEGARequest email] - Returns the URL (or an empty string)
+ * ...or for the old format:
```
simonwng, line 2519: ?

bindings/ios/MEGASdk.h
```diff
@@ -2502,6 +2502,63 @@ typedef NS_ENUM(NSUInteger, BackupHeartbeatStatus) {
...
+ * - [MEGARequest password] - Returns the text for the possitive button (or an empty string)
+ * - [MEGARequest link] - Returns the link for the possitive button (or an empty string)
+ *
+ * If there isn't any new PSA to show, onRequestFinish will be called with the error
+ * code MEGAErrorTypeApiENoent
+ *
+ * @param delegate MEGARequestDelegate to track this request
+ * @see [MEGASdk setPSAWithIdentifier:] [MEGASdk setPSAWithIdentifier:delegate:]
+ */
+- (void)getPSAIncludingUrlWithDelegate:(id<MEGARequestDelegate>)delegate;
```
simonwng, line 2532: The method name is a bit verbose... what do you think of `getURLPSAWithDelegate`?

bindings/ios/MEGASdk.h
```diff
@@ -2502,6 +2502,63 @@ typedef NS_ENUM(NSUInteger, BackupHeartbeatStatus) {
...
+ * - [MEGARequest file] - Returns the URL of the image of the PSA
+ * - [MEGARequest password] - Returns the text for the possitive button (or an empty string)
+ * - [MEGARequest link] - Returns the link for the possitive button (or an empty string)
+ *
+ * If there isn't any new PSA to show, onRequestFinish will be called with the error
+ * code MEGAErrorTypeApiENoent
+ *
+ * @see [MEGASdk setPSAWithIdentifier:] [MEGASdk setPSAWithIdentifier:delegate:]
+ */
+- (void)getPSAIncludingUrl;
```
simonwng, line 2560: If we always pass a delegate (I think we should), then we don't need this method in binding.

lancy-mega: Yup right. We might end up not using it. Previously, I got a PR comment for not including the method without the delegate. https://github.com/meganz/sdk/pull/1639#discussion_r334799287
I think the general practice has been including both. @jnavarrom - Are you fine not including it?

simonwng: if we are following that practice, I think it is OK to add it.

bindings/ios/MEGASdk.h
lancy-mega: I think `URLPSA` is hard to notice that it is `URL - PSA` as everything is in caps. Let me know your thoughts.

bindings/ios/MEGASdk.h
lancy-mega: | SDK  | Android |
| ------------- | ------------- |
|  |  |

As I understand from SDK and Android documentation is that the new format of the PSA return URL and the old format returns information that can be used to display inside the app.

bindings/ios/MEGASdk.h
simonwng: haha, yeah, two abbreviations together. According to both Swift and Objective-c naming guide, abbreviations should avoid. Only standard ones can be acceptable. So I think we should go for:
`getURLPublicServiceAnnouncementWithDelegate`


https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/CodingGuidelines/Articles/APIAbbreviations.html#//apple_ref/doc/uid/20001285

https://swift.org/documentation/api-design-guidelines/
 
>Avoid abbreviations. Abbreviations, especially non-standard ones, are effectively terms-of-art, because understanding depends on correctly translating them into their non-abbreviated forms.The intended meaning for any abbreviation you use should be easily found by a web search.

bindings/ios/MEGASdk.h
simonwng: I mean for format of the comment is weird. why we do `...` here

simonwng: is `...or` a standard comment format?

bindings/ios/MEGASdk.h
jnavarrom: The main problem is that objective-c has no optional parameters such as c++.
I think we can add both methods, if the intermediate layer (megaapi) optionally allows to pass a listener as argument or not, the binding should do the same. We usually use the method by passing a delegate as an argument, but other developers could create their applications without passing the delegate.

Initially both were added: with delegate and without it... 
Another option would be to add the methods on demand as we or other developers need.

bindings/ios/MEGASdk.h
sergiohs84: When `getPSA()` is called, the old format is always used.
When `getPSAwithUrl()` is called, the SDK sends to the API an additional flag `"w":1` for the command `gpsa`. The API can return still return the old format, or the new one with only a URL (see SDK-1247).

jnavarrom: I think @simonwng is referring the ... in the documentation:

  * - MegaRequest::getEmail - Returns the URL (or an empty string))
  * ...or for the old format:   <------ THIS LINE

simonwng: Thanks @sergiohs84 , so according to the PR https://github.com/meganz/sdk/pull/2286/files, `getPSAwithUrl` returns both URL and old format data. I would suggest removing this line of comment:
`..or for the old format:`

bindings/ios/MEGASdk.h
simonwng: Thanks Navarro, I am OK with either way. It is easy to pass an empty delegate anyway, haha. Maybe to make it simple, we can add the non-delegate APIs on demand.

bindings/ios/MEGASdk.h
sergiohs84: Sure. Please, add a commit to update the documentation accordingly, @simonwng . Thanks

simonwng: @lancy-mega , please remove the `..or` comment from megaapi.h as well

simonwng: @lancy-mega , please remove the `..or` comment from megaapi.h as well

simonwjw (approved):

simonwng (approved):

sergiohs84 (approved):

PR #2471, opened 2021-01-07
SDK-1345. Enable session resume for writable folder links.

mattw-mega:
Session resume is also enabled for read-only folder links.
Resuming means we can load the locally cached node database instead of re-requesting from the server, just like a normal account
Of course for folder links it's not really using a session id, but follows the pattern of calling dumpsession() and re-login with the output of that.

### DIFF
--- a/src/megaclient.cpp
+++ b/src/megaclient.cpp
@@ -8248,7 +8248,7 @@ int MegaClient::readnodes(JSON* j, int notify, putsource_t source, vector<NewNod
                 Node::copystring(n->attrstring.get(), a);
                 n->setkeyfromjson(k);
 
-                // folder link access: first returned record defines root node and identity 
+                // folder link access: first returned record defines root node and identity
 				// (this code used to be in Node::Node but is not suitable for session resume)
                 if (ISUNDEF(*rootnodes))
                 {
@@ -9254,12 +9254,12 @@ int MegaClient::dumpsession(string& session)
             SymmCipher cipher;
             cipher.setkey((const byte *)sessionkey.data(), int(sessionkey.size()));
             cipher.ecb_encrypt(key.key, k);
-            memcpy(session.data()+1, k, sizeof k);
+            memcpy(const_cast<char*>(session.data())+1, k, sizeof k);
         }
         else
         {
             session.resize(sizeof key.key);
-            memcpy(session.data(), key.key, sizeof key.key);
+            memcpy(const_cast<char*>(session.data()), key.key, sizeof key.key);
         }
 
         session.append(sid.data(), sid.size());
--- a/src/megaclient.cpp
+++ b/src/megaclient.cpp
@@ -475,42 +475,6 @@ void MegaClient::mergenewshare(NewShare *s, bool notify)
     }
 }
 
-// configure for full account session access
-void MegaClient::setsid(const byte* newsid, unsigned len)
-{
-    auth = "&sid=";
-
-    size_t t = auth.size();
-    auth.resize(t + len * 4 / 3 + 4);
-    auth.resize(t + Base64::btoa(newsid, len, (char*)(auth.c_str() + t)));
-
-    sid.assign((const char*)newsid, len);
-}
-
-// configure for exported folder links access
-void MegaClient::setrootnode(handle h, const char *authKey)
-{
-    char buf[12];
-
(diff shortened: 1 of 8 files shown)


src/megaclient.cpp
```diff
@@ -15791,9 +15881,14 @@ handle MegaClient::getovhandle(Node *parent, string *name)
...
 
+bool MegaClient::loggedIntoFolder() const
+{
+    return !ISUNDEF(publichandle);
+}
+
 bool MegaClient::loggedIntoWritableFolder() const
 {
-    return mLoggedIntoWritableFolder;
+    return !ISUNDEF(publichandle) && !publichandleWriteAuth.empty();
```
sergiohs84, line 15891: What about:
```suggestion cpp
    return loggedIntoFolder() && !publichandleWriteAuth.empty();
```

src/megaclient.cpp
```diff
@@ -9151,6 +9171,41 @@ void MegaClient::login(const byte* session, int size)
...
+    else if (!session.empty() && session[0] == 2)
+    {
+        // folder link - read only or writable
+
+        CacheableReader cr(session);
+
+        byte v;
+        handle ph, rh;
+        byte k[FOLDERNODEKEYLENGTH];
+        string a1, a2, padding;
```
sergiohs84, line 9183: Could we give meaningful names to the variables, please? :pray:

sergiohs84 (commented):
Perhaps the `logout` command should remove the "session" used to log in into the folder link. It would be more in line with the behavior for account (despite the session is not invalidated by a logout). Feel free to disregard the suggestion.

src/megaclient.cpp
mattw-mega: The class already has member variables such as `publichandle` (which is what ph is short for) - keeping these variables short like that means we won't accidentally use a member variable instead of one of these temporary holding variables.   And I think the pattern is clear - for every one of these, we extract it from the string input, and if all of them are successfully extracted, then the values are assigned to the member variables that these represent.   They aren't long lived, each one is only mentioned twice (plus once to declare) and within a few lines of each other, and you can see exactly which member variable the data actually goes into in the last mention.    If we extracted those directly into the member variables, and failed halfway, the state would be a mess so it's much better to confirm extraction of all first, and then update all the member variables in an all-or-nothing "strong guarantee" fashion.
Thanks very much for the review, and for testing the changes too.
About logout, that is an interesting thought, and something I wondered about too.   However, it's possible that the folder link might be in use in multiple places (eg.  same app running on main phone and backup phone, both logged into the same link - or personal phone and work phone), and we wouldn't want logging out of one to cause the other to fail, I think.    Probably in this scenario, cancelling the link from the same place it was created (typically from the webclient) would be the way to go.

sergiohs84: Hi Matt. I disagree with your arguments to use these variable names. It's unclear to me what's `a1` and `a2` purpose, in example, requiring to go through `setrootnode()` and see the usage. A name like `publicHandleWriteAuth` will help readability. In case the variable name in the class clashes with the local variable's name, it's because the class member needs refactoring to `mClassMember`, with the prefix, which aims to avoid those clashes and also invalidates the point: _we won't accidentally use a member variable instead of..._
I'll apply the renaming.

As for using temporary buffers for unserializing, I agree it's the way to go. Exactly the same that is usually done when parsing data from API for many commands. No problem at all with that approach.

sergiohs84:
> About logout, that is an interesting thought, and something I wondered about too. However, it's possible that the folder link might be in use in multiple places (eg. same app running on main phone and backup phone, both logged into the same link - or personal phone and work phone), and we wouldn't want logging out of one to cause the other to fail, I think. Probably in this scenario, cancelling the link from the same place it was created (typically from the webclient) would be the way to go.

As for the logout, I don't understand the use case you mention. When the SDK instance is logged into a folder link, it's not logged into the owner's account. Therefore, a logout will never invalidate the folder link itself, it cannot, but it can forget about the "session" that is saved (the public handle and the auth token). In other words, when the SDK is logged in into an account, a logout invalidates the session. I'm suggesting whether the SDK logged into a folder link should invalidate (meaning to delete, in this case) the "session" upon logout (despite both public handle and auth token remain valid, but for analogy with a user's session).

sergiohs84 (approved):

PR #2462, opened 2020-12-22
SDK-1302. Register sync in backup center (sp) before being added anew or copied from app cache into sdk's

polmr:
Added completion functions to addsync and new copySyncConfig methods.

PR #2477, opened 2021-01-13
SDK-1351. Avoid calling onSyncStateChanged before onSyncAdded

mattw-mega:
It was happening for resumed syncs.
Also updated the auto tests to confirm this an catch any future change that could reintroduce the issue.

### DIFF
--- a/tests/integration/SdkTest_test.cpp
+++ b/tests/integration/SdkTest_test.cpp
@@ -32,6 +32,11 @@
 #define LOCAL_TEST_FOLDER (string(getenv("HOME"))+"/synctests_mega_auto")
 #endif
 
+
+#define SSTR( x ) static_cast< const std::ostringstream & >( \
+        (  std::ostringstream() << std::dec << x ) ).str()
+
+
 using namespace std;
 
 MegaFileSystemAccess fileSystemAccess;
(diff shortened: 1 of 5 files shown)


tests/integration/SdkTest_test.cpp
```diff
@@ -5656,11 +5656,128 @@ TEST_F(SdkTest, SyncBasicOperations)
...
+        }
+        return stateMap[sync->getTag()];
+    }
+
+    bool anyErrors = false;
+    void check(bool b)
+    {
+        if (!b)
+        {
+            assert(false);
```
sergiohs84, line 5684: I think we should avoid this `assert`. Otherwise, tasks may fail prematurely rather than in the right place when the `anyErrors` is checked. Also, the `assert` will prevent the rest of tests to run. 
Remember there could be some tasks in Jenkins that run in Debug.

sergiohs84 (changes requested):

sergiohs84:
@mattw-mega sorry for the inconvenience, but I'd tried to rebase this branch with Pablo's, and it ended up with 2 commits that shouldn't be there. I've cherry-picked your original commit directly to Pablo's branch here: https://github.com/meganz/sdk/pull/2462/commits/92eec949c22d7bb68fe571478cd36889f6eb5a25

PR #2485, opened 2021-01-15
SDK-1357. fix crash on ephemeral account creation

mattw-mega:
(cherry picked from commit 11fba17aadb30fe1003cdb5fe31a326582b1daf2)


sergiohs84:
This release is not suitable for sync-enabled clients, so no need to patch it.

PR #2488, opened 2021-01-15
 SDK-1328. Cookie dialog: new API flag and user attr (to release/v3.8.0)

sergiohs84:
This feature might be included in the release v3.8.0, if Robert confirms to do so


sergiohs84:
Finally decide to not include this feature in the v3.8.0. Closing the PR...

PR #2494, opened 2021-01-19
SDK-1362. Fix issues with remote paths changes in syncs

polmr:
updates on remote paths for syncs are only considered for active syncs after sync revamping. 
This reverts that and makes notice of inactive syncs changes in remote paths.
Thus allows for apps to notice that and also be able to re-enable syncs whose remote went to the //bin an back

### DIFF
--- a/src/megaclient.cpp
+++ b/src/megaclient.cpp
@@ -7094,19 +7094,21 @@ void MegaClient::notifypurge(void)
 
 #ifdef ENABLE_SYNC
 
-        // fail active syncs
         //update sync root node location and trigger failing cases
         handle rubbishHandle = rootnodes[RUBBISHNODE - ROOTNODE];
         // check for renamed/moved sync root folders
-        syncs.forEachRunningSync([&](Sync* sync) {
+        //syncs.forEachRunningSync([&](Sync* sync) {
+        syncs.forEachUnifiedSync([&](UnifiedSync& us){
+            // fail active syncs
 
-            Node* n = nodebyhandle(sync->getConfig().getRemoteNode());
+
+            Node* n = nodebyhandle(us.mConfig.getRemoteNode());
             if (n && (n->changed.attrs || n->changed.parent || n->changed.removed))
             {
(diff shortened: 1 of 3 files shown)


src/megaclient.cpp
```diff
@@ -7094,19 +7094,21 @@ void MegaClient::notifypurge(void)
 
 #ifdef ENABLE_SYNC
 
-        // fail active syncs
         //update sync root node location and trigger failing cases
         handle rubbishHandle = rootnodes[RUBBISHNODE - ROOTNODE];
         // check for renamed/moved sync root folders
-        syncs.forEachRunningSync([&](Sync* sync) {
+        //syncs.forEachRunningSync([&](Sync* sync) {
```
sergiohs84, line 7100:
```suggestion
```

sergiohs84 (commented):
It looks good to me. But let's wait for @mattw-mega review, please.
Thanks

mattw-mega (changes requested):
Hi, now that we are using forEachUnifiedSync, it will also process Syncs which have a config but have failed already.   I think once a sync failed, we should not change the reason (eg. sync fails and a week later the root node is moved to the trash.  We shouldn't update the failure code at that time), so we should also add a condition for this processing, that the sync is at least not currently in an error state.

polmr:
> Hi, now that we are using forEachUnifiedSync, it will also process Syncs which have a config but have failed already. I think once a sync failed, we should not change the reason (eg. sync fails and a week later the root node is moved to the trash. We shouldn't update the failure code at that time), so we should also add a condition for this processing, that the sync is at least not currently in an error state.

Ey Matt, there is a check in place for us.mSync (there is an active sync) before changing the state [failSync]:-)
I'm pushing a commit to move the check up, to avoid looping to find the rubbish bin and clarify the code a little bit.

vtmateos:
Tested ok on macOS.

sergiohs84 (approved):

mattw-mega (approved):

PR #2495, opened 2021-01-19
SDK-1363. Cache Pro Level to detect upgrades

sergiohs84:
As a side-effect of SDK-1327, the app gets one `onAccountUpdate()` upon initialization, even when the account is not down/upgraded. To avoid it, we need to know the previous value. This commit aims to solve it.

### DIFF
--- a/include/mega/megaclient.h
+++ b/include/mega/megaclient.h
@@ -889,7 +889,7 @@ public:
     storagestatus_t ststatus;
 
     // cacheable status
-    std::map<int64_t, std::shared_ptr<CacheableStatus>> mCachedStatus;
+    std::map<int64_t, CacheableStatus> mCachedStatus;
 
     // warning timestamps related to storage overquota in paywall mode
     vector<m_time_t> mOverquotaWarningTs;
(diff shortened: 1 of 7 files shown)


include/mega/megaclient.h
```diff
@@ -1834,8 +1831,14 @@ class MEGA_API MegaClient
     void setKeepSyncsAfterLogout(bool keepSyncsAfterLogout);
 #endif
 
-    void loadCacheableStatus(std::shared_ptr<CacheableStatus> status);
+    // adds the new record to the map in memory and to the DB. Also initialized dedicated vars
```
polmr, line 1834:
```suggestion
    // adds the new record to the map in memory and to the DB. Also initializes dedicated vars
```

src/utils.cpp
```diff
@@ -2247,7 +2247,7 @@ bool CacheableStatus::serialize(std::string* data)
     return const_cast<const CacheableStatus*>(this)->serialize(*data);
 }
 
-std::shared_ptr<CacheableStatus> CacheableStatus::unserialize(class MegaClient *client, const std::string& data)
+CacheableStatus* CacheableStatus::unserialize(class MegaClient *client, const std::string& data)
```
polmr, line 2250: we probably want to return nullptr in the fail cases

include/mega/types.h
```diff
@@ -820,7 +821,7 @@ class CacheableStatus : public Cacheable
     bool serialize(string* data) override;
 
     // deserializes the string to a SyncConfig object. Returns null in case of failure
```
polmr, line 823:
```suggestion
    // deserializes the string to a SyncConfig object. Returns null in case of failure.
    // returns a pointer to the unserialized value, owned by client
```

sergiohs84: It's the same for other `unserialize()` methods, but doc never hurts :) I'll add it.

src/utils.cpp
sergiohs84: Totally agree. I wonder why the compiler doesn't complain. A pointer to `CacheableStatus` shouldn't be infered from `{}`.
I'll fix it.

mattw-mega (approved):

PR #2505, opened 2021-01-25
iOS Binding - add ability to run request listener callback in background threads

simonwng:
https://jira.developers.mega.co.nz/browse/IOS-4463

This is to boost app performance to move away from unnecessary UI threads switching. It is recommended to use ListenerQueueTypeCurrent to avoid thread switching and boost app performance.

SDK Objective-C++ Binding Guide:
https://confluence.developers.mega.co.nz/pages/viewpage.action?pageId=3639396#SDKObjectiveC++BindingGuide-ListenerCallbacks

### DIFF
--- a/bindings/ios/MEGASdk.mm
+++ b/bindings/ios/MEGASdk.mm
@@ -64,6 +64,7 @@ using namespace mega;
 @property (nonatomic, assign) std::set<DelegateMEGALoggerListener *>activeLoggerListeners;
 
 - (MegaRequestListener *)createDelegateMEGARequestListener:(id<MEGARequestDelegate>)delegate singleListener:(BOOL)singleListener;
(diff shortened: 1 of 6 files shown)


bindings/ios/MEGASdk.mm
```diff
@@ -2604,9 +2605,13 @@ - (void)sendEvent:(NSInteger)eventType message:(NSString *)message {
 #pragma mark - Private methods
 
 - (MegaRequestListener *)createDelegateMEGARequestListener:(id<MEGARequestDelegate>)delegate singleListener:(BOOL)singleListener {
+    return [self createDelegateMEGARequestListener:delegate singleListener:singleListener queueType:ListenerQueueTypeMain];
```
simonwng, line 2608: Use `ListenerQueueTypeMain` for backwards compatibility, so we don't need to change any existing code.

It is recommended to use `ListenerQueueTypeCurrent` if possible. I have audited CU, CU target folder and CU heartbeat related APIs to use `ListenerQueueTypeCurrent`.

lancy-mega (approved):

PR #2508, opened 2021-01-26
SDK-1372. Protect against invalid `statusTable`

sergiohs84:
iOS reported crashes at `fetchStatusTable()`. Analog to the checkup of `sctable` before calling `fetchsc()`, the same should apply to the status table too.


alber2510 (approved):

kenkit:
I'm on the master branch, latest code btw

PR #2513, opened 2021-01-27
SDK-1376. A fix for 0xFFFF marker pointers used in syncdown()

mattw-mega:


include/mega/node.h
```diff
@@ -205,6 +205,10 @@ struct MEGA_API Node : public NodeCore, FileFingerprint
         bool parent : 1;
         bool publiclink : 1;
         bool newnode : 1;
+
+        // this field is only used internally in syncdown()
+        bool syncdown_node_matched_here : 1;
+
```
sergiohs84, line 211: Should it be enclosed within `ENABLE_SYNC`??
```suggestion cpp

#ifdef ENABLE_SYNC
        // this field is only used internally in syncdown()
        bool syncdown_node_matched_here : 1;
#endif
```

sergiohs84 (changes requested):

sergiohs84 (approved):

PR #2515, opened 2021-01-27
SDK-1374. iOS networking layer memory leak

jnavarrom:
It seems `ares_pending` is causing the leak, avoiding the deletion of the `httpctx`

### DIFF
--- a/src/posix/net.cpp
+++ b/src/posix/net.cpp
@@ -1608,6 +1608,7 @@ void CurlHttpIO::request_proxy_ip()
     
 
 #if TARGET_OS_IPHONE
+    httpctx->ares_pending = 0;
     send_request(httpctx);
 #else
     if (ipv6proxyenabled)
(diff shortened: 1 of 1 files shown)


src/posix/net.cpp
```diff
@@ -1608,6 +1608,7 @@ void CurlHttpIO::request_proxy_ip()
     
 
 #if TARGET_OS_IPHONE
+    httpctx->ares_pending = 0;
```
jnavarrom, line 1611: We are not using c-ares on iOS

simonwng (approved):

src/posix/net.cpp
simonwng: Leaks are fixed, thanks Navarro!

sergiohs84 (approved):

catalinr-m (commented):
Simply setting `ares_pending` to `0` is potentially worrisome, because in a couple of functions the same variable is decremented without being checked for valid value first, and there are no checks for `TARGET_OS_IPHONE` there either. See `CurlHttpIO::proxy_ready_callback()` and `CurlHttpIO::ares_completed_callback()`.

I don't know the code, but after reading that "_We are not using c-ares on iOS_", it might be better to enclose the code dealing with `ares` between some `#ifdef USE_ARES` guards (i.e. defined when `#ifndef TARGET_OS_IPHONE`). That would also eliminate the dead code under iOS.

What do others think?

PR #2523, opened 2021-02-01
SDK-1331. Persistance of sync configs after logout

sergiohs84:


dwmega (approved):

mattw-mega (changes requested):
There is actually already a branch and a PR for this - using a subset of the command set that David added for Backup months ago, so that one is much more compatible with future merges - I think we should close this one.  task/merge_megacli_updated_sync_commands_for_backup
I'll be merging the latest JSON config changes to that one today and double checking the 1331 expected behaviour.  thanks

sergiohs84:
Superseeded by https://github.com/meganz/sdk/pull/2526

PR #2526, opened 2021-02-02
SDK-1331. Task/merge megacli updated sync commands for backup

mattw-mega:


include/mega/types.h
```diff
@@ -334,6 +334,17 @@ typedef enum {
...
+{
+    // Sync is not operating in a backup capacity.
+    SYNC_BACKUP_NONE = 0,
+    // Sync is mirroring the local source.
+    SYNC_BACKUP_MIRROR = 1,
+    // Sync is monitoring (and propagating) local changes.
+    SYNC_BACKUP_MONITOR = 2
+}
+SyncBackupState;
+
```
sergiohs84, line 347: This is not used in the branch

include/mega/types.h
```diff
@@ -363,6 +374,7 @@ enum SyncError {
     TOO_MANY_ACTION_PACKETS = 25,           // Too many changes in account, local state discarded
     LOGGED_OUT = 26,                        // Logged out
     WHOLE_ACCOUNT_REFETCHED = 27,           // The whole account was reloaded, missed actionpacket changes could not have been applied
+    BACKUP_MODIFIED = 28                    // Backup has been externally modified.
```
sergiohs84, line 377: This is not used in this PR

include/megaapi.h
```diff
@@ -12133,9 +12133,12 @@ class MegaApi
          * ToS infringment), the MegaRequest::getParamType indicates the error that
          * triggered the automatic logout (MegaError::API_EBLOCKED for the example).
          *
+         * @param keepSyncConfigsFile Allow sync configs to be recovered if the same user logs in again
+         *        The file containing sync configs is encrypted so there's no privacy issue.
+         *        This is provided for backward compatibiltiy for MEGAsync.
          * @param listener MegaRequestListener to track this request
          */
-        void logout(MegaRequestListener *listener = NULL);
+        void logout(bool keepSyncConfigsFile, MegaRequestListener *listener = nullptr);
```
sergiohs84, line 12141: If we are going to change the public interface, you need to create tickets for all apps

src/megaapi_impl.cpp
```diff
@@ -21659,21 +21660,23 @@ void MegaApiImpl::sendPendingRequests()
                 syncConfig.setError(syncError);
             }
 
-            client->copySyncConfig(syncConfig, [this, request](UnifiedSync *unifiedSync, const SyncError &syncError, error e)
-            {
-                if (!e && !unifiedSync)
-                {
-                    e = API_ENOENT;
-                }
+            client->ensureSyncUserAttributes([this, request, syncConfig](Error e){
```
sergiohs84, line 21663: No error management?? If `e` reports an error, we cannot continue with this request.

src/megaclient.cpp
```diff
@@ -4260,9 +4237,16 @@ void MegaClient::removeCaches()
...
-    if (!mKeepSyncsAfterLogout)
+    if (keepSyncsConfigFile)
     {
+        // Special case backward compatibility for MEGAsync
+        // The syncs will be disabled, if the user logs back in they can then manually re-enable.
+        syncs.disableSyncs(LOGGED_OUT, false);
+    }
+    else
+    {
+        // TODO: actually delete file?
```
sergiohs84, line 4248: `if (!keepSyncsConfigFile)`, then yes, we should actually delete the file.

examples/megacli.cpp
```diff
@@ -6724,7 +6661,7 @@ void DemoApp::request_error(error e)
     if ((e == API_ESID) || (e == API_ENOENT))   // Invalid session or Invalid folder handle
     {
         cout << "Invalid or expired session, logging out..." << endl;
-        client->locallogout(false);
+        client->locallogout(false, false);
```
sergiohs84: The SDK behaves differently in this case, keeping sync configs by default --> megacli should behave the same.

examples/megacli.cpp
```diff
@@ -7180,7 +7117,7 @@ void DemoApp::whyamiblocked_result(int code)
         if (code != 500 && code != 700)
         {
             cout << "Logging out..." << endl;
-            client->locallogout(true);
+            client->locallogout(true, false);
```
sergiohs84: The SDK behaves differently in this case, keeping sync configs by default --> megacli should behave the same.

sergiohs84 (changes requested):

src/megaapi_impl.cpp
```diff
@@ -14513,6 +14510,7 @@ void MegaApiImpl::request_error(error e)
 
     MegaRequestPrivate *request = new MegaRequestPrivate(MegaRequest::TYPE_LOGOUT);
     request->setFlag(false);
+    request->setTransferTag(true); //keepSyncConfigsFile
```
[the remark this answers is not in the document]
sergiohs84: Fixed in 34704fe

src/megaapi_impl.cpp
sergiohs84: Fixed in 4d41619

examples/megacli.cpp
sergiohs84: Actually, we should also remove caches too for this case.
Fixed in bca6388 and aaa0ff8

examples/megacli.cpp
sergiohs84: Fixed in bca6388

include/mega/types.h
sergiohs84: Removed in a9dc23d

include/mega/types.h
sergiohs84: Removed in a9dc23d

sergiohs84 (changes requested):

examples/megacli.cpp
```diff
@@ -8453,3 +8388,277 @@ void exec_banner(autocomplete::ACState& s)
         client->reqs.add(new CommandDismissBanner(client, stoi(s.words[2].s), m_time(nullptr)));
     }
 }
+
+#ifdef ENABLE_SYNC
+
+void exec_syncadd(autocomplete::ACState& s)
+{
+    static int SYNC_TAG = 2027;
```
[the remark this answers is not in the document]
sergiohs84: Solved in c6a0a2a

src/commands.cpp
```diff
@@ -4179,29 +4208,11 @@ bool CommandGetUserData::procresult(Result r)
...
+                    
+                    // uncomment this once the production API is updated to deliver this attribute
+                    //assert(u->getattr(ATTR_JSON_SYNC_CONFIG_DATA) == nullptr);
 
-                    // Persist the new attribute.
-                    client->putua(ATTR_JSON_SYNC_CONFIG_DATA,
-                                  reinterpret_cast<const byte*>(payload->data()),
-                                  static_cast<unsigned>(payload->size()),
-                                  0);
+                    client->ensureSyncUserAttributes(nullptr);
```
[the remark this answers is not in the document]
sergiohs84: Adressed in 0d27b07

src/megaclient.cpp
```diff
@@ -13083,7 +13081,95 @@ error MegaClient::checkSyncConfig(SyncConfig& syncConfig, LocalPath& rootpath, s
...
+                    reqs.add(new CommandGetUA(this, uid.c_str(), ATTR_JSON_SYNC_CONFIG_DATA, nullptr, 0,
+                        [this](error e) {                 ensureSyncUserAttributesCompleted(e); },
+                        [this](byte*, unsigned, attr_t) { ensureSyncUserAttributesCompleted(API_OK); },
+                        [this](TLVstore*, attr_t) {       ensureSyncUserAttributesCompleted(API_OK); } ));
+                }
+                else
+                {
+                    LOG_info << "JSON config data created: " << error(e);
+
+                    ensureSyncUserAttributesCompleted(API_OK);
```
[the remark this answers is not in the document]
sergiohs84: Fixed in 9f16e6c

src/megaclient.cpp
sergiohs84: In my tests it does delete the file. We should remove the TODO

examples/megacli.cpp
```diff
@@ -8453,3 +8388,277 @@ void exec_banner(autocomplete::ACState& s)
...
+                   << " byte(s) across "
+                   << sync->localnodes[FILENODE]
+                   << " file(s) and "
+                   << sync->localnodes[FOLDERNODE]
+                   << " folder(s).\n";
+          }
+          else
+          {
+              // Display what status info we can.
+              auto msg = config.syncErrorToStr();
```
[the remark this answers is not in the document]
mattw-mega: Nice one, thanks

src/megaclient.cpp
```diff
@@ -13106,8 +13192,6 @@ void MegaClient::copySyncConfig(SyncConfig& config, SyncCompletionFunction compl
         }
         else
         {
-            config.setBackupId(backupId);
```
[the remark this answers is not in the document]
mattw-mega: The caller passes in `config`, which used to be a non-const reference.   Making that parameter const meant we could have a simpler syntax for passing that variable to this function (as it was a lambda member).   The caller does not check the backupId in the object it passed anyway, that was removed some time ago.   The caller instead has their completion function called, which includes a reference to the UnifiedSync, and its SyncConfig (which is a copy of this one passed into the function, and subsequently updated with the backupId).    So, we don't need to call setBackupId here, the set value is not used, and we can't call it now with the object being a const paramter.

src/megaclient.cpp
mattw-mega: removed

src/megaclient.cpp
```diff
@@ -13083,7 +13081,95 @@ error MegaClient::checkSyncConfig(SyncConfig& syncConfig, LocalPath& rootpath, s
...
+        {
+            // attributes already exist.
+            if (completion) completion(API_OK);
+            return;
+        }
+    }
+    else
+    {
+        // If there's no user object, there can't be user attributes
+        if (completion) completion(API_EEXIST);
```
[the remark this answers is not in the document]
mattw-mega: updated to ENOENT

examples/megacli.cpp
```diff
@@ -8453,3 +8388,277 @@ void exec_banner(autocomplete::ACState& s)
...
+    }
+
+    // Do we have full access to the target node?
+    if (!client->checkaccess(targetNode, FULL))
+    {
+        cerr << targetPath
+             << ": Insufficient privileges."
+             << endl;
+        return;
+    }
```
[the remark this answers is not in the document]
mattw-mega: Yes, good point.  All validation is performed in the SDK Core

examples/megacli.cpp
```diff
@@ -300,6 +296,46 @@ const char* errorstring(error e)
     }
 }
 
+#ifdef ENABLE_SYNC
+
+const char* syncstatename(const syncstate_t state)
```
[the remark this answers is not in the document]
mattw-mega: moved to SyncConfig

src/commands.cpp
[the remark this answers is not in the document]
mattw-mega: good point, and nice catch, thanks

src/megaclient.cpp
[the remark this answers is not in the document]
mattw-mega: good catch, thanks

src/megaclient.cpp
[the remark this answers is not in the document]
mattw-mega: On second thought - and after testing with MEGAsync - we do need this here of course, good spotting.   This is the config that gets passed to the Syncs object for creation.   I had thought we were within the Syncs' construction already.  anyway, that is restored now (in a new modifiable copy of config).

src/megaclient.cpp
```diff
@@ -13093,20 +13092,24 @@ void MegaClient::ensureSyncUserAttributes(std::function<void(Error)> completion)
 {
     // If the attributes are not available yet, we make or get them.
 	// Then the completion function is called.
-	
+
+    // we rely on storing this function to remember that we have an
+    // operation in progress, so we don't allow nullptr
+    assert(!!completion);
```
sergiohs84, line 13098: Nice one!

src/sync.cpp
```diff
@@ -3393,6 +3393,11 @@ bool JSONSyncConfigIOContext::deserialize(SyncConfig& config, JSON& reader) cons
 
         case TYPE_TARGET_HANDLE:
             config.mRemoteNode = reader.gethandle(MegaClient::NODEHANDLE);
+            if ((config.mRemoteNode & 0xFFFFFFFFFFFF) == (UNDEF & 0xFFFFFFFFFFFF))
+            {
+                // we can have a much nicer solution when NodeHandle is merged from the sync rework branch
+                config.mRemoteNode = UNDEF;
+            }
```
sergiohs84, line 3400: Actually, `mRemoteNode` is not initialized when it reaches the `gethandle()`. If we initialize it to `UNDEF` in the declaration of the variable, when we reach this point, we could test for `if (!ISUNDEF(config.mRemoteNode))` and it will be enough, won't it?
If you have created a new class NodeHandle to avoid the issues of handles of 6 bytes, I hope that all refactoring is targeting `develop` as an enhancement / maintenance :crossed_fingers:

sergiohs84 (approved):

PR #2529, opened 2021-02-03
SDK-1383 - Restore the legacy DB behavior

dwmega:
Databases are now created using the correct database version.

A bank of unit tests have been added to validate the expected behavior of open, probe, rootPath.

### DIFF
--- a/include/mega/db.h
+++ b/include/mega/db.h
@@ -159,8 +159,8 @@ enum DbOpenFlag
 
 struct MEGA_API DbAccess
 {
-    static const int LEGACY_DB_VERSION = 11;
-    static const int DB_VERSION = LEGACY_DB_VERSION + 1;
+    static const int LEGACY_DB_VERSION;
+    static const int DB_VERSION;
 
     DbAccess();
 
--- a/include/mega/db/sqlite.h
+++ b/include/mega/db/sqlite.h
@@ -26,21 +26,6 @@
 #include <sqlite3.h>
 
 namespace mega {
-class MEGA_API SqliteDbAccess : public DbAccess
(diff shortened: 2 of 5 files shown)


include/mega/db.h
```diff
@@ -159,8 +159,8 @@ enum DbOpenFlag
 
 struct MEGA_API DbAccess
 {
-    static const int LEGACY_DB_VERSION = 11;
-    static const int DB_VERSION = LEGACY_DB_VERSION + 1;
+    static const int LEGACY_DB_VERSION;
```
dwmega, line 162: Necessary in order for the unit tests to successfully link.

include/mega/db/sqlite.h
```diff
@@ -65,7 +50,30 @@ class MEGA_API SqliteDbTable : public DbTable
     ~SqliteDbTable();
 
     bool inTransaction() const override;
+
+    LocalPath dbFile() const;
+};
+
+class MEGA_API SqliteDbAccess : public DbAccess
```
dwmega, line 57: Moved so open(...) can have a covariant return type.

include/mega/db/sqlite.h
```diff
@@ -65,7 +50,30 @@ class MEGA_API SqliteDbTable : public DbTable
     ~SqliteDbTable();
 
     bool inTransaction() const override;
+
+    LocalPath dbFile() const;
```
dwmega, line 54: Added so we can query the DB path from the unit tests.

mattw-mega (approved):

sergiohs84 (approved):

PR #2548, opened 2021-02-10
SDK-1397. Introduce NodeHandle helper class, fix some tests

mattw-mega:
This PR is starting to merge some improvements back from the sync rework branch
 - adding NodeHandle class.  It wraps up 6 byte node handles, to avoid accidental 8 byte usages, and avoid common mistakes with compiler errors/warnings, and asserts.
 - Fix some test cases where the listeners were not being used, and race conditions caused the tests to get confused.
 - The File class is updated to use NodeHandle - it's a good place to start as the handle contained there is not referenced by too many lines.
 - Over time other usages should migrate to NodeHandle also, and eventually we should be able to do away with the (deliberately awkward) set6byte() and as8byte() members.

### DIFF
--- a/include/mega/megaclient.h
+++ b/include/mega/megaclient.h
@@ -1389,6 +1389,7 @@ public:
     // remove node subtree
     void deltree(handle);
 
+    Node* nodeByHandle(NodeHandle) const;
     Node* nodebyhandle(handle) const;
     Node* nodebyfingerprint(FileFingerprint*);
 #ifdef ENABLE_SYNC
@@ -1724,10 +1725,10 @@ public:
     Node *getrootnode(Node*);
 
     //returns true if the node referenced by the handle belongs to the logged-in account
-    bool isPrivateNode(handle h);
+    bool isPrivateNode(NodeHandle h);
 
     //returns true if the node referenced by the handle belongs to other account than the logged-in account
-    bool isForeignNode(handle h);
+    bool isForeignNode(NodeHandle h);
 
     // process node subtree
     void proctree(Node*, TreeProc*, bool skipinshares = false, bool skipversions = false);
--- a/include/mega/types.h
+++ b/include/mega/types.h
@@ -132,6 +132,7 @@ struct PendingContactRequest;
 class TransferList;
 struct Achievement;
 class SyncConfig;
+class SimpleLogger;
 
 namespace UserAlert
 {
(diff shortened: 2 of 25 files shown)


dwmega (dismissed):

include/mega/megaclient.h
```diff
@@ -1389,6 +1389,7 @@ class MEGA_API MegaClient
     // remove node subtree
     void deltree(handle);
 
+    Node* nodeByHandle(NodeHandle) const;
```
sergiohs84, line 1392: I think we should make a comment to the getter with `handle`:
```suggestion
    Node* nodeByHandle(NodeHandle) const;
    // obsolete, use version with NodeHandle as input param
```

include/mega/types.h
```diff
@@ -262,6 +263,27 @@ typedef enum { NOTLOGGEDIN, EPHEMERALACCOUNT, CONFIRMEDACCOUNT, FULLACCOUNT } se
 // in a 64-bit int
 typedef uint64_t handle;
 
+class NodeHandle
+{
+    // Handles of nodes are only 6 bytes.
+    // This class helps avoid issues when we don't save/restore the top 2 bytes when using an 8 byte uint64 to represent it
+    uint64_t h = 0xFFFFFFFFFFFFFFFF;
```
sergiohs84, line 270: Are we planning to get rid of the `UNDEF` at somepoint? otherwise, I think that we should use it here:
```suggestion
    uint64_t h = UNDEF;
```
However, it might be more clear to indicate its `0xFFFFFFFFFFFFFFFF` :thinking:

include/mega/types.h
```diff
@@ -262,6 +263,27 @@ typedef enum { NOTLOGGEDIN, EPHEMERALACCOUNT, CONFIRMEDACCOUNT, FULLACCOUNT } se
...
+public:
+    bool isUndef() { return (h & 0xFFFFFFFFFFFF) == 0xFFFFFFFFFFFF; }
+    NodeHandle& set6byte(uint64_t n) { h = n; assert((n & 0xFFFF000000000000) == 0); return *this; }
```
sergiohs84, line 273: Perhaps it would be nice to define some masks, for easier reading (I've already needed to count the `F`s many times).
Something like:
``` cpp
static const uint64_t leastSignificant6B = 0x0000FFFFFFFFFFFF;
static const uint64_t mostSignificant2B = 0xFFFF000000000000;
```

include/mega/utils.h
```diff
@@ -57,6 +57,7 @@ namespace mega {
 #define MAKENAMEID8(a, b, c, d, e, f, g, h) (nameid)((((uint64_t)a) << 56) + (((uint64_t)b) << 48) + (((uint64_t)c) << 40) + (((uint64_t)d) << 32) + ((e) << 24) + ((f) << 16) + ((g) << 8) + (h))
 
 std::string toNodeHandle(handle nodeHandle);
```
sergiohs84, line 59: Tag this one as obsolete, in pro of new one.

include/mega/types.h
```diff
@@ -262,6 +263,27 @@ typedef enum { NOTLOGGEDIN, EPHEMERALACCOUNT, CONFIRMEDACCOUNT, FULLACCOUNT } se
...
+    bool eq(NodeHandle b) { return (h & 0xFFFFFFFFFFFF) == (b.h & 0xFFFFFFFFFFFF); }
+    bool eq(handle b) { return (h & 0xFFFFFFFFFFFF) == (b & 0xFFFFFFFFFFFF); }
+    bool ne(handle b) { return (h & 0xFFFFFFFFFFFF) != (b & 0xFFFFFFFFFFFF); }
+    bool operator<(const NodeHandle& rhs) const { return h < rhs.h; }
+    handle as8byte() { return isUndef() ? 0xFFFFFFFFFFFFFFFF : (h & 0xFFFFFFFFFFFF); }
```
sergiohs84, line 278: Perhaps some comments about this `as8byte()` to indicate that:
```suggestion
    // TODO: this method is here for convenience/backwards compatibility. Once all handles are replaced by NodeHandles, it should be removed
    handle as8byte() { return isUndef() ? 0xFFFFFFFFFFFFFFFF : (h & 0xFFFFFFFFFFFF); }
```

src/utils.cpp
```diff
@@ -45,6 +45,10 @@ string toNodeHandle(handle nodeHandle)
     return string(base64Handle);
 }
 
+string toNodeHandle(NodeHandle nodeHandle)
+{
+    return toNodeHandle(nodeHandle.as8byte());
```
sergiohs84, line 50: If you plan to replace the old code, perhaps this method should not depend on the old code, but the other way round.

mattw-mega: I'm sure we'll do that eventually, but for this initial PR surely it's better to (a) leave the working code alone (even though it is very simple) and (b) keep the number of lines changed to a minimum

include/mega/types.h
mattw-mega: I was just aiming to keep the class small (ie fewer lines) ; the main thing is that the interface lets us use NodeHandles simply and effectively in the rest of the code.   As to whether 1+3 is clearer than 3+1, well, maybe :-)

include/mega/types.h
mattw-mega: I don't think this matters - either way is fine.   The FFs might make comparison with the other constants a little bit easier

mattw-mega: And, we will keep UNDEF for the 8 byte handles

dwmega (approved):

mattw-mega:
trigger compilation

PR #2557, opened 2021-02-17
SDK-1401. Generating thumbnails and previews for webP images (iOS)

jnavarrom:


lhr000lhrmega (approved):

lhr000lhrmega:
I was wondering is it possible that not hard code these supported types.
Based on my test, Image I/O(on  iOS 11) does not crash or show any side effect, even we added `webp` here.

src/gfx/GfxProcCG.mm
```diff
@@ -62,7 +62,7 @@
     if ([[NSBundle.mainBundle objectForInfoDictionaryKey:@"CFBundleExecutable"] isEqualToString:@"MEGAFiles"]) {
         return "";
     }
-    return ".bmp.cr2.crw.cur.dng.gif.heic.ico.j2c.jp2.jpf.jpeg.jpg.nef.orf.pbm.pdf.pgm.png.pnm.ppm.psd.raf.rw2.rwl.tga.tif.tiff.3g2.3gp.avi.m4v.mov.mp4.mqv.qt.";
+    return ".bmp.cr2.crw.cur.dng.gif.heic.ico.j2c.jp2.jpf.jpeg.jpg.nef.orf.pbm.pdf.pgm.png.pnm.ppm.psd.raf.rw2.rwl.tga.tif.tiff.3g2.3gp.avi.m4v.mov.mp4.mqv.qt.webp.";
```
simonwng, line 65: since we have error handling in Image I/O, shall we remove this format list?

simonwng (approved):

src/gfx/GfxProcCG.mm
jnavarrom: We can't remove this line. 
We have to implement the following virtual method:

```
// list of supported extensions (NULL if no pre-filtering is needed)
virtual const char* supportedformats();
```

To remove it, we would need changes in the SDK

simonwng: It has a comment:
```
// list of supported extensions (NULL if no pre-filtering is needed)
```
It seems if we return NULL, there will be no filtering?

jnavarrom: This is how the others implementations of the gfx layer implement `supportedFormat()`:

https://github.com/meganz/sdk/blob/develop/src/gfx/freeimage.cpp#L383
https://github.com/meganz/sdk/blob/develop/src/gfx/qt.cpp#L1425

simonwng: https://jira.developers.mega.co.nz/browse/SDK-1405 For further investigation of removing the format list

PR #2560, opened 2021-02-17
SDK-1403 - Correctly check whether a local path is syncable.

dwmega:
The SDK calls isLocalPathSyncable(...) to determine whether a given local path is acceptable.

One of the checks it performs is to verify that there is only a single active sync for a given local path.

There are times where isLocalPathSyncable(...) is called after the sync for a given path has already been committed to memory. Say, when enabling a sync. For this reason, isLocalPathSyncable(...) allows the caller to skip checking against a specified Backup ID. This prevents the function from comparing a given sync against itself.

The purpose of this change-set is to correct the logic that performs that skipping.


src/megaclient.cpp
```diff
@@ -12983,8 +12983,8 @@ error MegaClient::isLocalPathSyncable(const LocalPath& newPath, handle excludeBa
 
     error e = API_OK;
     syncs.forEachSyncConfig([&](const SyncConfig& config){
-
-        if (config.getBackupId() != excludeBackupId && excludeBackupId != UNDEF) // for the check inside addsync() when it's already present
+        // For the check inside addsync(...) when the config's already present.
+        if (excludeBackupId == UNDEF || config.getBackupId() != excludeBackupId)
```
sergiohs84, line 12987: Good catch, David. This is another bug introduced by the revamping of the sync-management & sync-heartbeats to pre-register syncs in the API before they are effectively added to the set.

Given that no sync-config can be added without a backupId, I'd say that we can simplify the check. I'd also make a more detailed comment:
```suggestion cpp
        // skip the check if the sync has not been registered yet (check from addsync() -> excludeBackupId = UNDEF)...
        // or if the sync has been registered already (check from eanblesync() -> excludeBackupId is this config's id)
        if (config.getBackupId() != excludeBackupId)
```

sergiohs84 (changes requested):

sergiohs84:
This branch needs to be a hotfix for v3.8.2, and it was forked from `develop`.
Closing in pro of https://github.com/meganz/sdk/pull/2561 and https://github.com/meganz/sdk/pull/2562

PR #2569, opened 2021-02-19
SDK-1407. Fix duplicates searching on outshares

jgandres:


Yenel (approved):
Tested, works fine.

src/megaapi_impl.cpp
```diff
@@ -11774,9 +11774,15 @@ MegaNodeList* MegaApiImpl::search(MegaNode *n, const char* searchString, MegaCan
         if (target == MegaApi::SEARCH_TARGET_OUTSHARE)
         {
             // Search on outshares
+            std::set<MegaHandle>outsharesHandles;
```
sergiohs84, line 11777:
```suggestion
            std::set<MegaHandle> outsharesHandles;
```

src/megaapi_impl.cpp
```diff
@@ -11774,9 +11774,15 @@ MegaNodeList* MegaApiImpl::search(MegaNode *n, const char* searchString, MegaCan
...
             unique_ptr<MegaShareList>shares (getOutShares(MegaApi::ORDER_NONE));
             for (int i = 0; i < shares->size() && !(cancelToken && cancelToken->isCancelled()); i++)
             {
+                if (outsharesHandles.find(shares->get(i)->getNodeHandle()) != outsharesHandles.end())
+                {
+                    continue;   // avoid duplicates
+                }
+                outsharesHandles.insert(shares->get(i)->getNodeHandle());
```
sergiohs84, line 11785: You can improve readability a bit:
```suggestion cpp
                handle h = shares->get(i)->getNodeHandle();
                if (outsharesHandles.find(h) != outsharesHandles.end())
                {
                    // shares list includes an item per outshare AND per sharee/user
                    continue;   // avoid duplicates
                }
                outsharesHandles.insert(h);
```
Also, should use the declared `h` in the next line too :)

sergiohs84 (changes requested):
Good catch (and patch) ;)

PR #2597, opened 2021-11-25
Allow sdk to have an out of source build directory

thebrave:
I really like for code I work on to be able to build from outside the work directory. 

There were some missing reference to include files, so I fixed them.


sergiohs84:
I fear this PR is obsolete. We moved from autotools to cmake+vcpkg, so these changes cannot be merged.
Thanks for the contribution, anyway.

PR #2628, opened 2022-11-08
autoconf m4 - (PEP632) python distutils deprecation patch

sssomeshhh:


sergiohs84:
MEGA has decided to stop maintaining Python bindings. The PR becomes obsolete. Closing...

Issue #2661, opened 2024-03-14
Compilation support on various BSDs


kikadf:
I am a satisfied user of the MEGA service, thank you for your work. I use the mega-sdk (megacmd) on Linux, on FreeBSD, as well as with minor modifications DragonFly BSD, NetBSD, and OpenBSD. In the case of FreeBSD and NetBSD, I maintain the megacmd package that is available in their package repositories, but I use it with full functionality and without issues on all BSD systems.

I am sending some pull requests that make successful compilation on the various BSDs.

I understand that DragonFly, NetBSD, and OpenBSD are not officially supported platforms and there are no plans for that, however, only simple changes (that do not affect other platforms) are needed to make the mega-sdk compile on these systems as well.

I hope you find these acceptable and worth to merging.

sergiohs84:
Hi @kikadf . Thanks a lot for your contribution. I've closed 2 out of 4 PRs you created, since they are obsolete. However, for the other two, we will add them to our internal repository first. We will notify you when they are merged and published in GitHub.
Once again, thank you.

sergiohs84:
Your changes will be included in the next release ;) 

![image](https://github.com/user-attachments/assets/5daa5821-2738-4b29-8130-b176ea8617ae)

Thanks for your contribution.

kikadf:
> Your changes will be included in the next release ;)
> 
Thank you!

sergiohs84:
Well, the MR missed the next release (8.0.0, not live yet), but it's already merged in our `develop` branch, so it will definitely added to the next release after 8.0.0. Thanks

PR #2662, opened 2024-03-14
Add support to other BSDs

kikadf:
These changes add some extra #ifdef where FreeBSD already used.
Reference: #2661


sergiohs84:
The relevant commit from this PR has been cherry-picked and it will be merged into the private repository.
The next version of the SDK will include your changes. Thanks, @kikadf .

PR #2663, opened 2024-03-14
Fix libtool flags on DragonFly and NetBSD

kikadf:
These changes add --tag=CXX to libtool flags on DragonFly and NetBSD as on Linux or Android.
Reference: #2661


sergiohs84:
Thank you for the PR, @kikadf , but we have removed support for the old building system based on autotools.
Currently, the building system in place is based on CMake and VCPKG.

PR #2664, opened 2024-03-14
Compilation fix without AI_V4MAPPED

kikadf:
These changes fix compilation on Net- and OpenBSD, which don’t support AI_V4MAPPED. The solution doesn’t affect other platforms.
Reference: #2661


sergiohs84:
The changes in this PR are already included in https://github.com/meganz/sdk/pull/2662. Closing...

kikadf:
> The changes in this PR are already included in #2662. Closing...

This changes aren't included in #2662.

sergiohs84:
Oh, you are right, but that method has been removed from our code base. After solving the conflict, there were no changes.

PR #2665, opened 2024-03-14
Allow --with-rt configure option

kikadf:
These changes add --with-rt=...  configure option. With that possible to disable rt if it is present on the host system, or set the path. The changes don’t modify the compile, not necessary to update any build scripts. Tested on linux with every rt configure option, works fine.


sergiohs84:
We have moved our building system to CMake. In consequence, I fear the `configure.ac` and `src/include.am` are not any more part of the repository. On the other hand, support real-time capabilities is out of the scope for our SDK library. Thanks for the contribution, anyway.

PR #2676, opened 2024-06-02
Update x64-windows-mega.cmake

Kapil-Vermani:
Adding support for Visual Studio 2022 and changing references of Program Files (x86) to Program Files as in x64 system visual studio will be available in that folder


sergiohs84:
We have added support for Visual Studio 2022, but with slightly different changes. It will be published in the next release.
Thanks for your contribution, @Kapil-Vermani .

PR #2768, opened 2026-03-23
Fix missing ICU i18n linkage in non-vcpkg build path

ChrisLane:
## Summary

The non-vcpkg (system dependency) code path in `sdklib_libraries.cmake` links `ICU::uc` and `ICU::data` but omits `ICU::i18n`, which is required for `icu::Collator::createInstance()` used in `sdk/src/utils.cpp`. This causes an undefined reference error at link time:

```
undefined reference to `icu_78::Collator::createInstance(icu_78::Locale const&, UErrorCode&)'
```

The vcpkg code path correctly includes all three components (`i18n`, `uc`, `data`), but the system dependency path was missed.

This was introduced in 1f5c9cf8bd98a165aa65a9b05a6db118e35353e9.

## Fix

Add `ICU::i18n` to both the `find_package` and `target_link_libraries` calls in the non-vcpkg section, matching what the vcpkg section already does.

### DIFF
--- a/cmake/modules/sdklib_libraries.cmake
+++ b/cmake/modules/sdklib_libraries.cmake
@@ -110,8 +110,8 @@ macro(load_sdklib_libraries)
         pkg_check_modules(curl REQUIRED IMPORTED_TARGET libcurl)
         target_link_libraries(SDKlib PRIVATE PkgConfig::curl)
 
-        find_package(ICU COMPONENTS uc data REQUIRED)
-        target_link_libraries(SDKlib PRIVATE ICU::uc ICU::data)
+        find_package(ICU COMPONENTS i18n uc data REQUIRED)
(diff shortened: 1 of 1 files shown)


ChrisLane:
@polmr tagging you since this breaks the build for the [AUR megacmd package](https://aur.archlinux.org/packages/megacmd) and we're having to apply this patch ourselves currently.

@cjr-mega the breakage was introduced by your change.

cjr-mega:
Hi @ChrisLane ,
Thanks for the patch. We’ve applied the fix in our internal repository, and it will be included in the next release.

ChrisLane:
@cjr-mega the fix doesn't appear to be in the release since your comment (https://github.com/meganz/sdk/releases/tag/v10.12.0)

PR #2769, opened 2026-03-24
link i18n component of icu when not using VCPKG_ROOT

UlyssesZh:


cjr-mega:
Hi @UlyssesZh ,
Thanks for the patch. We’ve applied the fix in our internal repository, and it will be included in the next release.

PR #2775, opened 2026-09-14
Fix Linux descriptor exhaustion and sync-cache startup crash

gabriel90:
- Failed database opens caused a null-pointer dereference during sync startup.
- Curl’s idle-cache limit didn’t bound connections awaiting TLS shutdown; the core contained 750 closing upload connections.
- The fix bounds cached and total transfer connections, including after reconnects.
- An isolated TLS test reproduced exhaustion before the fix and passed afterward.
- Your rebuilt application successfully synced over 46,000 files.


gabriel90:
This fixes the issue of megasync crashing during a sync where there are a large amount of files to be synced.

